PR #32, opened 2013-02-07
`git subtree push` quieting

boinger:
When running `git subtree push --prefix /what/ever`, the stream of "1/x 2/x 3/x....n/x" noise is not helpful.  This patch just quashes that by pushing the stderr of the `git subtree split` subcommand of cmd_push to /dev/null.

### DIFF
--- a/contrib/subtree/git-subtree.sh
+++ b/contrib/subtree/git-subtree.sh
@@ -110,9 +110,9 @@ if [ -z "$prefix" ]; then
 fi
 
 case "$command" in
-	add) [ -e "$prefix" ] && 
+	add) [ -e "$prefix" ] &&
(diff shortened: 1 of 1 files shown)


contrib/subtree/git-subtree.sh
```diff
@@ -703,7 +703,7 @@ cmd_push()
 	    repository=$1
 	    refspec=$2
 	    echo "git push using: " $repository $refspec
-	    git push $repository $(git subtree split --prefix=$prefix):refs/heads/$refspec
+	    git push $repository $(git subtree split --prefix=$prefix 2>/dev/null):refs/heads/$refspec
 	else
 	    die "'$dir' must already exist. Try 'git subtree add'."
 	fi
```
elle76: 7070

PR #54, opened 2013-09-07
Improvements to hg-to-git script

otac0n:
Portability fixes regarding environment variables and removing files.
Performance and correctness fixes around reading metadata and branching.
Override autocrlf so that a binary-identical files are committed.
Other small bug fixes.
Small refactorings.


gitster:
No activity for more than 18 months since "I'll send patches to the list and then will close"?

Let's close it here first and see the patch there. Thanks.

PR #110, opened 2014-10-22
fix OS X system error, Update git-rebase.sh

mylukin:
fix OS X system error：
/usr/local/Cellar/git/2.1.2/libexec/git-core/git-rebase: line 47: git-sh-setup: No such file or directory

### DIFF
--- a/git-rebase.sh
+++ b/git-rebase.sh
@@ -44,8 +44,8 @@ abort!             abort and check out the original branch
 skip!              skip current patch and continue
 edit-todo!         edit the todo list during an interactive rebase
 "
-. git-sh-setup
-. git-sh-i18n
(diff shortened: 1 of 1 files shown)


zaggino:
@gitster I'm calling `git rebase --ignore-date master` (no `git-rebase` directly) yet this error still occurs on Mac. The git call itself is not done manually but rather through a node's `child_process.spawn` utility with a custom working directory. Should I be setting `$(git --exec-path)` variable manually before calling this? Ref: https://github.com/zaggino/brackets-git/issues/816

gitster:
Please take this to git@vger.kernel.org if you still have issues. There are more experts monitoring that list than this pull request to help you figure out configuration errors that may cause symptoms like this.

PR #116, opened 2014-11-13
[catalan]: change rama to branca

zblz:
In Catalan, 'rama' refers to a tree branch with leaves, whereas 'branca' is, in addition, used in the more abstract way of subdivisions of a subject. In my opinion, 'branca' is a better translation of 'branch'.

### DIFF
--- a/po/ca.po
+++ b/po/ca.po
@@ -136,50 +136,50 @@ msgstr ""
 #: branch.c:60
 #, c-format
 msgid "Not setting branch %s as its own upstream."
-msgstr "No establint la rama %s com la seva pròpia font."
+msgstr "No establint la branca %s com la seva pròpia font."
 
 #: branch.c:83
 #, c-format
 msgid "Branch %s set up to track remote branch %s from %s by rebasing."
 msgstr ""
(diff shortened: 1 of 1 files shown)


elboletaire:
Back to 80's....

I'm sorry but I cannot understand why you publish things like git or the linux kernel to github if we cannot colaborate directly using github :confused:

rumpelsepp:
It is just a mirror and quite a mess to maintain two places where development happens.

elboletaire:
Ok, then, why don't you move to github?

I'm currently answering this from my e-mail, like you would do in the
mailing list.

Also, you would gain a lot of collaborators like the creator of this issue
or like me. I don't know if the issue creator finally joined the mailing
list, that for such a change, I wouldn't.

Please don't take these messages as an "attack" to the way you work, I'm
just asking because I can't understand it.

..................................



Sent from my iPhone

El 08/07/2015, a les 17:29, Stefan Tatschner <notifications@github.com> va
escriure:

It is just a mirror and quite a mess to maintain two places where
development happens.

—
Reply to this email directly or view it on GitHub
<https://github.com/git/git/pull/116#issuecomment-119625549>.

PR #119, opened 2014-12-13
Show author in interactive rebase

dannykopping:
via @yurial - https://gist.github.com/yurial/1909363

### DIFF
--- a/git-rebase--interactive.sh
+++ b/git-rebase--interactive.sh
@@ -961,7 +961,7 @@ else
 	revisions=$onto...$orig_head
 	shortrevisions=$shorthead
 fi
-git rev-list $merges_option --pretty=oneline --abbrev-commit \
+git rev-list $merges_option --pretty=">%h (%an <%ae>) %s" \
 	--abbrev=7 --reverse --left-right --topo-order \
 	$revisions ${restrict_revision+^$restrict_revision} | \
 	sed -n "s/^>//p" |


rcdailey:
Interesting that the Git developers continue to use archaic methods of code sharing (i.e. mailing lists) instead of utilizing the wonderful features provided by Github. Maybe one day you all will wake up and use proper tooling. The way you do it now is too frustrating; I won't be contributing until github features are used.

petervanderdoes:
@rcdailey Linus gave an explanation in 2012 on why he doesn't do PR's on the Kernel repository:
https://github.com/torvalds/linux/pull/17#issuecomment-5654674 and
https://github.com/torvalds/linux/pull/17#issuecomment-5659933

I'm sure the same reasons apply to the git repository.

rcdailey:
I don't agree with Linus on this; I don't feel the problem is the pull request feature itself but rather his annoyances with the lack of discipline in the information people do (or don't) provide in a commit message. Let's keep using a wheelbarrow to haul piles of rocks instead of a tractor because I can feel closer to the rocks when I use my hands.

But I guess the old mage gets it the way he wants. Doesn't mean I have to have anything to do with it.

torvalds:
On Fri, Jan 2, 2015 at 12:56 PM, Robert Dailey <notifications@github.com> wrote:
>
> I don't agree with Linus on this

My heart bleeds.

> I don't feel the problem is the pull request feature itself but rather his annoyances with the lack of discipline in the information people do (or don't) provide in a commit message.

No. That too, but the pull requests from github are actively *crap*,
exactly the same way the web interface commit messages used to be
(and, quite frankly, still are, although now github at least *tries*
to encourage better commit message formats)

The github pull messages are not useful. They don't contain any of the
required information for a useful pull message.

In fact, the very pull request (and commit) in this whole thread shows
exactly *why* github pull requests and commits are pure and utter
crap:

  No sane explanation, no sign-off on the commit, no diffstat in the
pull request, no valid emails anywhere. Or even a valid real name. No
*nothing*.

Seriously. You can disagree with me all you want, but you are *wrong*.

Please just go away.

                    Linus

rcdailey:
You don't have to go to such great lengths to repeat yourself, I read all of these points on the original thread (kindly linked by peter). But I had to stop reading once you started calling people morons. For someone who has done so many great things, you can be a real a**hole (sadly it allows you to get away with it).

I'm not wrong, just opinionated the same as you. I don't agree with your opinions, it's that simple. Don't take it personally. The overhead of the issues you complain about is far outweighed by the unintuitive and complex nature of delivering code over mail protocol:

1. I have to install a mail client capable of viewing data in monospaced fonts (using Gmail as my mail reader is very convenient but I cannot use it for this; unified diffs are unreadable).
2. Setup and usage per platform is not the same nor as easy. For example, it is more of a pain on Windows platforms, not sure about Linux but probably easier.
3. Setting up git to send and receive patches via mail is incredibly complex and practically useless outside of linux & git development. It takes a very long time to setup and most pragmatic programmers will give up on it seeing as how there are much better facilities for code sharing (hint: pull requests).

At least github focuses on usability and it is incredibly functional. It isn't perfect, but what is when everyone wants it their way? Fortunately due to age difference I will most likely outlive you and after such time perhaps we will see git's methods of development improve.

Thanks for git, it's a great tool. But such a great thing is stifled by the stubbornness of one individual.

PS: Going away is the easy part!

mafagafogigante:
Installing such a mail client should not be hard.

Linux is easier? Imagine if it was also free.

If you are not determined enough even to set your environment up, how long do you intend to contribute?

"no syntax highlight."...

So you wrapped up with the "I will outlive you" argument? Shame is on you.

Some people are more stubborn than others, and I personally find Linus quite a bit stubborn. One way or the other, Git is my favorite VCS and if its developers used and still use a mailing list, why should I (or any other newbie) just point the finger and insult they or the way they do things? Many others out there likely share your opinion, but this is not so much of a democracy and changes of workflow should likely be discussed by developers in the mailing list.

Additionally, I can't help but depreciate people that open PRs here even when there is a freaking message telling you not to.

ThiefMaster:
I think the biggest disadvantage of a mailing list is the fact that you need to subscribe to it. If i want to contribute a small fix to a project I most likely do not care about any other discussions besides those related to my patch. But subscribing to a mailing list give me the whole package(tm). Sure, I could use filtering (either in my client or via sieve on the server), but why have all the uninteresting stuff arrive at my MTA in the first place?

peff:
> I think the biggest disadvantage of a mailing list is the fact that you need to subscribe to it.

Just FYI, you do not need to subscribe to the git mailing list. You can send to it without subscribing, and it is customary for responders to cc you on any replies. This is useful for anybody who is casually contributing to a list (e.g., git issues often bleed over from kernel development lists, and those kernel folks need only add git@vger to the cc of an existing thread).

The biggest friction point I have seen is that the git list does not accept HTML mail, and people using gmail have a very hard time turning it off.

npcode:
> The biggest friction point I have seen is that the git list does not accept HTML mail, and people using gmail have a very hard time turning it off.

And some email services does not allow users to send emails in non-HTML format.

mafagafogigante:
> The biggest friction point I have seen is that the git list does not accept HTML mail, and people using gmail have a very hard time turning it off.

As far as my experience goes, sending a plain text e-mail in GMail requires two mouse clicks (and receiving plain text is trivial, AFAIK). But I may be misunderstanding something.

rumpelsepp:
> Setting up git to send and receive patches via mail is incredibly complex and practically useless outside of linux & git development.

It's actually 5 lines in the config which you can mostly copy & paste: http://git-scm.com/docs/git-send-email#_example

If you're working on you're own branch it's two short commands to send patches. As you do it in the terminal it's much faster than clicking on github's web interface...

PR #124, opened 2015-01-23
Add support for detecting C# async methods in diff

thomaslevesque:
Currently C# async methods are not shown in diff hunk headers. I just
added the async keyword to the csharp method pattern so that they are
detected.

### DIFF
--- a/userdiff.c
+++ b/userdiff.c
@@ -136,7 +136,7 @@ PATTERNS("csharp",
 	 /* Keywords */
 	 "!^[ \t]*(do|while|for|if|else|instanceof|new|return|switch|case|throw|catch|using)\n"
 	 /* Methods and constructors */
-	 "^[ \t]*(((static|public|internal|private|protected|new|virtual|sealed|override|unsafe)[ \t]+)*[][<>@.~_[:alnum:]]+[ \t]+[<>@._[:alnum:]]+[ \t]*\\(.*\\))[ \t]*$\n"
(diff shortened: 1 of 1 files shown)


petervanderdoes:
#### Please close the request.

From the [README](https://github.com/git/git/blob/master/README)
> The user discussion and development of Git take place on the Git
> mailing list -- everyone is welcome to post bug reports, feature
> requests, comments and patches to git@vger.kernel.org (read
> Documentation/SubmittingPatches for instructions on patch submission).

Also the tagline of the repo here says:
>Git Source Code Mirror - This is a publish-only repository and all pull requests are ignored. Please follow Documentation/SubmittingPatches procedure for any of your improvements.

thomaslevesque:
Just went over the submission guidelines, and it's definitely not worth the trouble. Is there any reason why you make it so hard to contribute? I guess it's a good way to weed out people that are not serious enough about contributing to git, but it also discourages casual contributors...

peff:
It is not a deliberate attempt to make things hard. It is that none of the development happens on GitHub at all. This repository is purely a mirror. Dealing with pull requests here means that the maintainer has to have two separate workflows, taking in patches both from the mailing list and from here. It's more work for him.

If your next question is "why not abandon the mailing list and just use GitHub", that would be a much longer discussion. :)

thomaslevesque:
> If your next question is "why not abandon the mailing list and just use GitHub", that would be a much longer discussion. :)

I admit that the question crossed my mind ;)

Using a mailing list in 2015 seems a little old-school to me, but I assume there are good reasons for this...

PR #127, opened 2015-02-09
Fix a bug in bash autocompletion

MKorostoff:
There is a bug in bash completions that makes it so that git remotes cannot be listed properly.  When listing git remotes, the wrong directory is specified, namely `.git/refs/remotes` is wrongly configured as `.git/remotes` (which I believe was the location of this directory in a previous version of git).  This results in an error in which typing `git push or` + `tab` prints out `ls: .git/remotes: No such file or directory`.  Once this pull request is merged, git remotes will have proper autocompletion in bash once again.

## before this pull request:
gitbad

## after this pull request:
gitgood

### DIFF
--- a/contrib/completion/git-completion.bash
+++ b/contrib/completion/git-completion.bash
@@ -412,7 +412,7 @@ __git_refs_remotes ()
 __git_remotes ()
 {
 	local i IFS=$'\n' d="$(__gitdir)"
-	test -d "$d/remotes" && ls -1 "$d/remotes"
+	test -d "$d/refs/remotes" && ls -1 "$d/refs/remotes"
 	for i in $(git --git-dir="$d" config --get-regexp 'remote\..*\.url' 2>/dev/null); do
 		i="${i#remote.}"
 		echo "${i/.url*/}"


peff:
Please see the SubmittingPatches file. Development on git happens on the mailing list, and you'll need to send a patch there to be considered for inclusion.

I took a quick look at your patch, though, and I do not think it is doing the right thing. The directory `.git/remotes` is correct. It is a historical storage format for the list of remotes, one per file (this was later moved into the config file). The completion checks both locations.

Note that before the completion code runs `ls $d/remotes`, we check that `test -d $d/remotes` is true. This should prevent the error message you are seeing. I'm not sure why it is not in your case; I was not able to reproduce the problem locally.

MKorostoff:
Thanks for the reply @peff !  The difference in reproducibility *might* be related to OS or git version.  I'm not running anything too exotic (git 1.7, bash 3.2, osx 10.8) but I take you at your word that this code is w.a.d. for most users.

So perhaps a safer solution than changing the file path would be to simply send stderror on this line to dev/null, like so:

```shell
__git_remotes ()
{
	local i IFS=$'\n' d="$(__gitdir)"
	test -d "$d/remotes" && ls -1 "$d/remotes" 2>/dev/null #<----- fixed!
	for i in $(git --git-dir="$d" config --get-regexp 'remote\..*\.url' 2>/dev/null); do
		i="${i#remote.}"
		echo "${i/.url*/}"
	done
}
```

I'll package this up as a patch and submit to the list serve.  Thanks again!

peff:
Yeah, I don't see a problem with adding the redirect as an extra precaution against cruft from `ls` (e.g., the file might be racily removed, or we might not have permission to access it, or whatever).

MKorostoff:
@peff I'm pretty confused by the [SubmittingPatches](https://github.com/git/git/blob/master/Documentation/SubmittingPatches) document.  In particular the part about how to email your patches (include the patch inline, but don't copy/paste?)

This might be a bit of a big ask, but do you skype?  My user name is MKorostoff, could really use a bit of help here

rumpelsepp:
Maybe this helps: http://git-scm.com/docs/git-send-email

MKorostoff:
thanks @rumpelsepp !  I was able to submit my patch :+1:

PR #130, opened 2015-02-28
l10n: fr add yes/no hint to question

jnavila:
As discussed on the mailing list, I added a hint to the french translation that the question should be replied with yes/no.


jnavila:
Ooops... wrong repo. Sorry.

PR #131, opened 2015-03-08
git-prompt: non recursive ls to find if there are untracked file

reubenpeeris:
When identifying if there are any untracked files there is no need to do
a recursive search. If there are many untracked directories the
recursive search can be very slow.

### DIFF
--- a/contrib/completion/git-prompt.sh
+++ b/contrib/completion/git-prompt.sh
@@ -487,7 +487,7 @@ __git_ps1 ()
 
 		if [ -n "${GIT_PS1_SHOWUNTRACKEDFILES-}" ] &&
 		   [ "$(git config --bool bash.showUntrackedFiles)" != "false" ] &&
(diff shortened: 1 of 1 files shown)


rumpelsepp:
Please submit this to the mailing list instead:
https://github.com/git/git/blob/master/Documentation/SubmittingPatches#L137

contrib/completion/git-prompt.sh
```diff
@@ -487,7 +487,7 @@ __git_ps1 ()
```
angigail: .mp3

gitster:
Sounds sensible from a cursory look; care to submit a properly formatted patch to git@vger.kernel.org, following Documentation/SubmittingPatches?

Thanks.

gitster:
No activity for a while since the last suggestion. Closing it.
Please do not hesitate to send a polished patch to git@vger.kernel.org for proper reviews and discussion.

Thanks.

PR #132, opened 2015-03-11
Document prefetching side-effect of "git fetch --dry-run"

orent:
git fetch --dry-run has a useful side-effects of prefetching remote objects. When using a remote over a slow link this is useful and may be safely executed in the background without any unwanted effects. Prefetching will significantly speed up a subsequent git fetch.

### DIFF
--- a/Documentation/fetch-options.txt
+++ b/Documentation/fetch-options.txt
@@ -29,7 +29,10 @@ the current repository has the same history as the source repository.
 
 ifndef::git-pull[]
 --dry-run::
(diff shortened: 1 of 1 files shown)


sgoblin:
All pull requests are ignored on this repository.

rumpelsepp:
https://github.com/git/git/blob/master/Documentation/SubmittingPatches

gitster:
The change makes sense, I think. Please send it to git@vger.kernel.org with a sign-off (see URL rumpelsepp gave in the comment above).

PR #133, opened 2015-03-20
Lines of context

kgoess:
Other products like Atlassian's Fisheye or TracBrowser have the option when looking at diffs on their UI to select how many lines of context to show with the diff, a la "git show -U20 f16c70c4". 

This patch sticks a selectbox onto the UI with a javascript trigger when it changes, and accepts a "u=20" url parameter. The UI looks like this

git-context-lines

We've been using this for a while, and it's pretty handy.


rumpelsepp:

https://github.com/git/git/blob/master/Documentation/SubmittingPatches

gitster:
Yup. What rumpelsepp said. The goal and UI approach both sound sane, though.

PR #134, opened 2015-04-01
Fix svn clone issue with tag names containing colons

daniel-sc:
As observed e.g. here: https://groups.google.com/forum/#!topic/git-users/XDFXUPZKrqk (at least under windows) git-svn has a problem if svn tag names contain special characters that are not allowed in os folder names (e.g. colon).

### DIFF
--- a/perl/Git/SVN.pm
+++ b/perl/Git/SVN.pm
@@ -2362,7 +2362,9 @@ sub _new {
 		             "refs/remotes/$prefix$default_ref_id";
 	}
 	$_[1] = $repo_id;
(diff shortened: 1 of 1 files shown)


daniel-sc:
I won't read through the (rather lengthy) document on sending pathes to mailing lists. So anyone please feel free to send this patch to some mailing list or whatever process is correct!

gitster:
Is this limited to tag names? Doesn't the same issue exist for branch names?

After munging the subversion tag name (or branch name for that matter) like this patch does, would it not cause trouble when changes need to be pushed back via dcommit etc.?

In any case, the patch needs to be discussed on the list and with area experts like Eric Wong.

PR #135, opened 2015-04-07
fix global bash variable leak on __gitcompappend

marcioAlmada:
How to reproduce:

```bash
__gitcompappend "something"
echo $x # outputs 'something'
```

Or more indirectly:

```bash
git checkout [tab-tab]
echo $x # outputs the name of the last branch on completion list
```

No big deal but it's annoying to know that `$x` is lurking :laughing:

### DIFF
--- a/contrib/completion/git-completion.bash
+++ b/contrib/completion/git-completion.bash
@@ -186,7 +186,7 @@ fi
 
 __gitcompappend ()
 {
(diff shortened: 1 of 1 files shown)


marcioAlmada:
Oh, this repository is read only. I'll go through the mailing list.

PR #137, opened 2015-05-20
include final slash in ~/.zsh/_git/

durantschoon:
Indicating that _git is a folder and not a file might save a new zsh user a world of pain. I learned this one the hard way.

Signed-off-by: B. Durant Schoon <durant.schoon@gmail.com>

### DIFF
--- a/contrib/completion/git-completion.zsh
+++ b/contrib/completion/git-completion.zsh
@@ -11,7 +11,7 @@
 #
 #  zstyle ':completion:*:*:git:*' script ~/.git-completion.zsh
 #
-# The recommended way to install this script is to copy to '~/.zsh/_git', and
+# The recommended way to install this script is to copy to '~/.zsh/_git/', and
 # then add the following to your ~/.zshrc file:
 #
 #  fpath=(~/.zsh $fpath)


rtyley:
Hi @durantschoon. At the moment, the Git project unfortunately does _not_ accept patches by pull request - you have to send them to the mailing list. The process for doing this is [surprising involved](https://github.com/git/git/blob/1eb0545c/Documentation/SubmittingPatches#L120-L464) - the fact that the process is so unfamiliar & fiddly for the majority of Git users was discussed as a problem at the Git Contributors conference about a month ago. We weren't able to reach agreement on a change, but a suggestion was made that perhaps a bridge could be created from GitHub pull requests to the mailing list.

This is what I'm trying to do. So, **I'm really glad you've opened this PR**, because I'd like you to try out the new service, called [**_submitGit_**](https://submitgit.herokuapp.com/).

You login to the service using GitHub OAuth, select your PR, get submitGit to send it to you in order to check it looks ok, then, when you're ready, get submitGit to send it on to the mailing list.

You would be the first person to use the service for real, and I'd appreciate your feedback. Bare in mind there are a couple of things wrong with your submission (commit message line length longer than 72 chars, and no signed-off-by at the end), so you'd need to fix those before sending them to the mailing list (obviously, they're things that _submitGit_ should, and eventually will, warn you about).

Let me know what you think - or if you prefer the current submission process, that would really interest me.

thanks,
Roberto

_Incidentally, I'm a [developer at the Guardian](http://www.theguardian.com/profile/roberto-tyley), and do a bit of [Git stuff](https://rtyley.github.io/bfg-repo-cleaner/) - nice to meet you_

gitster:
@rtyley 
Perhaps submitGit can gain a feature where trusted users or reviewers can send out e-mails on other people's behalf, as long as the original contributor signed-off the patch correctly already in their commit?

I do not know if this commit is signed-off (or GitHub UI hiding it), so the above may not apply to
this one, though.

yiqiejieyoukeneng:
00051f1

yiqiejieyoukeneng:
00051f1

rtyley:
> Perhaps submitGit can gain a feature where trusted users or reviewers can send out e-mails on other people's behalf, as long as the original contributor signed-off the patch correctly already in their commit?

@gitster I think it's probably best that we encourage people to submit their own PRs to the list (rather than have trusted reviewers adopt them), even if they are signed-off - hopefully there shouldn't be any reason why PR creators can't do that?

Abandonment is sad, but what we _could_ do regarding that is this: add a webhook callback to the git/git GitHub repository settings, firing whenever a pull request is opened - it would just trigger submitGit to immediately make a guiding comment on the PR, leading the user to https://submitgit.herokuapp.com/, so they can correctly submit the change while still focused on the task.
 
> I do not know if this commit is signed-off (or GitHub UI hiding it), so the above may not apply to
> this one, though.

This one's not signed off as it happens, you can see the full text of the _commit_ (which is a different field to the _pull request description_, tho' the PR desc defaults to using the text of the commit message) here: https://github.com/durantschoon/git/commit/0f19473fb35ffd4624d25fcf807202432a120938 ... GitHub doesn't hide any kind of footer, including 'Signed-off-by', thankfully.

ThiefMaster:
To be honest, I **passionately hate** using mailing lists. Especially if I have to subscribe to the list to post to it. Especially if I submit a small patch somewhere chances are good that I **do not care** about all the unrelated development discussion, "pull requests", etc. happening on that list. So I can understand everyone who prefers a quick GitHub PR over the hassle of submitting a patch on a mailing list.

rumpelsepp:
>  hassle of submitting a patch on a mailing list.

```
git format-patch HEAD^1
git send-email --to="git@vger.kernel.org" *.patch
```
Where is the hassle? It is even faster than creating such pull requests, where I am supposed to create a fork and hit the green "pull request button" in the web interface…

…and you don't need to subscribe to the mailing list; just send your patch. It's two commands. :)

louy2:
What configuration does `git send-email` need? I don't believe git can send email by itself since it's not an SMTP client.

ThiefMaster:
a sendmail-style tool i guess

rumpelsepp:
in your `.gitconfig`:

```
[sendemail]
	smtpencryption = tls
	smtpserver = smtp.myserver.com
	smtpuser = user@myserver.com
	smtpserverport = 587
```

stefanbeller:
+ThiefMaster

You don't have to subscribe to the mailing list to send patches (at least for git mailing list this is true).
Also that's the whole point of the submitGit app so you can do a pull request and the app turns it into a proper mailing list submission.

james69miller:
N vivo

dscho:
@durantschoon I see that all these years later, this patch would still apply.

Please wrap your commit message to <=76 columns per line, add your [sign off](https://git-scm.com/docs/SubmittingPatches#dco) and send the patch to the mailing list for review. You can use [GitGitGadget](https://gitgitgadget.github.io), [submitGit](https://submitgit.herokuapp.com) or [send it manually](https://github.com/git-for-windows/git/blob/master/CONTRIBUTING.md#submit-your-patch).

durantschoon:
Closing this, hopefully https://github.com/git/git/pull/609 will take it's place.

durantschoon:
its

PR #143, opened 2015-06-01
msvc-build: fix the broken msvc-build scripts (test)

PhilipOakley:
This sequence fixes the msvc-build script that generates
the Visual Studio 2008 git.sln file.

The inclusion of the msvc-build script still has nuances because of
it's location.

The later commits are rough drafts of the change to the Makefile
allowing the installation of the VS2008 build products. They don't
work yet.

This test rebase onto master is so that the patch submission
process can be tested.
(https://github.com/git/git.github.io/blob/master/rev_news/drafts/edition-4.md)

Philip


PhilipOakley:
This is re-arranged with some improvements to the commit messages. While it is the same length it should now be able to be processed by GitSubmit.

dscho:
Maybe we can close this in favor of https://github.com/gitgitgadget/git/pull/149 (which I am about to send in the vague direction of the Git mailing list)?

PhilipOakley:
closing this in favour of gitgitgadget#149

PR #144, opened 2015-06-01
msvc-build short (test)

PhilipOakley:
The first ~18 commits of the fixing of the msvc-build script which generated
Visual Studion 2008 .sln and .vcproj files.

The series is short to fit within the 'patch submission' limit (20 commits) on
https://submitgit.herokuapp.com/


rtyley:
@PhilipOakley can you tell me how long the patch series would normally be? I'm happy to increase the patch submission limit if it's not too huge.

PhilipOakley:
@PhilipOakley can you tell me how long the patch series would normally be? I'm happy to increase the patch submission limit if it's not too huge.


In this case it was 21 commits! (pull 143). One thought I had was you could raise the limit for sending to self to double (i.e. 40), but leave the git@vger at it's current limit, which would allow folk to test their series, but break it up for submission & review (with note on the page so folk know that's why;-)

I also had the problem that the patches too a very long time to arrive, so I ended up sending two requests (7 mins apart). The first patch took 51 minutes to traverse the internet, with 18/18 taking just under two hours. The second series was 'similar', again a note on the page would stop folks worrying.

Philip

jasonwilliams200OK:
Wouldn't it make sense to include cmake instead?
CMake will also take care of various versions of Visual Studio and Xcode etc.
https://github.com/kitware/cmake

PhilipOakley:
tl;dr, the purpose is to use VS as the gui (i.e. user facing), not make for Windows.
  Wouldn't it make sense to include cmake instead?
  CMake will also take care of various versions of Visual Studio and Xcode etc.
  https://github.com/kitware/cmake


The reasoning for repairing and bug fixing the '--gui' part of the msvc-build is that it allows (Windows) programmers to explore the git code using a tool they are most likely familiar with (and available in free versions, though not open ;-). That is, produce a git.sln and associated files so that VS2008 (and onward) can visualise and compile the code. I'm sort of one of them!

The current 'make' (as part of the regular Msysgit/SDK/installer) does provide executables suitable for direct use, so that's not 'broken'.

Ultimately, I'd want the current Makefile to have enough switches such that the code produced by VS2008 could be installed and tested, just as if it had been produced by the gcc compiler. With luck and a following wind it will function correctly as well.

I don't see the Windows-SDK project adding another make tool without very good reason. They are very short of developers, so the extra maintenance load, while no doubt small, probably would not be worth it. The SDK is pretty heavy already. It (using cmake) is not something I could take on (it's taken me long enough to get this far with fixing the msvc-build!). I know many folk are very happy with cmake, so it's not a quality problem. [I didn't checked formal licence compatibility either]
--
Philip

dscho:
I am about to send https://github.com/gitgitgadget/git/pull/149 via GitGitGadget.

Maybe we can close this here PR, then?

PhilipOakley:
closing this in favour of gitgitgadget#149

PR #145, opened 2015-06-06
Generate tcsh_completion_script even if 'set noclobber' is in effect

arielf:


rtyley:
Entertainingly, the long subject line on the commit is wrapping and causing submitGit to crash... https://submitgit.herokuapp.com/git/git/pull/145

PR #147, opened 2015-06-08
fsck: don't ignore broken reflog entries

mhagger:
Previously, if a reflog entry's old or new SHA-1 was not resolvable to
an object, that SHA-1 was silently ignored. Instead, report such cases
as errors.


mhagger:
Hmmm, I was going to try sending this with submitGit, but it doesn't seem to support adding a cover letter. Never mind, I'll send it the old-fashioned way.

PR #165, opened 2015-08-04
Update branch.c cmd_branch instructions

kweij:
Textual change in instruction to hint the correct git command: track upstream instead of delete.

Screenshot of the current behaviour:
git-bug-hint-track-upstream

The proposed behaviour instructs to use

``` shell
    git branch -u origin/develop
```

instead of

``` shell
    git branch -d origin/develop
```

to track upstream


musab22161:
تحديث

kweij:
Deprecated

PR #167, opened 2015-08-27
Update Git Credential Docs To Be Correct

WillHipschman:
Without EOF on Win10

E:\OnPrem\P9>git credential fill
10:36:14.353089 git.c:348               trace: built-in: git 'credential' 'fill'
protocol=http
host=localhost:8080
path=tfs/DefaultCollection/_git/P3.git/info/lfs/objects

warning: invalid credential line:
fatal: unable to read credential from stdin

With EOF on Win10

E:\OnPrem\P9>git credential fill
protocol=http
host=localhost:8080
path=tfs/DefaultCollection/_git/P3.git/info/lfs/objects
^Z
Fatal: NullReferenceException encountered.
': s/DefaultCollection/_git/P3.git/info/lfs/objects
': s/DefaultCollection/_git/P3.git/info/lfs/objects
protocol=http
host=localhost:8080
path=tfs/DefaultCollection/_git/P3.git/info/lfs/objects
username=
password=

### DIFF
--- a/Documentation/git-credential.txt
+++ b/Documentation/git-credential.txt
@@ -54,7 +54,7 @@ credential` following these steps:
 +
 For example, if we want a password for
 `https://example.com/foo.git`, we might generate the following
-credential description (don't forget the blank line at the end; it
+credential description (don't forget the blank line and EOF at the end; it
 tells `git credential` that the application finished feeding all the
 information it has):
 


peff:
>     E:\OnPrem\P9>git credential fill
>     10:36:14.353089 git.c:348 trace: built-in: git 'credential' 'fill'
>     protocol=http
>     host=localhost:8080
>     path=tfs/DefaultCollection/_git/P3.git/info/lfs/objects
>     
>     warning: invalid credential line:
>     fatal: unable to read credential from stdin

I suspect this is due to CRLF in your input.

The `^Z` you send is probably being interpreted as EOF by your libc's stdio routines. It's not meaningful on non-Windows platforms, and it would generally be a bad thing to add:

    $ git credential fill
    url=https://example.com
    ^Z
    warning: invalid credential line: 
    fatal: unable to read credential from stdin

I do think there is a problem in the documentation, though, which is that it claims you need the blank line at all. AFAICT that has never been the case, and it is a bug in the documentation.

WillHipschman:
How would you feel about a comment calling out ^Z for windows users? 

As is it expects tribal knowledge--you have to know how line endings work in different systems and it's difficult to find a solution without asking someone who knows.

peff:
You shouldn't really need ^Z either. A true EOF is fine. So for a toy example where you're inputting from a terminal, ^Z is one way of doing that. But in a real application, you would typically have a pipe to `git credential`'s stdin, and you signal EOF by closing the pipe. E.g.:

    echo url=https://example.com | git credential fill

should just work without needing ^Z on any system.

Can you be more clear about what you're trying to accomplish?

peff:
Also, you may want to take this to the [git mailing list](http://git-scm.com/community), which is where most discussion happens.

WillHipschman:
I was trying to debug a git credential fill issue in an application that works by piping input like you describe above.

The error I was receiving was generic, so I wanted to validate that git credential fill was actually returning the credentials I expected. I ran git credential fill from the cli and saw this behavior. 

Piping works from the cli, but there is nothing in the docs to suggest it shouldn't work the way I originally tried it.

I'm not blocked by this--I only suggest the change so that people don't have to hit the same issue I did and spend an hour trying to figure out how to get it to run.

WillHipschman:
I'll close this and follow up on the mail list.

m6j9b12:
Nice family business....still the wrong and disrespectful way...but it is
your legacy afterall

m6j9b12:
Never forget...move past improve ...learn

PR #174, opened 2015-09-10
Makefile: use SHELL_PATH when running generate-cmdlist.sh

asedeno:
Some /bin/sh implementations can't deal with $() arithmetic and command
substitution. If we already have a better shell on hand, we should use it.

Fixes the build on SunOS, probably others.

Signed-off-by: Alejandro R. Sedeño <asedeno@mit.edu>


asedeno:
I'll deal with this in the supported fashion shortly.

PR #176, opened 2015-09-12
poll: honor the timeout on Win32

dscho:
Ensure that when passing a pipe, the gnulib poll replacement will not
return 0 before the timeout has passed.

Not obeying the timeout (and merely returning 0) causes pathological
behavior when preparing a packfile for a repository and taking a
long time to do so.  If poll were to return 0 immediately, this would
cause keep-alives to get sent as quickly as possible until the packfile
was created.  Such deviance from the standard would cause megabytes (or
more) of keep-alive packets to be sent.

GetTickCount is used as it is efficient, stable and monotonically
increasing.  (Neither GetSystemTime nor QueryPerformanceCounter have
all three of these properties.)

Signed-off-by: Edward Thomson <ethomson@microsoft.com>
Signed-off-by: Johannes Schindelin <johannes.schindelin@gmx.de>

### DIFF
--- a/compat/poll/poll.c
+++ b/compat/poll/poll.c
@@ -446,7 +446,7 @@ poll (struct pollfd *pfd, nfds_t nfd, int timeout)
   static HANDLE hEvent;
   WSANETWORKEVENTS ev;
   HANDLE h, handle_array[FD_SETSIZE + 2];
-  DWORD ret, wait_timeout, nhandles;
+  DWORD ret, wait_timeout, nhandles, start = 0, elapsed, orig_timeout = 0;
   fd_set rfds, wfds, xfds;
   BOOL poll_again;
   MSG msg;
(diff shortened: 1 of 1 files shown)


submitgit:
@dscho sent this to the mailing list with [_submitGit_](https://github.com/rtyley/submitgit) - here on [Gmane](http://mid.gmane.org/0000014fc2adfd6a-40802cb4-acb3-4b45-ac12-ce046f6c408b-000000@eu-west-1.amazonses.com), [MARC](http://marc.info/?i=0000014fc2adfd6a-40802cb4-acb3-4b45-ac12-ce046f6c408b-000000@eu-west-1.amazonses.com) [](0000014fc2adfd6a-40802cb4-acb3-4b45-ac12-ce046f6c408b-000000@eu-west-1.amazonses.com)

dscho:
This patch has been accepted: ef8b53e78c534d732fe58a46da8dbd922287a56b

PR #180, opened 2015-10-10
add annotation for two consecutive asterisks

geekpi:
I tried two consecutive asterisks pattern today but failed today. I was confused and tried several ways. Finally I found this pattern is supported since 1.8. So I add this annotation for user to read it more clearly.

### DIFF
--- a/Documentation/gitignore.txt
+++ b/Documentation/gitignore.txt
@@ -113,7 +113,7 @@ PATTERN FORMAT
    For example, "/{asterisk}.c" matches "cat-file.c" but not
    "mozilla-sha1/sha1.c".
 
-Two consecutive asterisks ("`**`") in patterns matched against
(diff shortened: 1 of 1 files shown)


revi:
Read [here](https://github.com/git/git/blob/master/Documentation/SubmittingPatches) to learn how to submit patch for git repository. They don't accept pull requests.

dscho:
@geekpi I don't think that our command manual pages are littered with information when certain options started to be supported.

Having said that, if you still want to get this integrated, please add supporting information to the commit message, add your [sign off](https://git-scm.com/docs/SubmittingPatches#dco), rebase your patch and send the patch to the mailing list for review. You can use [GitGitGadget](https://gitgitgadget.github.io), [submitGit](https://submitgit.herokuapp.com) or [send it manually](https://github.com/git-for-windows/git/blob/master/CONTRIBUTING.md#submit-your-patch).

PR #182, opened 2015-10-24
checkout: progress on non-tty. progress with lf

eantoranz:
Checkout:
Need to be able to monitor (as in from the *process* that we use to spawn git) how the progress of a checkout operation is going. Given that the progress is suppressed if not working with a TTY, it's very difficult to do so. My idea is to start the checkout process and pull the output for analysis of the progress through stderr.

So.... added these two options to checkout:
--progress-no-tty: option to be able to write progress even if not working on a TTY
--progress-lf: option to be able to write LF characters at the end of a progress line instead of CR.

### DIFF
--- a/builtin/checkout.c
+++ b/builtin/checkout.c
@@ -37,6 +37,8 @@ struct checkout_opts {
 	int overwrite_ignore;
 	int ignore_skipworktree;
 	int ignore_other_worktrees;
+	int progress_lf;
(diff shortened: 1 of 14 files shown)


eantoranz:
I have developed a smaller version of this change trying not to affect the other parts not related to checkout. Will close this pull request and create a new one with the other version.

PR #183, opened 2015-10-24
checkout: progress on non-tty. progress with lf

eantoranz:
**checkout**
I need to be able to monitor how the checkout process is going from the process that spawns git. If not working on a tty, it's very difficult to do so. To make life a little bit simpler, I added checkout 2 options to ease progress monitoring. If they are not used, previous git behavior would be used, so not to break current use cases.

**--progress-no-tty**: option to write progress even if not working on a TTY
**--progress-lf**: option to print progress using LF instead of CR

### DIFF
--- a/builtin/checkout.c
+++ b/builtin/checkout.c
@@ -37,6 +37,8 @@ struct checkout_opts {
 	int overwrite_ignore;
 	int ignore_skipworktree;
 	int ignore_other_worktrees;
+	int progress_lf;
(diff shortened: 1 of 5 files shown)


eantoranz:
Ok.... so this repo doesn't accept pull requests. Reading the documentation to know how to submit patches (https://github.com/git/git/blob/master/Documentation/SubmittingPatches).

Cheers!

PR #184, opened 2015-10-28
Fix dcommit when a renamed file has @ in its name

ddobrev:
If a file has '@' in its name and the file is renamed - for example moved to another directory - dcommit-ting to an SVN repository yields this error:

Assertion failed: (svn_uri_is_canonical(child_uri, NULL)), function uri_skip_ancestor, file subversion/libsvn_subr/dirent_uri.c, line 1520.
error: git-svn died of signal 6

### DIFF
--- a/perl/Git/SVN/Editor.pm
+++ b/perl/Git/SVN/Editor.pm
@@ -145,7 +145,7 @@ sub url_path {
 	my ($self, $path) = @_;
 	if ($self->{url} =~ m#^https?://#) {
 		# characters are taken from subversion/libsvn_subr/path.c
-		$path =~ s#([^~a-zA-Z0-9_./!$&'()*+,-])#sprintf("%%%02X",ord($1))#eg;
+		$path =~ s#([^~a-zA-Z0-9_./!$&'()*+,@-])#sprintf("%%%02X",ord($1))#eg;
 	}
 	$self->{url} . '/' . $self->repo_path($path);
 }


abruzzihraig:
I got the same issue, pls merge the PR. Thanks

salabaha:
Just run in to the same issue. Really frustrating.

onglai:
#184

perl/Git/SVN/Editor.pm
```diff
@@ -145,7 +145,7 @@ sub url_path {
 	my ($self, $path) = @_;
 	if ($self->{url} =~ m#^https?://#) {
 		# characters are taken from subversion/libsvn_subr/path.c
-		$path =~ s#([^~a-zA-Z0-9_./!$&'()*+,-])#sprintf("%%%02X",ord($1))#eg;
+		$path =~ s#([^~a-zA-Z0-9_./!$&'()*+,@-])#sprintf("%%%02X",ord($1))#eg;
 	}
 	$self->{url} . '/' . $self->repo_path($path);
```
bdewitt1991: View

dscho:
@ddobrev this patch seems to apply, still.

Sadly, the Git project does not accept PRs, so you will have to use either [GitGitGadget](https://gitgitgadget.github.io), [submitGit](https://submitgit.herokuapp.com) or [send it manually](https://github.com/git-for-windows/git/blob/master/CONTRIBUTING.md#submit-your-patch).

Before that, though, please note that you are asked to wrap the commit message at <=76 columns per line, and add your [sign off](https://git-scm.com/docs/SubmittingPatches#dco).

Further, the comment above the line changed in this PR seems to refer to:
https://github.com/apache/subversion/blob/trunk/subversion/libsvn_subr/path.c

But I cannot find anything in that file that merits adding the `@` character to that range... Maybe you can dig deeper a little?

dscho:
@ddobrev gentle ping?

ddobrev:
Hello @dscho. I apologize for being forced to disappoint you but I no longer have either the time or the need to work on this piece. The project I used git-svn for has long concluded and while the problem is real, I don't have the possibility any more to move my PR to some other system, let alone investigate further per your suggestion. I can tell you this much about back then: I have been able to reliably reproduce the problem, I have described it, I have suggested a fix and I have tested it as working. In case you find my code of substandard quality, you can think of it as just a reference implementation. However, at the time of writing this I all the assistance I'm able to offer.

dscho:
Okay. Feel free to close your PRs in the future once you are unable to work on them anymore.

PR #192, opened 2016-01-25
added check for zlib.h to configure.ac

krichter722:


dscho:
Is this not covered by the subsequent `AC_LANG_PROGRAM([#include <zlib.h>], ...`?

If you still think that this patch should make it in, please do describe the rationale in the commit message, add your [sign off](https://git-scm.com/docs/SubmittingPatches#dco) and send the patch to the mailing list for review. You can use [GitGitGadget](https://gitgitgadget.github.io), [submitGit](https://submitgit.herokuapp.com) or [send it manually](https://github.com/git-for-windows/git/blob/master/CONTRIBUTING.md#submit-your-patch).

PR #194, opened 2016-01-28
userdiff: Add 'interface' support to PHP

mamchenkov:
PHP supports interfaces since version 5.
http://php.net/manual/en/language.oop5.interfaces.php


dscho:
This was addressed by 1ab631647e3addd247c5482e8a3d4c9df7978c62 in the meantime.

PR #196, opened 2016-02-10
Fix misseplling when adding submodule

gregorhollmig:
"Namenmit" are actually two words "Namen" and "mit"


rtyley:
@gregorhollmig although the official patch submission process for Git is based on emailed patches, the Git project _does_ accept GitHub PRs that have been sent through https://submitgit.herokuapp.com/ - you can give that a try if you like - assuming this fix is valid?

gregorhollmig:
thanks, I will try that!

PR #197, opened 2016-02-11
Match synopsis to description of man git-diff

Hotschke:
I find it rather confusing when the description of the synopsis is in a different order and the number of items do not match.

### DIFF
--- a/Documentation/git-diff.txt
+++ b/Documentation/git-diff.txt
@@ -9,19 +9,22 @@ git-diff - Show changes between commits, commit and working tree, etc
 SYNOPSIS
 --------
 [verse]
-'git diff' [options] [<commit>] [--] [<path>...]
+'git diff' [options] [--] [<path>...]
+'git diff' [options] --no-index [--] [<path>...]
 'git diff' [options] --cached [<commit>] [--] [<path>...]
-'git diff' [options] <commit> <commit> [--] [<path>...]
+'git diff' [options] <commit> [--] [<path>...]
+'git diff' [options] <commit>..<commit> [--] [<path>...]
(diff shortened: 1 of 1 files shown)


Documentation/git-diff.txt
```diff
@@ -9,19 +9,22 @@ git-diff - Show changes between commits, commit and working tree, etc
...
 
 DESCRIPTION
 -----------
-Show changes between the working tree and the index or a tree, changes
-between the index and a tree, changes between two trees, changes between
-two blob objects, or changes between two files on disk.
+Show (i) changes between the working tree and the index or a tree,
+(ii) changes between two files on disk, (iii) changes between the 
+index and a tree, (iv) changes between two trees, or (v) changes
+between two blob objects.
```
CiaranReen: Can we not have these on separate lines for more clarity?

rtyley:
Although the official patch submission process for Git is based on emailed patches, the Git project does accept GitHub PRs that have been sent through https://submitgit.herokuapp.com/ - you can give that a try if you like

Documentation/git-diff.txt
Hotschke: Well, I guess it depends on what people prefer. For instance, if you have wider screen should this be only on one to two lines or occupy always more screen height.  I tend to keep it in one paragraph and reflow according to the available screen width. I admit that inline numbering is not pretty.
Therefore, I would remove my proposed inline numbering. I just would reorder the parts to match order of synopsis and description.

However, since github pull request are  not the place to propose changes. I will close this pull request. When I find the time I will submit to mailing list.

Hotschke:
@rtyley Thanks for the heads up. I will close it and will use the mailing list when I find the time for it.

PR #198, opened 2016-02-14
fix bash completion for "git -C <path>"

hanyong:
take care "-C" option from "__git_main",
fix "__git_dir" relative to "-C <path>".
make completion work for
"git -C xxx ", "git -C xxx checkout " and so on.

### DIFF
--- a/contrib/completion/git-completion.bash
+++ b/contrib/completion/git-completion.bash
@@ -2571,7 +2571,7 @@ _git_whatchanged ()
 
 __git_main ()
 {
-	local i c=1 command __git_dir
+	local i c=1 command __git_dir __cwd
 
 	while [ $c -lt $cword ]; do
 		i="${words[c]}"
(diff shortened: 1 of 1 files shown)


dscho:
This is still a feature that Git lacks, so maybe I can ask you to rebase it and submit it to the Git mailing list for review?

One important thing before that, though, is that the reviewers on the Git mailing list are *very* picky about commit messages. So please allow me to try my hand at modifying the commit message in a way that I think will be accepted more readily:

```
completion: handle `git -C <path> --git-dir=<directory>` correctly

The combination of `-C` and `--git-dir` needs to be handled specially
when the latter specifies a relative path: Git would first change the
current working directory to the path indicated via `-C`, and the Git
directory would be relative to *that* directory.

One interesting quirk is that the `-C` option may come after
the `--git-dir` option, or that either option is passed multiple times.

Therefore, we must be careful to remember the values and to combine them
only after we saw all options.

Signed-off-by: 涧泉 <observer.hany@alibaba-inc.com>
```

You will notice a couple of things:
- The commit message is wrapped at <=76 columns per line.
- The first line is like a super-short summary, starting with the prefix "completion:".
- The commit message explains why the `__cwd` variable is necessary, without actually talking about the actual implementation details.
- Finally, there is a [sign off](https://git-scm.com/docs/SubmittingPatches#dco) (which is required by the Git project).

Feel free to use this commit message verbatim when rebasing.

After you rebased this to the current `master`, please send the patch to the mailing list for review. You can use [GitGitGadget](https://gitgitgadget.github.io), [submitGit](https://submitgit.herokuapp.com) or [send it manually](https://github.com/git-for-windows/git/blob/master/CONTRIBUTING.md#submit-your-patch).

hanyong:
@dscho Thank you very much for educate me a lot about submit patches to Git.

After try to rebase the commit to current `master`, I found that `-C` and `--git-dir` already been take into account (for commit [80ac0744b180f815aca190059705c0aed80d16f9](https://github.com/git/git/commit/80ac0744b180f815aca190059705c0aed80d16f9) ), so I will close this PR .

PR #199, opened 2016-02-22
Update diff-highlight

PeterDaveHello:
Use `#!/usr/bin/env perl` instead of `#!/usr/bin/perl` so that it can works on FreeBSD


PeterDaveHello:
#200

PR #201, opened 2016-02-23
push: shorten "push.default is unset" warning message

moy:
The warning was purposely long, both to explain the situation properly
and to give a strong incentive to set push.default explicitly. This was
important before the 2.0 transition, and remained important for a while
after, so that new users get push.default explicitly in their
configuration and do not experience inconsistent behavior if they ever
used an older version of Git.

The warning has been there since version 1.8.0 (Oct 2012), hence we can
expect the vast majority of current Git users to have been exposed to
it, and most of them have already set push.default explicitly. The
switch from 'matching' to 'simple' was planned for 2.0 (May 2014), but
actually happened only for 2.3 (Feb 2015).

The warning is mostly seen by beginners, who have not set their
push.default configuration (yet). For many of them, the warning is
confusing because it talks about concepts that they have not learned and
asks them a choice that they are not able to make yet. See for example

  http://stackoverflow.com/questions/13148066/warning-push-default-is-unset-its-implicit-value-is-changing-in-git-2-0

(1260 votes for the question, 1824 for the answer as of writing)

Shorten the warning and mention only the way to remove the warning
without changing the behavior. Keep a pointer to the documentation so
that people willing to learn can still find the alternative behaviors
easily.

Eventually, the warning should be removed completely, but this can wait
a couple more releases or years.

Signed-off-by: Matthieu Moy <Matthieu.Moy@imag.fr>

### DIFF
--- a/Documentation/git-push.txt
+++ b/Documentation/git-push.txt
@@ -37,6 +37,13 @@ the default `<refspec>` by consulting `remote.*.push` configuration,
 and if it is not found, honors `push.default` configuration to decide
 what to push (See linkgit:git-config[1] for the meaning of `push.default`).
 
+When neither the command-line nor the configuration specify what to
+push, the default behavior is used, which corresponds to the `simple`
+value for `push.default`: the current branch is pushed to the
+corresponding upstream branch, but as a safety measure, the push is
+aborted if the upstream branch does not have the same name as the
+local one.
+
 
 OPTIONS[[OPTIONS]]
 ------------------
--- a/builtin/push.c
+++ b/builtin/push.c
@@ -204,37 +204,6 @@ static void setup_push_current(struct remote *remote, struct branch *branch)
 	add_refspec(branch->name);
 }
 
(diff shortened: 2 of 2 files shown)


submitgit:
@moy sent this commit (https://github.com/git/git/compare/70bd996071bf724a7b151c81a165e8837179d6d7...9680a260438c7877d54e7b242b6109a4f64ad1e8) as a patch to the mailing list with [_submitGit_](https://github.com/rtyley/submitgit) - here on [Gmane](http://mid.gmane.org/000001530ea408ed-2b71a34a-32bb-434c-bba5-fdac28193e9c-000000@eu-west-1.amazonses.com), [MARC](http://marc.info/?i=000001530ea408ed-2b71a34a-32bb-434c-bba5-fdac28193e9c-000000@eu-west-1.amazonses.com) [](000001530ea408ed-2b71a34a-32bb-434c-bba5-fdac28193e9c-000000@eu-west-1.amazonses.com)

PR #202, opened 2016-02-23
activate diff.renames by default

moy:
I have always wondered why diff.renames was not activated by default.
I've had it to `true` in my configuration for 9 years, and I've been
teaching newbies to set it for a while without issue. I think it's
time to activate it by default, but please let me know if I missed a
reason to keep it to false.

In any case, the first 3 patches are useful cleanups.

### DIFF
--- a/Documentation/diff-config.txt
+++ b/Documentation/diff-config.txt
@@ -111,7 +111,7 @@ diff.renames::
 	Whether and how Git detects renames.  If set to "false",
 	rename detection is disabled. If set to "true", basic rename
 	detection is enable.  If set to "copies" or "copy", Git will
-	detect copies, as well.  Defaults to false.  Note that this
+	detect copies, as well.  Defaults to true.  Note that this
 	affects only 'git diff' Porcelain like linkgit:git-diff[1] and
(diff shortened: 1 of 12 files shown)


SHwareSystemsDev:
A reason I would leave it as false is: given two files in a working tree with sufficiently similar content, a diff to a prior commit where only a base file is present may score a copy with fewer changes to the base higher than a rename with more changes, and falsely report what was historically the copy as the rename instead.

As these operations can occur as a batch inbetween distinct add/commit operations the repository and hosting file system may not record enough information to make any detection heuristic reliable. Granted, for many workflows it doesn't matter which came first so the option could default to enabled, but to make a precise audit trail more likely leaving it disabled I'd think is preferable.

moy:
@SHwareSystemsDev : please use the mailing-list for discussion:

http://thread.gmane.org/gmane.comp.version-control.git

I don't understand your objection about copies. My patch does not activate copies detection by default, just renames. If a file exists with the same name in the base commit, then Git will not activate rename detection.

moy:
Merged in `next` by Junio. Let's close the PR on GitHub.

PR #206, opened 2016-03-10
Gitweb snapshot selective

jimklimov:
A series of commits that add:
* ability to specify files/dirs for gitweb snapshot (tarball just a subset of a repo)
* debugging via permission on server side and URI argument
* unit-testing of the above


jimklimov:
FYI : the 6-patch set was submitted by e-mail using hints from https://github.com/git/git/pull/137

dscho:
I see that https://public-inbox.org/git/1457702689-9084-1-git-send-email-jim@jimklimov.com/#t never saw any replies.

My best guess is that this is because the commit messages look more like Subversion-style oneliners than like the commits you usually find in the Git history.

See for example https://public-inbox.org/git/20190604172238.GJ4641@comcast.net/T/#m47da28d57215cc937e202da0d81a6206b185d771. As you will readily recognize the commit message there makes kind of a case why this is a desirable change. It uses only one paragraph to do so, but there are plenty of examples in Git's commit history of commit messages that give quite a bit more background and motivation to the patches.

You can also see that this patch series of two patches was introduced with a cover letter that explains the broader structure, and I would suggest that also for your patches.

Also, you probably noticed that all patches have a [sign off](https://git-scm.com/docs/SubmittingPatches#dco), please do the same.

It may have been a busy time back then, too, maybe you want to wait until v2.22.0 is released (according to https://tinyurl.com/gitCal it is expected this Friday) and send a new iteration?

dscho:
@jimklimov v2.22.0 was released three weeks ago. Maybe now would be a good time?

jimklimov:
Hello, sorry this sort of fell out of the loop in the busy mailbox :)
I'll try to update the PR formalities soon.

jimklimov:
For now, rebased over current master and updated with Signed-off-by tags.

I'll have to go over the commits and re-push later to make sense of what and why I wrote years ago ;)

jimklimov:
Reviewed (and rearranged a bit) the commit contents and updated their descriptions, although rather following the reasonable Problem/Solution pattern promoted in several ecosystems I'm in nowadays.

jimklimov:
Unit-tests for these changes passed locally with `./configure && make && make -C t/ gitweb-test` however I see some failures like on CI for the `make test`... I assume the upstream I based against was a bit broken (some tests are even commented as known breakage).

jimklimov:
Had a few iterations above for documentation - asciidoc not markdown, no `syntax-highlight` tool in build env, etc.

dscho:
>  I see some failures like on CI for the `make test`...

I looked at the logs, and it seems to have been a (hopefully intermittent) timeout while getting the Homebrew cask. I restarted the build.

jimklimov:
In Travis I see this for one scenario:
````
Test Summary Report
-------------------
t9502-gitweb-standalone-parse-output.sh          (Wstat: 256 Tests: 13 Failed: 1)
  Failed test:  13
  Non-zero exit status: 1
  Parse errors: No plan found in TAP output
t7405-submodule-merge.sh                         (Wstat: 0 Tests: 18 Failed: 0)
  TODO passed:   17
Files=872, Tests=20633, 526 wallclock secs ( 9.49 usr  1.89 sys + 686.64 cusr 299.85 csys = 997.87 CPU)
Result: FAIL
Makefile:52: recipe for target 'prove' failed
````

with Linux /  Xcode: xcode10.1 C / jobname=GIT_TEST_GETTEXT_POISON

but have little idea what that means so far.

Seems I can reproduce locally however, with
````
~/git/t$ GIT_TEST_GETTEXT_POISON=YesPlease make clean t9502-gitweb-standalone-parse-output.sh
````
so will try to figure it out.

**Comment by jimklimov** [comment]
Ok, so it seems the new tests which expect `...bad path - empty output and/or HTTP-404` send requests that instead return a `Status: 500 Internal Server Error` when `GIT_TEST_GETTEXT_POISON=YesPlease`. Without the latter, server does return 404 as expected.

**Comment by jimklimov** [comment]
Looking at how this is my fault... per https://github.com/git/git/blob/master/git-sh-i18n.sh#L66

**Comment by jimklimov** [comment]
Yes... so it `500 - Failed to transmit output from git-archive&lt;br/&gt;&lt;pre&gt;fatal: # GETTEXT POISON #<span class="cntrl">\n</span>&lt;/pre&gt;&lt;br/&gt;` probably due to comparison at https://github.com/git/git/pull/206/files#diff-00703a794a540acf45e225abd6aeda3bR7512

I wonder how else (better) such parsing can be handled... would it help (certainly would be "more correct" than what I PR'ed now) to export locale settings into `$cmd` at https://github.com/git/git/pull/206/files#diff-00703a794a540acf45e225abd6aeda3bR7451 ?

**Comment by jimklimov** [comment]
Forcing `LANG=C LC_ALL=C LANGUAGE=C` into it in the web-server might not hurt, but did not help against poisoning...

**Comment by dscho** [comment]
It's curious how the Azure Pipeline passes, though. Any idea why?

**Comment by jimklimov** [comment]
Not sure. Miggt run a differebt set of tests? Per test script, poisoned gettext is quite OS-dependent.

I now lean towards turning a blind eye and accepting both HTTP 404 and 500 as valid outcomes for test, lacking a way to determine error type or details other than by parsing error texts.

**Comment by dscho** [comment]
> Not sure.

Well, you could have a look ;-) From the output of https://travis-ci.org/git/git/jobs/565980864 (you have to click open the "ci/print-test-failures.sh" part, then search for "not ok"):

```
------------------------------------------------------------------------
test-results/t9502-gitweb-standalone-parse-output.out...
------------------------------------------------------------------------
[...]
+rm -f gitweb.body file_list
+BRANCH=master
+gitweb_run p=.git;a=snapshot;h=master;sf=tar;f=third
+GATEWAY_INTERFACE=CGI/1.1
+HTTP_ACCEPT=*/*
+REQUEST_METHOD=GET
+QUERY_STRING=p=.git;a=snapshot;h=master;sf=tar;f=third
+PATH_INFO=
+export GATEWAY_INTERFACE HTTP_ACCEPT REQUEST_METHOD QUERY_STRING PATH_INFO
+pwd
+GITWEB_CONFIG=/home/travis/build/git/git/t/trash directory.t9502-gitweb-standalone-parse-output/gitweb_config.perl
+export GITWEB_CONFIG
+rm -f gitweb.log
+perl -- /home/travis/build/git/git/t/../gitweb/gitweb.perl
+command /usr/bin/perl -- /home/travis/build/git/git/t/../gitweb/gitweb.perl
+perl -w -e 
		open O, ">gitweb.headers";
	while (<>) {
			print O;
			last if (/^\r$/ || /^$/);
		}
		open O, ">gitweb.body";
		while (<>) {
			print O;
		}
		close O;
	 gitweb.output
+command /usr/bin/perl -w -e 
		open O, ">gitweb.headers";
		while (<>) {
			print O;
			last if (/^\r$/ || /^$/);
		}
		open O, ">gitweb.body";
		while (<>) {
			print O;
		}
		close O;
	 gitweb.output
+grep ^[[] gitweb.log
+true
+egrep ^Status: 404 
+head -1 gitweb.headers
+[ ! -s gitweb.body -o -n  ]
error: last command exited with $?=1
not ok 13 - snapshot certain objects: have expected content in master branch - subdir name is required in requested nested path (bad path - empty output and/or HTTP-404)
#	
#		rm -f gitweb.body file_list &&
#		BRANCH=master &&
#		gitweb_run "p=.git;a=snapshot;h=$BRANCH;sf=tar;f=third" &&
#		[ ! -s gitweb.body -o -n "`head -1 gitweb.headers | egrep "^Status: 404 "`" ]
#	
```

As indicated by the end of the log:

```
------------------------------------------------------------------------
Trash directories embedded in this log can be extracted by running:

  curl https://api.travis-ci.org/v3/job/565980864/log.txt |./ci/util/extract-trash-dirs.sh
```

If you do that, you will find that there _is_ a `gitweb.headers` file, and its contents read like this:

```
Status: 500 Internal Server Error
Content-Type: text/html; charset=utf-8
```

That's no good. We clearly expect a 404 ("Not Found"), and instead get the totally unhelpful 500 ("Internal Server Error").

So what to do about this?

In the least, a bit more investigation is in order. For example, `gitweb.body` clearly contains this indicative line:

```html
500 - Failed to transmit output from git-archive&lt;br/&gt;&lt;pre&gt;fatal: # GETTEXT POISON #<span class="cntrl">\n</span>&lt;/pre&gt;&lt;br/&gt;
```

While it is not enough to say what is going wrong, it points a clear way forward how to debug this: put in debug print statements (`print STDERR` should print into that `gitweb.log` file, if the existing line is any indication), and force-push to trigger the CI build. You should probably delete `azure-pipelines.yml` and delete the parts in `.travis.yml` that do not concern themselves with macOS nor `GETTEXT_POISON`, and you probably want to restrict testing to the respective script by replacing the line `T = $(sort $(wildcard t[0-9][0-9][0-9][0-9]-*.sh))` in `t/Makefile` by `T = t9502-gitweb-standalone-parse-output.sh`, so that you have a quicker turnaround between editing the file locally and seeing the result of the Travis build.

I could imagine, for example, that `$readSize` is 0 in https://github.com/git/git/pull/206/files#diff-00703a794a540acf45e225abd6aeda3bR7509, and it could also be that `$cmd` is strange or something. In any case, it would be interesting to see `$retError` verbatim (which the `die()` call probably feeds through the gettext machinery, so you cannot see it in the log).

And yes, I do think that this needs to be fixed. We cannot simply show a 500 when a 404 is the right status to report.

jimklimov:
The answer I gave was to "unsure why it passed here and not there" part ;)

Thanks for the analysis, it seems to match my findings in many rounds of
`make check` and direct calls to test scripts locally.

Essentially the problematic case is that we get a zero-sized `stdout` (so
no info to glean from that) and an error message from the `git` program
with details.

With normal localization, the message we handle specially says that there
were no pathnames in the repo to git-archive matching the request, so it is
an HTTP-404 quite naturally. Other messages cause an unclassified internal
server error (500).

As far as I can see, `git` client exit codes do not point to specific error
conditions so are also rather useless here (at least was so with the
version I looked at; now I'm not with a computer to re-verify quickly). I
guess this too might be addressed directly, but could cause havoc for
decades of scripts relying on something else?.. Though *if* the choice is
"zero vs. nonzero" there is not so much to break. Having a clear exit code
used only for file-not-found cases would solve this PR's CI issue
regardless of locales.

With poisoned gettext we get the same blurb for anything normally reported
on `stderr`, so any error is a HTTP-500 by looking at face value. This is
the bit I wondered about cheating for CI script alone (if poisoned and got
500 then...).

With other locales than `C` or `en*` we probably do too get something not
expected, and I don't think I pushed a solution to that. I *think* now it
would be safe to just force the `C` LANG, LOCALE, LC_ALL etc. for gitweb.pl
calling the `git archive` and maybe restore the `ENV` afterwards (not sure
if this process ends just afterwards).

Jim

jimklimov:
I wonder also if we could try `git ls <pattern from request>` to check if the answer is empty but result is(? Hopefully) a success by exit code.

jimklimov:
And I see I haven't pushed up my latest iterations made before vacation, it seems ;)

dscho:
> One azure test failed fast, but I'm fairly sure I am not behind this:

Indeed. This should be an intermittent failure only, probably a network glitch. Happens. The job is re-running.

dscho:
This looks pretty good now.

While I am not sure that the core developers on the Git mailing list will like the bundled `IPC::Run` (and will likely suggest to load it dynamically, disabling the feature when that package cannot be found), it is now time to submit this to the Git mailing list for review.

Please make sure that your commit messages focuses on answering the question "why?" more than on "how?", and that it wraps at <= 76 columns per line. You might want to look at the existing commit history for inspiration.

Finally send the patch series to the mailing list for review. You can use [GitGitGadget](https://gitgitgadget.github.io), [submitGit](https://submitgit.herokuapp.com) or [send it manually](https://github.com/git-for-windows/git/blob/master/CONTRIBUTING.md#submit-your-patch).

jimklimov:
Oh boy, here we go again :)

Thanks for greening the build, and I'll see what I can do about commit texts, but probably not today.

As for IPC::Run... Originally I thought it would be a "hard" requirement, but then realized that perl allows for optional `use` of libs and via exceptions, so I technically could flag that it is or is not loaded. However the fallback would effectively require to restore the original solution, which was to run git-archive once in case it was all ok on stdout, or re-run it to capture stderr instead and look if it was a fault with finding the file matches. Maybe now though it could fall back right to `git ls*`... Anyway, the downside is that we are making an extra fork (and processing load) to just analyze the error, and potentially running something in different conditions the second time around so the second-run error has a non-100% chance of describing the original failure.

I think on consistency and performance side, running once trumps the not adding dependencies (or bundling, so there are no new build/package required dependencies - for which there is precedent in codebase). But indeed, let's see what reviewers would say. Maybe I should copy the thoughts from above into that cover mail though :)

dscho:
> Maybe I should copy the thoughts from above into that cover mail though :)

Might make a ton of sense :-)

dscho:
> Originally I thought it would be a "hard" requirement, but then realized that perl allows for optional `use` of libs via exceptions, so I technically could flag that it is or is not loaded.

That is a really good idea.

> However the fallback would effectively require to restore the original solution, which was to run git-archive once in case it was all ok on stdout, or re-run it to capture stderr instead and look if it was a fault with finding the file matches.

That should be a better alternative to copying `IPC::Run` into Git's source code for the only reason to support gitweb's archive function.

After all, the exception would not be triggered if the `IPC::Run` is installed, which anybody installing gitweb should most likely do.

> > Maybe I should copy the thoughts from above into that cover mail though :)
> 
> Might make a ton of sense :-)

This still applies 100%.

dscho:
@jimklimov gentle ping?

jimklimov:
Gentle ENOTIME :( Had to dedicate mine to dayjob and some other projects that moved more dynamically. Maybe will get back to this (especially the part about optional IPC::Run dependencies) around x-mas breaks, not sure if before...

dscho:
> the part about optional IPC::Run dependencies

That would be good, as I suspect that it will make the patch _so_ much easier to get into git.git.

dscho:
@jimklimov gentle happy new year ;-)

AzrulHaruna230696:
Duplicate of #

jimklimov:
Hectic start of year, still keep bumping into this tab in my browser, but could not complete the dual-implem (ugly workarounds AND prefer IPC::Run where available) :(

Dedicating a bit of worktime to make a feature we need is relatively easy, dedicating some more to make tests for it to be known-reliable is reasonable, but spending the other 80% of time to conform to this or that upstream's rituals (every project has some) is not easy to justify for those who pay the bills... nor for otherwise suppportive family... and gets backlogged in spare time enthusiasm... :\

dscho:
Why duplicate implementation? If I were you, I'd just fall back to the current code if IPC::Run is not available.

jimklimov:
I did this about 4 years ago so gotta re-read to remember for sure, but
IIRC the original gitweb code had no selective snapshots at all.

So my first shot was with original-like code that ran a command, and either
returned its good result or ran it again to grab the stderr in case of
errors, to discern the cause of error (so 404 or 500) - which is
inefficient and potentially prone to race-faults (two cmds looking at
different HEAD states for example).

Then the second shot was using any (alas all non-standard) perl runner to
grab *both* stdout, stderr and exit code.

And there's also a problem of localized (poisoned gettext) error messages...

All in all, a can of worms that needs a few hours in a row to wrap a head
around :) and then figure out a cover letter for committee of wizened
elders :)

jimklimov:
Rebasing over current master (and rephrasing commit messages) to modernize the PR. Hope to get back to figuring out how to avoid adding dependency on IPC::Run Perl module if that's such a show-stopping issue...

jimklimov:
Rebased the code proposed by PR over today's state of `upstream/master` branch and won the fight with GitGitGadget :)

jimklimov:
I've learned from similar hassles in a few other side projects, and seems that the implementation sufficient for the case in gitweb after all is not as complex as the larger solutions in IPC::Run and others, just a couple of pretty routines.

Waiting for tests to pass in https://github.com/jimklimov/git/tree/gitweb-snapshot-selective-no-IPC-Run and then will carry the change over here.

gitweb/gitweb.perl
```diff
@@ -7790,7 +7790,7 @@ sub git_snapshot {
 		my $error = $@ || 'Unknown failure';
 		print $cgi->header(-status => '500 Execute git-archive failed');
 		if ($DEBUG) {
-			printf STDERR "Execute git-archive failed with exit-code $gitcode (" + $error + ")";
+			printf STDERR "Execute git-archive failed with exit-code $gitcode ($error)";
```
dscho, line 7784: You will want to squash the commit "gitweb.perl : typo fix in debug printout" into the commit that introduced the problem.

dscho (commented):
Please make sure to drop the commits that are reverted in the last part of the patch series. The reviewers on the Git mailing list much prefer seeing patch series that do not introduce bugs that are fixed within the same patch series.

jimklimov:
Yes, I started on it, but then stumbled on having the newer commits replacing the code originally developed with IPC::Run instead of the "upstream" code that had neither way of processing the stdout and stderr consciously. With these commits I can probably easily replace the iterations, swapping one call pattern for another, but it would be much more effort to make IPC::Run "disappear" from commit history completely.

With these two reverted commits it would not be a big hassle to drop either, beside "introducing" a code history that could not pass tests in some of its intermediate commits (e.g. using IPC::Run but not having it anymore, if someone checks out into that point). No idea if it is a problem for this codebase and its traditions.

dscho:
If you don't make `IPC::Run` disappear from the commit history completely, you will run into the very problem that reviewers will have to review your moot attempt to use it. If I were you, I'd try to avoid that: in my experience, reviewers don't like to see their time spent that way.

jimklimov:
I am tempted to squash much of the change set into a handful of commits that lose the iterations and hassle invested, but indeed may be more subjected for review. Sort of like the original 4 IIRC commits were. Half a decade ago... :D

dscho:
> I am tempted to squash much of the change set into a handful of commits that lose the iterations and hassle invested, but indeed may be more subjected for review. Sort of like the original 4 IIRC commits were. Half a decade ago... :D

@jimklimov I know it has been a while, is this PR still on your radar?

jimklimov:
Oh, fell through the cracks I guess, across several workplace and FOSS project switches. And years.

I guess the feature still makes sense (if nobody merged an equivalent yet), so I hope to take another look to tie up the loose ends.

Did any git project side requirements change significantly during this time?

IIRC one big issue was that this feature needed to differentiate causes of `git` program failure, but with exit codes equal and `stdout`/`stderr` forbidden for parsing (CI checked with spewer of bogus strings as logs), there was basically no good way to tackle that reliably. If these constraints are still the case, I'd welcome portable suggestions (e.g. can new exit-codes be defined?) :)

dscho:
TBH I haven't paid attention to the gitweb side of things...

PR #251, opened 2016-06-07
Simplify STARTTLS and SSL setup

pprindeville:
Net::SMTP has supported SSL (without needing Net::SMTP::SSL) since 2.33, and STARTTLS as well.

Setting up either mechanism is a tricky or a bit of a kludge, and moving to a newer version of the module removes this ugliness.

We pick 2.34 because it had some useful fixes to the original STARTTLS functionality.

### DIFF
--- a/Documentation/RelNotes/2.10.4.txt
+++ b/Documentation/RelNotes/2.10.4.txt
@@ -1,0 +1,4 @@
+Git v2.10.4 Release Notes
+=========================
+
+This release forward-ports the fix for "ssh://..." URL from Git v2.7.6
--- a/Documentation/RelNotes/2.11.3.txt
+++ b/Documentation/RelNotes/2.11.3.txt
@@ -1,0 +1,4 @@
+Git v2.11.3 Release Notes
+=========================
+
+This release forward-ports the fix for "ssh://..." URL from Git v2.7.6
--- a/Documentation/RelNotes/2.12.4.txt
+++ b/Documentation/RelNotes/2.12.4.txt
@@ -1,0 +1,4 @@
+Git v2.12.4 Release Notes
+=========================
+
+This release forward-ports the fix for "ssh://..." URL from Git v2.7.6
--- a/Documentation/RelNotes/2.13.5.txt
+++ b/Documentation/RelNotes/2.13.5.txt
@@ -1,0 +1,4 @@
+Git v2.13.5 Release Notes
+=========================
+
+This release forward-ports the fix for "ssh://..." URL from Git v2.7.6
--- a/Documentation/RelNotes/2.14.1.txt
+++ b/Documentation/RelNotes/2.14.1.txt
@@ -1,0 +1,4 @@
+Git v2.14.1 Release Notes
+=========================
+
+This release forward-ports the fix for "ssh://..." URL from Git v2.7.6
--- a/Documentation/RelNotes/2.7.6.txt
+++ b/Documentation/RelNotes/2.7.6.txt
@@ -1,0 +1,25 @@
+Git v2.7.6 Release Notes
+========================
+
+Fixes since v2.7.5
+------------------
+
+ * A "ssh://..." URL can result in a "ssh" command line with a
+   hostname that begins with a dash "-", which would cause the "ssh"
+   command to instead (mis)treat it as an option.  This is now
+   prevented by forbidding such a hostname (which will not be
+   necessary in the real world).
+
+ * Similarly, when GIT_PROXY_COMMAND is configured, the command is
+   run with host and port that are parsed out from "ssh://..." URL;
+   a poorly written GIT_PROXY_COMMAND could be tricked into treating
+   a string that begins with a dash "-".  This is now prevented by
+   forbidding such a hostname and port number (again, which will not
+   be necessary in the real world).
+
+ * In the same spirit, a repository name that begins with a dash "-"
+   is also forbidden now.
+
+Credits go to Brian Neel at GitLab, Joern Schneeweisz of Recurity
+Labs and Jeff King at GitHub.
+
--- a/Documentation/RelNotes/2.8.6.txt
+++ b/Documentation/RelNotes/2.8.6.txt
@@ -1,0 +1,4 @@
+Git v2.8.6 Release Notes
+========================
+
+This release forward-ports the fix for "ssh://..." URL from Git v2.7.6
--- a/Documentation/RelNotes/2.9.5.txt
+++ b/Documentation/RelNotes/2.9.5.txt
@@ -1,0 +1,4 @@
+Git v2.9.5 Release Notes
+========================
+
+This release forward-ports the fix for "ssh://..." URL from Git v2.7.6
--- a/RelNotes
+++ b/RelNotes
@@ -1,1 +1,1 @@
-Documentation/RelNotes/2.14.0.txt
+Documentation/RelNotes/2.14.1.txt
--- a/cache.h
+++ b/cache.h
@@ -1146,6 +1146,14 @@ char *strip_path_suffix(const char *path, const char *suffix);
 int daemon_avoid_alias(const char *path);
 extern int is_ntfs_dotgit(const char *name);
 
+/*
+ * Returns true iff "str" could be confused as a command-line option when
+ * passed to a sub-program like "ssh". Note that this has nothing to do with
+ * shell-quoting, which should be handled separately; we're assuming here that
+ * the string makes it verbatim to the sub-program.
+ */
+int looks_like_command_line_option(const char *str);
+
 /**
  * Return a newly allocated string with the evaluation of
  * "$XDG_CONFIG_HOME/git/$filename" if $XDG_CONFIG_HOME is non-empty, otherwise
--- a/connect.c
+++ b/connect.c
@@ -578,6 +578,11 @@ static struct child_process *git_proxy_connect(int fd[2], char *host)
 
 	get_host_and_port(&host, &port);
 
+	if (looks_like_command_line_option(host))
+		die("strange hostname '%s' blocked", host);
+	if (looks_like_command_line_option(port))
+		die("strange port '%s' blocked", port);
+
 	proxy = xmalloc(sizeof(*proxy));
 	child_process_init(proxy);
 	argv_array_push(&proxy->args, git_proxy_command);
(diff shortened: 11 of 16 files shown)


rsy140588:
sir i dont need to need git please dont any mail to me.....i m just trying
what is git.....

On Wed, Jun 8, 2016 at 3:26 AM, Philip Prindeville <notifications@github.com

pprindeville:
Hmm...  Travis is trying to build against Net::SNMP 2.31.  Where is it getting that version from?  And how do we express the packaging requirement that 2.34 or later is needed?

git-send-email.perl
```diff
@@ -299,7 +304,7 @@ sub signal_handler {
                     "dump-aliases" => \$dump_aliases);
 usage() unless $rc;
 die "--dump-aliases incompatible with other options\n"
-    if !$help and $dump_aliases and @ARGV;
+    if (!$help) and $dump_aliases and @ARGV;
```
dscho: Shouldn't the closing parenthesis be at the *end*, i.e. *after* the `@ARGV`?

git-send-email.perl
```diff
@@ -200,13 +204,14 @@ sub do_edit {
...
-		} @_;
+		} @args;
 	} else {
-		system('sh', '-c', $editor.' "$@"', $editor, @_);
+		system('sh', '-c', $editor.' "$@"', $editor, @args);
 		if (($? & 127) || ($? >> 8)) {
 			die("the editor exited uncleanly, aborting everything");
 		}
 	}
+	return;
```
dscho: This `return` is bogus, correct?

git-send-email.perl
```diff
@@ -411,6 +416,7 @@ sub read_config {
 			$smtp_encryption = 'ssl';
 		}
 	}
+	return;
```
dscho: This `return` is also unnecessary.

git-send-email.perl
```diff
@@ -601,14 +615,14 @@ sub is_format_patch_arg {
 	if ($f eq "--") {
 		push @rev_list_opts, "--", @ARGV;
 		@ARGV = ();
-	} elsif (-d $f and !is_format_patch_arg($f)) {
+	} elsif (-d $f and (!is_format_patch_arg($f))) {
```
dscho: It is well-established in Git's source code that `!` binds stronger than `and`. I'd vote to leave this line as-is.

git-send-email.perl
```diff
@@ -920,7 +935,7 @@ sub validate_address {
 			valid_re => qr/^(?:quit|q|drop|d|edit|e)/i,
 			default => 'q');
 		if (/^d/i) {
-			return undef;
+			return;
```
dscho: This is incorrect. The `validate_address` function clearly is expected to return an address.

git-send-email.perl
sallenwilliams: Face F

On Friday, June 10, 2016, Johannes Schindelin <notifications@github.com>
wrote:

> In git-send-email.perl
> <https://github.com/git/git/pull/251#discussion_r66608127>:
>
> > @@ -411,6 +416,7 @@ sub read_config {
> >  			$smtp_encryption = 'ssl';
> >  		}
> >  	}
> > +	return;
>
> This return is also unnecessary.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/git/git/pull/251/files/c6659b11db7db4e087a12a20954f4afe9ac13c6f#r66608127>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AADGmfXf632Db5_7TBnRwk7psGGhccHaks5qKV8IgaJpZM4IwZoQ>
> .
>


-- 
Plz excuse typos, sending from my handheld.

dscho:
> Travis is trying to build against Net::SNMP 2.31

Did you mean Net::SMTP?

According to https://docs.travis-ci.com/user/ci-environment/ the container used to build uses Ubuntu Precise. Which has Net::SNMP 6.0.1: https://launchpad.net/ubuntu/precise/+source/libnet-snmp-perl. The Net::SMTP package is indeed at 2.31: https://launchpad.net/ubuntu/precise/+source/perl (look in `/perl-5.14.2/cpan/libnet/Net`). I had a quick look and it seems to be at version 2.31 even in Trusty, so I guess that most users will be safely stuck with that version and therefore Git should not require a newer version, either.

git-send-email.perl
pprindeville: It's not bogus so much as it's superfluous.  The function not ending with an explicit 'return;' is the same as ending with a non-value statement or a 'return undef'.  PBP prefers the former.

git-send-email.perl
pprindeville: The parens around the entire expression are optional.  PBP likes explicit parens because some people can't remember their operator precedence.   I don't agree with all PBP rules... Most Perl operator precedence is fairly obvious (except perhaps for 'next', 'last', etc. being considered operators).

git-send-email.perl
pprindeville: It's there for clarity, according to PBP.

git-send-email.perl
pprindeville: Well, if you've ever used -pedantic in gcc, you know that it encourages parens even when the grouping is seemingly obvious.  Perlcritic is the same way... It does occasionally generate some 'noise'... If the tool is to be useful, you find a way to get rid of the noise so that all that is left is meaningful messages.  That might mean throwing in extraneous parens to shut it up.

git-send-email.perl
pprindeville: Unless you're dropping the address, in which case there wouldn't be one.

pprindeville:
Yes, sorry.  Typo.  Net::SMTP.  Debian 8 ships with 2.33 (which would work at a minimum but is lacking some useful fixes).  Fedora 23 is at 3.08.

git-send-email.perl
dscho: On Fri, 10 Jun 2016, Philip Prindeville wrote:

> > @@ -299,7 +304,7 @@ sub signal_handler {
> >                      "dump-aliases" => \$dump_aliases);
> >  usage() unless $rc;
> >  die "--dump-aliases incompatible with other options\n"
> > -    if !$help and $dump_aliases and @ARGV;
> > +    if (!$help) and $dump_aliases and @ARGV;
> 
> The parens around the entire expression are optional.  PBP likes explicit parens because some people can't remember their operator precedence.   I don't agree with all PBP rules... Most Perl operator precedence is fairly obvious (except perhaps for 'next', 'last', etc. being considered operators).

Sure. PBP might recommend this.

All I was saying was that the coding style you introduce is inconsistent
with the existing one. Should you ever choose to submit these patches
upstream, you will be asked to change that.

Don't shoot the messenger.

git-send-email.perl
dscho: On Fri, 10 Jun 2016, Philip Prindeville wrote:

> > @@ -920,7 +935,7 @@ sub validate_address {
> >  			valid_re => qr/^(?:quit|q|drop|d|edit|e)/i,
> >  			default => 'q');
> >  		if (/^d/i) {
> > -			return undef;
> > +			return;
> 
> Unless you're dropping the address, in which case there wouldn't be one.

It is funny. Elsewhere this PR introduce parentheses and return statements
where they could be considered unnecessary and distracting. And here, an
explicit return value (the `validate_address` function *is* expected to
return a value) is removed, puzzling every reader.

If I was asked to accept this patch, I would not.

pprindeville: They're 'necessary' to the extent that compliance with Perl Best Practices as enforced by perlcritic is necessary.  You either subscribe to that style of coding or you don't.

A lot of people have gripes with perlcritic: I have some minor issues of my own with it.

So what are we really disagreeing about?  That PBP is a waste of time?  Then let's talk about that.

How are 'return undef' and 'return' different, in your opinion?

As best as I can tell, they are exactly the same semantically, only one conforms to PBP and the other doesn't.

Would parens around:

$x = $y == 10;

be necessary or not?  The answer is, "it depends on what you're trying to do."

If the above expression is inside the condition of an "if", and you're trying to set $x to whether $y is equal to 10 or not, then they aren't necessary.  If you're trying to assign $x the value of $y and then test if that's equal to 10, then they are necessary.  And if you're trying to make the grouping completely unambiguous (as PBP does), then they are necessary (regardless of where they go).

So it's a question of (1) what's the goal and (2) is that a worthwhile goal?

git-send-email.perl
pprindeville: If you're saying that conforming to PBP isn't worthwhile, and we should drop the patch, then that's fine.  PBP isn't everyone's cup of tea.

git-send-email.perl
gitster: IIRC, "return;" and "return undef;" are not semantically equivalent depending on the calling context (i.e. is the function called in scalar context or list context?), and PBP suggests that the former is less error prone most of the time.

pprindeville:
Junio: these patches require Net::SMTP 2.34 but Ubuntu Precise is at 2.31 (and therefore Travis is too), so the test automation fails.

How to handle this?

gitster:
> these patches require Net::SMTP 2.34 but Ubuntu Precise is at 2.31 (and therefore Travis is too), so the test automation fails.
> How to handle this?

Sorry, I do not quite understand the question. If you _know_ that many in the current user base use older version and your patch fails with that version, the only way to "handle" it would be to make it not to fail with the older version, while allowing the resulting version with the patch work better with the newer version, no?

pprindeville:
> If you know that many in the current user base use older version and your patch fails with that version, the only way to "handle" it would be to make it not to fail with the older version, while allowing the resulting version with the patch work better with the newer version, no?

Or alternatively hold back on merging this PR until Net::SMTP 2.34 or later becomes ubiquitous.

pprindeville:
@ERICLAMARCUS1971: Sorry, "split" meaning what?

Also, I'm confused why there are failing checks... These seem to be a Travis configuration issue.

mmwradcliffe:
What. What does any of this mean. I'm not even sure how I'm a developer.
Help me understand this please someone.

On Nov 10, 2016 11:17 AM, "ERICLAMARCUS1971" <notifications@github.com>
wrote:

*@ERICLAMARCUS1971* approved this pull request.

split

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/git/git/pull/251#pullrequestreview-8078311>, or mute
the thread

pprindeville:
> If you know that many in the current user base use older version and your patch fails with that version, the only way to "handle" it would be to make it not to fail with the older version, while allowing the resulting version with the patch work better with the newer version, no?

Sounds like the test should be skipped if Net::SMTP < 2.34...

dscho:
> Sounds like the test should be skipped if Net::SMTP < 2.34...

No. Because then we would ignore that this test is broken with `NET::SMTP`.

> What's the probability of someone running a current git against an ancient version of the Net::Cmd suite, anyway?

If Travis uses it, you can bet that there are still users stuck with that version.

I guess the safest bet is to wait with this patch until *at least* Travis has the new version.

PR #255, opened 2016-06-17
add dependent libs info in `INSTALL`

wjm3219:


dscho:
Git can be compiled on a plethora of setups, including a couple of different Linux distributions (and your patch targets only a fraction of those, not every Linux distribution has the `*-devel` naming scheme).

If you think that you can make this a bit more general, please do so and then send the patch to the mailing list for review. You can use [GitGitGadget](https://gitgitgadget.github.io), [submitGit](https://submitgit.herokuapp.com) or [send it manually](https://github.com/git-for-windows/git/blob/master/CONTRIBUTING.md#submit-your-patch).

PR #273, opened 2016-07-19
Check and set terms

pranitbauva1997:
cc: @chriscool @larsxschneider 

I can't use `git-format-patch` and `git-send-email` now so I am using submitgit now.

### DIFF
--- a/builtin/bisect--helper.c
+++ b/builtin/bisect--helper.c
@@ -10,11 +10,11 @@ static const char * const git_bisect_helper_usage[] = {
 
 int cmd_bisect__helper(int argc, const char **argv, const char *prefix)
 {
-	int next_all = 0;
+	enum { NEXT_ALL = 1 } cmdmode = 0;
 	int no_checkout = 0;
 	struct option options[] = {
-		OPT_BOOL(0, "next-all", &next_all,
-			 N_("perform 'git bisect next'")),
+		OPT_CMDMODE(0, "next-all", &cmdmode,
+			 N_("perform 'git bisect next'"), NEXT_ALL),
 		OPT_BOOL(0, "no-checkout", &no_checkout,
 			 N_("update BISECT_HEAD instead of checking out the current commit")),
 		OPT_END()
@@ -23,9 +23,14 @@ int cmd_bisect__helper(int argc, const char **argv, const char *prefix)
 	argc = parse_options(argc, argv, prefix, options,
 			     git_bisect_helper_usage, 0);
 
-	if (!next_all)
+	if (!cmdmode)
 		usage_with_options(git_bisect_helper_usage, options);
 
-	/* next-all */
-	return bisect_next_all(prefix, no_checkout);
+	switch (cmdmode) {
+	case NEXT_ALL:
+		return bisect_next_all(prefix, no_checkout);
+	default:
+		die("BUG: unknown subcommand '%d'", cmdmode);
+	}
+	return 0;
 }
--- a/builtin/bisect--helper.c
+++ b/builtin/bisect--helper.c
@@ -2,19 +2,73 @@
 #include "cache.h"
 #include "parse-options.h"
 #include "bisect.h"
+#include "refs.h"
 
 static const char * const git_bisect_helper_usage[] = {
 	N_("git bisect--helper --next-all [--no-checkout]"),
+	N_("git bisect--helper --check-term-format <term> <orig_term>"),
 	NULL
 };
 
+/*
+ * Check whether the string `term` belongs to the set of strings
+ * included in the variable arguments.
+ */
+LAST_ARG_MUST_BE_NULL
+static int one_of(const char *term, ...)
+{
+	int res = 0;
+	va_list matches;
+	const char *match;
+
+	va_start(matches, term);
+	while (!res && (match = va_arg(matches, const char *)))
+		res = !strcmp(term, match);
+	va_end(matches);
+
+	return res;
+}
+
+static int check_term_format(const char *term, const char *orig_term)
+{
+	int res;
+	char *new_term = xstrfmt("refs/bisect/%s", term);
+
+	res = check_refname_format(new_term, 0);
+	free(new_term);
+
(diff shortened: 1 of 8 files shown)


builtin/bisect--helper.c
```diff
@@ -2,30 +2,525 @@
...
+static int check_term_format(const char *term, const char *orig_term)
+{
+	struct strbuf new_term = STRBUF_INIT;
+	strbuf_addf(&new_term, "refs/bisect/%s", term);
+
+	if (check_refname_format(new_term.buf, 0)) {
+		strbuf_release(&new_term);
+		return error(_("'%s' is not a valid term"), term);
+	}
+	strbuf_release(&new_term);
```
larsxschneider: you could store the output of `check_refname_format` in a variable, release `new_term` and then check the variable. I think balancing the "init" and "release" calls like that helps to avoid memory errors. However, I am not sure what the "git" projects preferred way here is. @chriscool what do you think?

builtin/bisect--helper.c
```diff
@@ -2,30 +2,525 @@
...
+	va_end(matches);
+
+	return res;
+}
+
+static int check_term_format(const char *term, const char *orig_term)
+{
+	struct strbuf new_term = STRBUF_INIT;
+	strbuf_addf(&new_term, "refs/bisect/%s", term);
+
```
larsxschneider: super minor: I wonder why you created a newline here. The two lines above and the 5 lines below are a logical "block", no?

builtin/bisect--helper.c
```diff
@@ -2,30 +2,525 @@
...
+	if (check_term_format(bad, "bad") || check_term_format(good, "good"))
+		return -1;
+
+	fp = fopen(git_path_bisect_terms(), "w");
+	if (!fp)
+		return error_errno(_("could not open the file BISECT_TERMS"));
+
+	res = fprintf(fp, "%s\n%s\n", bad, good);
+	fclose(fp);
+	return (res < 0) ? -1 : 0;
```
larsxschneider: AFAIK even if fprintf succeeds, that doesn't mean the write was successful because the strings stays in a buffer. fclose internally calls flush to write the buffer to the file. this operation can fail and therefore fclose can fail. bottom line: I would check for the return code of fclose, too.

builtin/bisect--helper.c
```diff
@@ -2,30 +2,525 @@
...
+}
+
+static int bisect_reset(const char *commit)
+{
+	struct strbuf branch = STRBUF_INIT;
+
+	if (!commit) {
+		if (strbuf_read_file(&branch, git_path_bisect_start(), 0) < 1) {
+			printf("We are not bisecting.\n");
+			return 0;
```
larsxschneider: I think you are not releasing `branch` here

builtin/bisect--helper.c
```diff
@@ -2,30 +2,525 @@
...
+	struct strbuf tag = STRBUF_INIT;
+	struct strbuf commit_name = STRBUF_INIT;
+	struct object_id oid;
+	struct commit *commit;
+	struct pretty_print_context pp = {0};
+	FILE *fp;
+
+	if (!strcmp(state, terms->term_bad.buf))
+		strbuf_addf(&tag, "refs/bisect/%s", state);
+	else if(one_of(state, terms->term_good.buf, "skip", NULL))
```
larsxschneider: missing whitespace between if and (

builtin/bisect--helper.c
```diff
@@ -2,30 +2,525 @@
...
+		strbuf_addf(&tag, "refs/bisect/%s-%s", state, rev);
+	else
+		return error(_("Bad bisect_write argument: %s"), state);
```
larsxschneider: shouldn't you release tag and comit_name here?

builtin/bisect--helper.c
chriscool: @larsxschneider Yeah I think it is better to do what you suggest here. I don't think there is a prefered way for this in the project.

What I would suggest would be to use xstrfmt() instead of strbuf to save one line.

builtin/bisect--helper.c
chriscool: It is needed to release `branch` only if some memory has been allocated for it. I am not sure it is the case here. It depends on what `strbuf_read_file()` does before failing.

builtin/bisect--helper.c
chriscool: I don't think it is needed to release them as assigning STRBUF_INIT is not enough to allocate memory for them.

larsxschneider: thanks for the explanation!

builtin/bisect--helper.c
pranitbauva1997: It wasn't looking quite nice to me.

builtin/bisect--helper.c
pranitbauva1997: Thanks for the suggestion. Will use it!

builtin/bisect--helper.c
pranitbauva1997: I will take a `logical OR` with fclose() then. Thanks, I didn't know about the flush part.

submitgit:
@pranitbauva1997 sent these 12 commits (https://github.com/git/git/compare/08bb3500a2a718c3c78b0547c68601cafa7a8fd9...e4ebeeadcad18730f406ec62a46b25031a52cd74) as a set of patches to the mailing list with [_submitGit_](https://github.com/rtyley/submitgit) - here on [Gmane](http://mid.gmane.org/010201560a4781be-e8418d6a-5996-45cd-b11b-ca25894ad7f2-000000@eu-west-1.amazonses.com), [MARC](http://marc.info/?i=010201560a4781be-e8418d6a-5996-45cd-b11b-ca25894ad7f2-000000@eu-west-1.amazonses.com) [](010201560a4781be-e8418d6a-5996-45cd-b11b-ca25894ad7f2-000000@eu-west-1.amazonses.com)

PR #279, opened 2016-07-25
gitweb: escape link body in format_ref_marker

abrauchli:
Fix a case where an html link can be generated from unescaped input
resulting in invalid strict xhtml or potentially injected code.

An overview of a repo with a tag "1.0.0&0.0.1" would previously result
in an unescaped amperstand in the link body.

Signed-off-by: Andreas Brauchli <a.brauchli@elementarea.net>


abrauchli:
Superseeded by #283

PR #284, opened 2016-08-04
remove shallow check on unshallow

gjbianco:
seeing as this is simply an alias to `--depth`, it is unexpected that this would throw a fatal error when `--depth` does not


gjbianco:
Added/removed simple change to re-trigger build (timed out last time for unknown reasons).

builtin/fetch.c
```diff
@@ -1280,8 +1280,6 @@ int cmd_fetch(int argc, const char **argv, const char *prefix)
 	if (unshallow) {
 		if (depth)
 			die(_("--depth and --unshallow cannot be used together"));
```
dscho: The commit message's claim that `--unshallow` and `--depth` are aliases is contradicted by these three lines.

dscho:
@gjbianco gentle ping?

dscho:
@gjbianco gentle ping?

builtin/fetch.c
gjbianco: It seems like the original intent for the `--unshallow` flag was to be an alias for `--depth= 2147483647`. At least according to the original commit: https://github.com/git/git/commit/4dcb167fc3536db0e78c50f239cd3a19afd383fa

It is kinda strange that the original implementation doesn't seem to match that intent. The documentation itself doesn't seem to address it one way or the other (I was definitely surprised when my CI/CD pipeline was failing due to this check).

As is, I don't really see why I would ever use `--unshallow` (at least in a build pipeline) given that it is not idempotent.

dscho: That commit's message makes it quite clear that `--depth=2147483647` was an really unintuitive and unfortunate way to do what `--unshallow` does correctly.

So I really think that it is incorrect to say that `--unshallow` is just an alias to `--depth=2147483647`.

In any case, you _could_ of course take these patches to the Git mailing list for review (I don't expect the reactions there to be much different from mine, though).

If you want to do that (i.e. send these patches to the Git mailing list), first please make sure that your patch series does not contain any merge commits, that your commit messages focus on answering the question "why?" more than on "how?", and that they wrap at <= 76 columns per line. Also do make sure to add your [sign off](https://git-scm.com/docs/SubmittingPatches#sign-off).

Finally actually send the patches to the mailing list for review. You can use [GitGitGadget](https://gitgitgadget.github.io), [submitGit](https://submitgit.herokuapp.com) or [send it manually](https://github.com/git-for-windows/git/blob/master/CONTRIBUTING.md#submit-your-patch).

gjbianco: The original purpose of `--unshallow` seems to be as a direct alias to `--depth=2147483647`. I'm not sure how much more explicitly they could have stated that than how they did:

> Make easy-to-remember --unshallow an alias for --depth=2147483647.

Regardless of intent, both commands do the same thing. The _sole_ difference between the two (at least that I'm seeing) is that `--unshallow` has an extra check to explicitly make it _not_ idempotent. I don't see any value in this check. If not having that check causes problems elsewhere, then shouldn't `--depth` also have that check?

dscho: > Regardless of intent, both commands do the same thing.

That is an implementation detail that could change at any time.

The purpose of `--depth=<n>` _never_ is to unshallow, even if a large enough `<n>` will do just that. And what constitutes "large enough" depends on the shape of the repository. The purpose of `--depth=<n>` is to deepen a branch if it is shallow, and that _might_ result in the branch becoming unshallow. It might also result in the exact opposite: using `--depth=<n>` on a non-shallow branch might very well make it shallow!

The purpose of `--unshallow`, on the other hand, is always to make a shallow branch (or set of branches) non-shallow.

As a consequence, it makes a total lot of sense to error out if the user wanted to unshallow a branch that is already not shallow. The same _cannot_ be said for `--depth=<n>`, no matter how large the `<n>`.

Don't get hung up on the things going on under the hood _for the moment_. That's not what matters. The intention of the user interface is what matters.

dscho:
@gjbianco I am not sure what to do about this patch. On the one hand, in my opinion it does the wrong thing. On the other hand, you might insist that it is correct, in which case you would have to clean up the branch a bit before submitting the patch for review:

- No merge commits. The patch review process on the Git mailing list only handles patches, i.e. diffs of commits with only one parents.
- It should probably be just a single patch.
- Please make sure that your commit message focuses on answering the question "why?" more than on "how?", and that it wraps at <= 76 columns per line.
- Also do make sure to add your [sign off](https://git-scm.com/docs/SubmittingPatches#sign-off).

dscho:
(Just to clarify: you will want to use an interactive rebase for cleaning up your branch.)

dscho:
> I am not sure what to do about this patch. On the one hand, in my opinion it does the wrong thing. On the other hand, you might insist that it is correct, in which case you would have to clean up the branch a bit before submitting the patch for review

@gjbianco what will it be? Send the patch to the Git mailing list, or abandon this PR?

dscho:
@gjbianco gentle ping?

dscho:
> > I am not sure what to do about this patch. On the one hand, in my opinion it does the wrong thing. On the other hand, you might insist that it is correct, in which case you would have to clean up the branch a bit before submitting the patch for review
> 
> @gjbianco what will it be? Send the patch to the Git mailing list, or abandon this PR?

Okay, since I am obviously being forced to [talk to the hand](https://www.youtube.com/watch?v=SXk7TuxrDBI), I feel free to interpret the silence in any which way I want, and I pick the latter option: abandoning this PR.

PR #285, opened 2016-08-11
http.c: instead of hardcoding gzip encoding, ask for curls defaults

dambaev:
try to build git against curl which has been built without zlib support.
Then try git clone https://chromium.googlesource.com/chromium/src, you will get "fatal: https://chromium.googlesource.com/chromium/src/info/refs not valid: is this a git repository?" because curl will return compressed response to remote-curl. This is due to hardcoded "gzip" requirements.

CURL API docs: "If a zero-length string is set like "", then an Accept-Encoding: header containing all built-in supported encodings is sent."

### DIFF
--- a/http.c
+++ b/http.c
@@ -1568,7 +1568,7 @@ static int http_request(const char *url,
 
 	curl_easy_setopt(slot->curl, CURLOPT_URL, url);
 	curl_easy_setopt(slot->curl, CURLOPT_HTTPHEADER, headers);
-	curl_easy_setopt(slot->curl, CURLOPT_ENCODING, "gzip");
+	curl_easy_setopt(slot->curl, CURLOPT_ENCODING, "");
 
 	ret = run_one_slot(slot, &results);
 


dambaev:
travis failed because it expects hardcoded gzip too:
cat >exp <<EOF
> GET /smart/repo.git/info/refs?service=git-upload-pack HTTP/1.1
> Accept: */*
> Accept-Encoding: **gzip**
> Pragma: no-cache
< HTTP/1.1 200 OK
< Pragma: no-cache
< Cache-Control: no-cache, max-age=0, must-revalidate
< Content-Type: application/x-git-upload-pack-advertisement
> POST /smart/repo.git/git-upload-pack HTTP/1.1
> Accept-Encoding: **gzip**
> Content-Type: application/x-git-upload-pack-request
> Accept: application/x-git-upload-pack-result
> Content-Length: xxx
< HTTP/1.1 200 OK
< Pragma: no-cache
< Cache-Control: no-cache, max-age=0, must-revalidate
< Content-Type: application/x-git-upload-pack-result
EOF
but curl may support more/less depends on test environment

rugk:

dscho:
Doesn't that setting basically *ask* for `gzip` encoding *iff* the server supports it? I.e. does it really harm to add it, or put another way: what problem does this PR solve?

(Hint: this points to a serious lack of explanation in the commit message.)

If you still think that this change might be desirable, please improve the commit message, add your [sign off](https://git-scm.com/docs/SubmittingPatches#dco) and then send the patch to the mailing list for review. You can use [GitGitGadget](https://gitgitgadget.github.io), [submitGit](https://submitgit.herokuapp.com) or [send it manually](https://github.com/git-for-windows/git/blob/master/CONTRIBUTING.md#submit-your-patch).

dambaev:
The issue is not a server. The issue is that git tells curl to accept "gzip" encoding without checking if curl is able to decode this encoding. So git thinks, that curl is always built with zlib support. And when curl does not supports gzip encoding, git gets encoded response, which it failes to accept. This patch asks curl to accept encodings, supported by curl (possibly, including gzip).

I will follow your advice, when I will have a time.

dscho:
> The issue is that git tells curl to accept "gzip" encoding without checking if curl is able to decode this encoding. So git thinks, that curl is always built with zlib support. And when curl does not supports gzip encoding, git gets encoded response, which it failes to accept. This patch asks curl to accept encodings, supported by curl (possibly, including gzip).

Ah, see, this makes for an excellent explanation! Please put _that_ into the commit message. It could have prevented my misunderstanding, that's how good of a commit message it will make.

dscho:
@dambaev gentle ping?

dambaev:
sorry for being such an unresponsive person. I had finally had time to follow your suggestions and it's turned out, that the same issue had been already addressed in a commit 1a53e692afd417897d76bff4ce54bc05a3a976b2 of a master branch dated 23 May 2018, so this pull request is not more relevant

PR #286, opened 2016-08-14
make rebase respect core.hooksPath if set

ryenus:
when looking for pre-rebase and post-rewrite hooks, git-rebase
only looks for hooks dir using `git rev-parse --git-path hooks`,
which didn't consider the path overridden by core.hooksPath.


ryenus:
Now it's clear that it's better to fix `git rev-parse --git-path hooks` to return the correct hooks path even if overridden by `core.hooksPath`, as [proposed by Dscho][1].

[1]: https://public-inbox.org/git/xmqq1t1nbdza.fsf@gitster.mtv.corp.google.com/T/#m57092f4d56333db4dac0d633e28e1d89030416ce

PR #291, opened 2016-08-31
Viewing Unpushed Git Commits

gustavoconci:


rugk:

rsy140588:
Dear i do not wanat to get me any kinds of Git...........Understand

anatolyborodin:
@rsy140588 go to https://github.com/watching and unwatch all the repos you don't watch.

Also check the settings here https://github.com/settings/notifications

But there will still be the threads like this one, where you posted a comment. You need to unsubscribe manually there (see the button "Unsubscribe" to the left?) Also check https://github.com/git/git/notifications

gitster:
Retitle to hint that readers can tell that this is about "git-prompt" with only the first line of your commit log message.

Why do you even need "--stat" when running "diff" whose output you do not want to see?

The current branch may not be set to push to origin/$current_branch_name, and there are at least two failure cases in your patch. (1) the current branch may instead configured be pushed to origin/$other_branch. (2) the current branch may not even have origin/ counterpart. What should happen in these (and possibly other) cases need to be thought through and either coded to cope with them, or explained in the log message why the patch "punts" on them.

These proposed changes should be sent to git@vger.kernel.org, not here.  As somebody already pointed out, the pull requests to this GitHub repository will not be merged.

PR #295, opened 2016-09-19
Format tag verify cleandiff

SantiagoTorres:
Here's a cleaner version of the diff. I just realized that moving the code over would not be as trivial as I thought. Here's what I did:

1.- Squashed one of the commits (in which the commit itself was a squash from something else)
2.- Fixed up some commented lines on commit 603a94a 
3.- Re-ordered commits so they make more logical sense (global -> add quiet flag -> expose the interface-> use all of it to --format both verify functions).

I however had some issues when relocating the show_ref_item calls because we need a "name" parameter for the ref and this is not available under gpg_verify_tag.

Maybe we should send this as-is right now?

### DIFF
--- a/builtin/tag.c
+++ b/builtin/tag.c
@@ -30,8 +30,9 @@ static const char * const git_tag_usage[] = {
 
 static unsigned int colopts;
 static int force_sign_annotate;
+static const char *fmt_pretty;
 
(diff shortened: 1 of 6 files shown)


SantiagoTorres:
woops, this was meant for local development 👎

PR #306, opened 2016-11-16
bultin/check-ignore.c: Remove requirement for -z and --stdin

PAntoine:
check-ignore -z works fine without --stdin. The docs say this is valid.
This change makes the code match the docs.

Signed-off-by: Peter Antoine <git@peterantoine.me.uk>


PAntoine:
Need to modify the tests.
Will push at a later time.

PR #324, opened 2017-02-17
git-p4: supports text encoding of history and author for non-unicode p4d

exawon:


AraHaan:
I wonder where git's bugtracker is.

PR #328, opened 2017-02-27
git-p4: Supports text encoding of history and author for non-unicode p4d

exawon:
This will be useful if you don't have any chance to try migration of p4d to unicode mode.

### DIFF
--- a/git-p4.py
+++ b/git-p4.py
@@ -957,6 +957,15 @@ def wildcard_present(path):
     m = re.search("[*#@%]", path)
     return m is not None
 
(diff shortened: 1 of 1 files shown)


dscho:
@exawon I see that you deleted your fork, so you would not be able to update this PR even if you wanted to. Therefore, I will close it.

If you still want to pursue this contribution, please make sure that your commit message focuses on answering the question "why?" more than on "how?", and that it wraps at <= 76 columns per line. Also do make sure to add your [sign off](https://git-scm.com/docs/SubmittingPatches#dco).

PR #329, opened 2017-02-27
git-p4: Preserve commiter

exawon:
This will be useful when your commits are submitted to p4d and sync back to git if you don't have permission to preserve user on perforce.


dscho:
@exawon I see that you deleted your fork, so you would not be able to update this PR even if you wanted to. Therefore, I will close it.

If you still want to pursue this contribution, please make sure that your commit message focuses on answering the question "why?" more than on "how?", and that it wraps at <= 76 columns per line. Also do make sure to add your [sign off](https://git-scm.com/docs/SubmittingPatches#dco).

PR #330, opened 2017-02-27
Changed post-increment to pre and few variable initialization.

ashishpro:
Pre-increment is considered to be more efficient than post increment because it don't involves keeping a copy of the previous value around and adds a little extra code to it. It simply does his job without occupying unwanted space.
Reduced the LOC by long similar initialization.

### DIFF
--- a/apply.c
+++ b/apply.c
@@ -288,7 +288,7 @@ static uint32_t hash_line(const char *cp, size_t len)
 {
 	size_t i;
 	uint32_t h;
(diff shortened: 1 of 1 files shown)


gitster:
> Pre-increment is considered to be more efficient than post increment because it don't involves keeping a copy of the previous value around

Please stop the nonsense.

This project is not in C++, compilers know between i++ and ++i used in void context would not make any difference, and it is an established pattern since K&R to write for loops with post-increment.

  for (i = ...; i < ...; i++)

PR #332, opened 2017-02-28
Add a Travis matrix "coordinate" for 32-bit Linux

dscho:
This Pull Request really is only here to test this with the real Travis CI.


dscho:
@larsxschneider this could end up addressing your https://github.com/larsxschneider/git/issues/20...

PR #340, opened 2017-03-21
Refactor sub-process support out of convert.c into a separate module

benpeart:
We have a couple of patch series we’re working on (ObjectDB/Read-Object, Watchman integration) where we could use the ability to have a background process running that can accept multiple commands thus avoiding the overhead of spawning a new process for every command.  I’ve taken the code from LFS that does this and refactored it into a separate “sub-process” module so that we can centralize and reuse this logic.  I also updated convert.c to use this new module.

I'll be posting this to the mailing list but understood that if I open a PR then a build for various platforms would be kicked off and tests run.  I've verified this builds and passes all tests on Windows but no other platforms.

### DIFF
--- a/Documentation/technical/api-sub-process.txt
+++ b/Documentation/technical/api-sub-process.txt
@@ -1,0 +1,55 @@
+sub-process API
+===============
+
+The sub-process API makes it possible to run background sub-processes
+that should run until the git command exits and communicate with it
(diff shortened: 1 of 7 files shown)


rtyley:
You could also give [submitGit](https://submitgit.herokuapp.com/) a try for submitting to the mailing list, if you haven't done it before: https://submitgit.herokuapp.com/git/git/pull/340

PR #343, opened 2017-04-12
Reread the git-rebase-todo file after every successful "exec" command

shicks:
It's possible that exec'd commands may change the TODO.  Prior to 2.12, this was the behavior, but the changes to sequencer.c made it clobber any changes instead of picking them up.

### DIFF
--- a/sequencer.c
+++ b/sequencer.c
@@ -1200,6 +1200,7 @@ struct todo_list {
 	struct todo_item *items;
 	int nr, alloc, current;
 	int done_nr, total_nr;
+        struct stat_data stat;
 };
 
 #define TODO_LIST_INIT { STRBUF_INIT }
@@ -1330,6 +1331,7 @@ static int count_commands(struct todo_list *todo_list)
 static int read_populate_todo(struct todo_list *todo_list,
 			struct replay_opts *opts)
 {
+        struct stat st;
 	const char *todo_file = get_todo_path(opts);
 	int fd, res;
 
@@ -1342,6 +1344,11 @@ static int read_populate_todo(struct todo_list *todo_list,
 		return error(_("could not read '%s'."), todo_file);
 	}
 	close(fd);
+
+        res = stat(todo_file, &st);
+        if (res)
+                return error(_("could not stat '%s'"), todo_file);
+        fill_stat_data(&todo_list->stat, &st);
 
 	res = parse_insn_buffer(todo_list->buf.buf, todo_list);
 	if (res) {
@@ -2032,6 +2039,20 @@ static int pick_commits(struct todo_list *todo_list, struct replay_opts *opts)
 			*end_of_arg = '\0';
 			res = do_exec(item->arg);
 			*end_of_arg = saved;
+
+                        /* Reread the todo file if it has changed. */
+                        if (res == 0) {
+                                struct stat st;
+                                if ((res = stat(get_todo_path(opts), &st)))
+                                        return res;
+                                if (match_stat_data(&todo_list->stat, &st)) {
+                                        todo_list_release(todo_list);
+                                        if ((res = read_populate_todo(todo_list, opts)))
+                                              return res;
+                                        todo_list->current = 0;
+                                        continue;
+                                }
+                        }
 		} else if (!is_noop(item->command))
 			return error(_("unknown command %d"), item->command);
 
--- a/t/t3428-rebase-edit-todo.sh
+++ b/t/t3428-rebase-edit-todo.sh
@@ -1,0 +1,22 @@
+#!/bin/sh
+
+test_description='rebase should reread the todo file if an exec modifies it'
+
+. ./test-lib.sh
+
+todo=$(git rev-parse --git-dir)/rebase-merge/git-rebase-todo
+
+test_expect_success setup '
+
+	>file &&
+	git add file &&
+	test_tick &&
+	git commit -m "Initial commit"
+'
+
+test_expect_success 'rebase exec modifies rebase-todo' "
+        git rebase HEAD -x 'echo x touch F >>\"$todo\"' &&
+        test -e F
+"
+
+test_done


sequencer.c
```diff
@@ -1290,7 +1290,6 @@ static int parse_insn_buffer(char *buf, struct todo_list *todo_list)
 
 	for (i = 1; *p; i++, p = next_p) {
 		char *eol = strchrnul(p, '\n');
-
```
dscho: Please leave whitespace in unrelated areas alone.

sequencer.c
```diff
@@ -2032,6 +2031,10 @@ static int pick_commits(struct todo_list *todo_list, struct replay_opts *opts)
 			*end_of_arg = '\0';
 			res = do_exec(item->arg);
 			*end_of_arg = saved;
+
+                        /* Reread the TODO in case the command changed it. */
+                        if (res == 0)
+                          return sequencer_continue(opts);
```
dscho: That is quite a bit heavy-handed, and wastes resources...

t/t3405-rebase-malformed.sh
```diff
@@ -69,4 +69,11 @@ test_expect_success 'rebase commit with diff in message' '
 	test_cmp G G0
 '
 
+test_expect_success 'rebase exec modifies rebase-todo' '
+        git rebase HEAD -x "echo x \"echo foo >H\" >> \"$(
```
dscho: Should this not fail, if it modifies the todo file?

shicks: Why should it fail?  With this patch, modifying the todo file is supported.

sequencer.c
shicks: Made it lighter.

sequencer.c
shicks: Oops, forgot to revert from some debugging.

shicks:
I've implemented the size/mtime/inode check as you suggested, and added a failing test case to demonstrate where this approach falls short.

t/t3405-rebase-malformed.sh
dscho: You add this test case to the "malformed" test script... Wrong script?

In any case, it would be easier to read and understand if we used something like `touch telltale` instead of the `echo` with the quoted quotes. Also, for readability, I would strongly suggest setting `todo="$(git rev-parse --git-path rebase-merge/git-rebase-todo)"` and then using that, and instead of appending to the `git-rebase-todo` script, simply replace it with a single command.

dscho: And actually, you can leave the test in here if you replace the `git-rebase-todo` file with a single word: `invalid`. Then the rebase should fail. But of course, that might not catch a problem where it fails for the wrong reasons.

dscho:
It is a bit unelegant to get the stat data each time before exec'ing the command. Why not do that in `read_populate_todo()` and have a new `struct cache_time` member of the `struct todo_list`?

dscho:
Also, please Sign-off on your commit in the commit message (to indicate that you testify that you are legally entitled to contribute this change), so that I can go and apply my suggestions rather than telling you to do so.

t/t3405-rebase-malformed.sh
shicks: That's testing a very different thing, though.  I've moved the test into a new file: t3428-rebase-edit-todo.sh, since it doesn't really fit into the others.

shicks:
I store a struct stat_data, which is a whole lot nicer and (via match_stat_data) automatically checks (nanosecond) mtime, inode, and size all at once.  So that seemed like a win.  Thanks for all the suggestions!

shicks:
Ping?

sequencer.c
```diff
@@ -1200,6 +1200,7 @@ struct todo_list {
 	struct todo_item *items;
 	int nr, alloc, current;
 	int done_nr, total_nr;
+        struct stat_data stat;
```
dscho: I don't mind fixing these things myself, but when you contribute this patch to the mailing list, you may get upset comments about using spaces for indentation. To avoid such comments, I recommend using tabs.

sequencer.c
```diff
@@ -2032,6 +2039,20 @@ static int pick_commits(struct todo_list *todo_list, struct replay_opts *opts)
 			*end_of_arg = '\0';
 			res = do_exec(item->arg);
 			*end_of_arg = saved;
+
+                        /* Reread the todo file if it has changed. */
+                        if (res == 0) {
```
dscho: The mailing list reviews are sometimes, how to put it, uhm, let's say, special in their ways. They would most likely take offense at `res == 0`, claiming that the rest of the code uses `!res` for these kinds of tests. I recommend changing it before the reviewers go wild.

sequencer.c
```diff
@@ -2032,6 +2039,20 @@ static int pick_commits(struct todo_list *todo_list, struct replay_opts *opts)
...
+                                struct stat st;
```
dscho: It is a matter of taste, but I really like to keep indentation small. I would move this `struct stat st` declaration just after `int saved = *end_of_arg;` to save on one indentation level (especially given that we expect to need that variable, because we expect the `exec` command to succeed most of the time). As I said, it is a matter of taste, and your taste may be different.

sequencer.c
```diff
@@ -2032,6 +2039,20 @@ static int pick_commits(struct todo_list *todo_list, struct replay_opts *opts)
...
+                                if ((res = stat(get_todo_path(opts), &st)))
+                                        return res;
```
dscho: This does not report what went wrong, possibly leaving the user puzzled. Maybe

```c
if (stat(get_todo_path(opts), &st))
    return error_errno(_("could not re-read '%s'"), get_todo_path(opts));
```

would make sense instead?

dscho: Oh, I see that there is one more thing the code does afterwards: it increments `todo_list->current`. And only then it checks whether `res` is non-zero. Maybe we should not return here but use

```c
if (stat(get_todo_path(opts), &st))
    res = error_errno(_("could not re-read '%s'"), get_todo_path(opts));
else if (match_stat_data(&todo_list->stat, &st)) {
    todo_list_release(todo_list);
    if (read_populate_todo(todo_list, opts))
        res = -1; /* error already reported */
    todo_list->current = -1; /* will be incremented below */
}
```

t/t3428-rebase-edit-todo.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+
+test_description='rebase should reread the todo file if an exec modifies it'
+
+. ./test-lib.sh
+
+todo=$(git rev-parse --git-dir)/rebase-merge/git-rebase-todo
```
dscho: I wonder why you do not need double quotes here, despite the fact that we are in a directory with spaces in its name. I guess it is because `--git-dir` returns `.git` here.

t/t3428-rebase-edit-todo.sh
```diff
@@ -0,0 +1,22 @@
...
+
+test_expect_success setup '
+
+	>file &&
+	git add file &&
+	test_tick &&
+	git commit -m "Initial commit"
```
dscho: Or much shorter (and less error-prone): `test_commit initial`

t/t3428-rebase-edit-todo.sh
```diff
@@ -0,0 +1,22 @@
...
+'
+
+test_expect_success 'rebase exec modifies rebase-todo' "
+        git rebase HEAD -x 'echo x touch F >>\"$todo\"' &&
```
dscho: It took me a moment to realize that the `x` is not a place holder here. It would most likely make sense to stave off such unnecessary confusion by spelling out `exec` explicitly.

Also: The quotes around `$todo` are unnecessary, too (otherwise the assignment above would have needed quotes), therefore it would make a lot of sense to lose them here, to make the thing more readable.

t/t3428-rebase-edit-todo.sh
dscho: BTW if that is the case here, i.e. if `todo=.git/rebase-merge/git-rebase-todo`, then it would make more sense to use that hard-coded assignment here. Or maybe even avoid the assignment and use the hard-coded path in the test case below.

dscho:
This looks very good already! I left a couple more comments, hopefully you find them helpful.

dscho:
Oh, I just saw that the commit message needs a little love, too. I started reworking your patch and will report back later (have to take a break now).

dscho:
Here you are: dscho/git@ffb8b1ddb7a

t/t3428-rebase-edit-todo.sh
shicks: I don't know what --git-dir returns, since I'm not terrible familiar with the test harness. If it's possible that --git-dir returns something with spaces then what I had was actually correct: she'll variable assignment is fully parsed before expansion, so that even if `$bar` has a space in its expansion, `foo=$bar` will still assign the whole thing without quotes. When passing to a redirection operator, however, the quotes are required.

But if we can always just assume `.git` is valid then hardcoding makes the most sense, and obviates the need for any of this.

dscho: --git-dir in this particular instance indeed returns .git/. so we can get away with hard-coding it.

In case of submodules or worktrees, this is not the case, but in the test suite, we know that we do not have such a beast, as we implicitly initialize a regular repository.

You are correct about the spaces, of course. I was actually thinking of arbitrary white space (which gets converted into spaces unless quoted).

peff: You don't need double quotes here even with the spaces, because variable assignments are special in bourne shell, and do not go through the extra round of whitespace splitting that command arguments do. So:

```
foo=$(echo look at all these spaces)/spaces_here_would_turn_this_into_a_command
```

is fine. The quotes don't hurt, of course (and I agree that just writing `.git` is fine for the test harness).

dscho: Try this: `foo=$(echo look  at    these       white spaces); echo "$foo"`.

peff: > `foo=$(echo look  at    these       white spaces); echo "$foo"`

That's not the lack of quotes around the variable assignment, though. The extra whitespace is between arguments to `echo`. So `echo` sees it turned into a list and puts a single space between each word. You can see the same effect without a variable assignment:

```
$ echo look  at    these       white spaces
look at these white spaces
```

taweesakteejantuk66:
thanks

dscho:
@shicks can this PR please be closed? The fix made it into git.git, no?

dscho:
For lurkers: this is the commit in git.git: 54fd3243daec7d59394f25846450d83c0012e57c

PR #366, opened 2017-05-29
git-p4 doc. Branch detection final listing simplified

goto1134:
Look at the listing:

```
git init depot
cd depot
git config git-p4.branchList main:branch1
git p4 clone --detect-branches //depot@all .
```

git p4 clone is used with the last dot that points to the current directory. It is easy to miss the dot, so I offer to use git p4 sync instead. It has the same functionality and syntax, except does not need to point the directory, because works in the current one. The result would be like:

```
git init depot
cd depot
git config git-p4.branchList main:branch1
git p4 sync --detect-branches //depot@all
```

### DIFF
--- a/Documentation/git-p4.txt
+++ b/Documentation/git-p4.txt
@@ -434,7 +434,7 @@ occur with:
 git init depot
 cd depot
 git config git-p4.branchList main:branch1
-git p4 clone --detect-branches //depot@all .
+git p4 sync --detect-branches //depot@all
 ----
 
 


dscho:
Thank you for your contribution.

Please add your [sign off](https://git-scm.com/docs/SubmittingPatches#dco) and then submit this patch to the Git mailing list for review (use e.g. [GitGitGadget](https://gitgitgadget.github.io) or [send it manually](https://github.com/git-for-windows/git/blob/master/CONTRIBUTING.md#submit-your-patch)), as the Git project does not accept PRs.

goto1134:
> Thank you for your contribution.
> 
> Please add your [sign off](https://git-scm.com/docs/SubmittingPatches#dco) and then submit this patch to the Git mailing list for review (use e.g. [GitGitGadget](https://gitgitgadget.github.io) or [send it manually](https://github.com/git-for-windows/git/blob/master/CONTRIBUTING.md#submit-your-patch)), as the Git project does not accept PRs.

Thank you.

I submitted a patch via mailing list

PR #372, opened 2017-06-06
send-email: Add more body tag parsing

kees:
The Linux kernel has several body tags[1] for other ways people can be
tagged in patch creation. This adds parsing these to git-send-email so
that more than just CCed people will get CCed.

[1] https://www.kernel.org/doc/html/latest/process/submitting-patches.html#using-reported-by-tested-by-reviewed-by-suggested-by-and-fixes

Signed-off-by: Kees Cook <keescook@chromium.org>

### DIFF
--- a/git-send-email.perl
+++ b/git-send-email.perl
@@ -1580,7 +1580,7 @@ foreach my $t (@files) {
 	# Now parse the message body
 	while(<$fh>) {
 		$message .=  $_;
-		if (/^(Signed-off-by|Cc): ([^>]*>?)/i) {
+		if (/^((?:Signed-off|Acked|Reviewed|Reported|Tested|Suggested)-by|Cc): ([^>]*>?)/i) {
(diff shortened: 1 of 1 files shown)


dscho:
Thank you for your contribution! Unfortunately, the Git project does not accept PRs, therefore this PR lingered for so long.

In the meantime, merge conflicts crept in. Would you mind rebasing on top of the current `master`?

Then, send the patch to the mailing list for review. You can use [GitGitGadget](https://gitgitgadget.github.io), [submitGit](https://submitgit.herokuapp.com) or [send it manually](https://github.com/git-for-windows/git/blob/master/CONTRIBUTING.md#submit-your-patch).

kees:
This functionality was done upstream in ef0cc1df90f6b6c2987ab2db8e0ccf2cfc421edf ("send-email: also pick up cc addresses from -by trailers")

PR #388, opened 2017-07-25
setup: update error message to be more meaningful

sivaraam:
The error message shown when a flag is found when expecting a
filename wasn't clear as it didn't communicate what was wrong
using the 'suitable' words in *all* cases.

Correct case,

        $ git rev-parse commit.c --flags
        commit.c
        --flags
        fatal: bad flag '--flags' used after filename

Incorrect case,

        $ git grep "test" -n
        fatal: bad flag '-n' used after filename

Change the error message to be general and communicative. This results
in the following output,

        $ git rev-parse commit.c --flags
        commit.c
        --flags
        fatal: found flag '--flags' in place of a filename

        $ git grep "test" -n
        fatal: found flag '-n' in place of a filename

Signed-off-by: Kaartic Sivaraam <kaarticsivaraam91196@gmail.com>

### DIFF
--- a/setup.c
+++ b/setup.c
@@ -230,7 +230,7 @@ void verify_filename(const char *prefix,
 		     int diagnose_misspelt_rev)
 {
 	if (*arg == '-')
-		die("bad flag '%s' used after filename", arg);
+		die("found flag '%s' in place of a filename", arg);
 	if (looks_like_pathspec(arg) || check_filename(prefix, arg))
 		return;
 	die_verify_filename(prefix, arg, diagnose_misspelt_rev);


submitgit:
@sivaraam sent this commit (https://github.com/git/git/compare/5800c63717ae35286a1441f14ffff753e01f7e2b...c13dcdd9ebdb4e70d5b44797fd275d3802e140f8) as a patch to the mailing list with [_submitGit_](https://github.com/rtyley/submitgit) - here on [public-inbox](https://public-inbox.org/git/0102015d7ae53b0a-a6505296-9257-4b0d-84d6-2152e17eb070-000000@eu-west-1.amazonses.com/), [MARC](http://marc.info/?i=0102015d7ae53b0a-a6505296-9257-4b0d-84d6-2152e17eb070-000000@eu-west-1.amazonses.com) [](0102015d7ae53b0a-a6505296-9257-4b0d-84d6-2152e17eb070-000000@eu-west-1.amazonses.com)

PR #393, opened 2017-07-29
setup: update error message to be more meaningful

sivaraam:
The error message shown when a flag is found when expecting a
filename wasn't clear as it didn't communicate what was wrong
using the 'suitable' words in *all* cases.

        $ git ls-files
        README.md
        test-file

Correct case,

        $ git rev-parse README.md --flags
        README.md
        --flags
        fatal: bad flag '--flags' used after filename

Incorrect case,

        $ git grep "some random regex" -n
        fatal: bad flag '-n' used after filename

The above case is incorrect as "some random regex" isn't a filename
in this case.

Change the error message to be general and communicative. This results
in the following output,

        $ git rev-parse README.md --flags
        README.md
        --flags
        fatal: option '--flags' must come before non-option arguments

        $ git grep "some random regex" -n
        fatal: option '-n' must come before non-option arguments

Signed-off-by: Kaartic Sivaraam <kaarticsivaraam91196@gmail.com>

### DIFF
--- a/setup.c
+++ b/setup.c
@@ -230,7 +230,7 @@ void verify_filename(const char *prefix,
 		     int diagnose_misspelt_rev)
 {
 	if (*arg == '-')
-		die("bad flag '%s' used after filename", arg);
+		die("option '%s' must come before non-option arguments", arg);
 	if (looks_like_pathspec(arg) || check_filename(prefix, arg))
 		return;
 	die_verify_filename(prefix, arg, diagnose_misspelt_rev);


submitgit:
@sivaraam sent this commit (https://github.com/git/git/compare/5800c63717ae35286a1441f14ffff753e01f7e2b...b7cd847237820812328c9be987441ae75e3aa0e0) as a patch to the mailing list with [_submitGit_](https://github.com/rtyley/submitgit) - here on [public-inbox](https://public-inbox.org/git/0102015d8e5dd2c8-be70b5d5-8a65-4366-8cac-84a0f31ba744-000000@eu-west-1.amazonses.com/), [MARC](http://marc.info/?i=0102015d8e5dd2c8-be70b5d5-8a65-4366-8cac-84a0f31ba744-000000@eu-west-1.amazonses.com) [](0102015d8e5dd2c8-be70b5d5-8a65-4366-8cac-84a0f31ba744-000000@eu-west-1.amazonses.com)

PR #397, opened 2017-08-18
git-new-workdir: prevent cd from writing to stdout

twz123:
Under certain circumstances, `cd` may write to stdout unexpectedly and break
the script. Most notably, this happens when the `CDPATH` environment variable
is set.


twz123:
Argh. Didn't know about `git worktree`...

PR #429, opened 2017-11-08
Use system env version of Perl

hanxue:
To prevent errors like this when running

```
git-svn: Perl lib version (5.24.1) doesn't match executable '/usr/bin/perl' version (5.18.2) at /usr/local/Cellar/perl/5.24.1//lib/perl5/5.24.1/darwin-thread-multi-2level/Config.pm line 62.
```

### DIFF
--- a/git-svn.perl
+++ b/git-svn.perl
@@ -1,4 +1,4 @@
-#!/usr/bin/perl
+#!/usr/bin/env perl
 # Copyright (C) 2006, Eric Wong <normalperson@yhbt.net>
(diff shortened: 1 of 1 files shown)


dscho:
What about all the other `.perl` scripts? Would they not need to be patched, too?

Besides, the `Makefile` already has a different solution for this kind of problem: set `PERL_PATH` (see https://github.com/git/git/blob/a50a68d537980509a3646bba55a4ea78fb7807e1/Makefile#L296).

PR #440, opened 2017-12-11
convert: add support for 'encoding' attribute

larsxschneider:
PR only for preliminary code review 😄 

---

Git and its tools (e.g. git diff) expect all text files in UTF-8
encoding. Git will happily accept content in all other encodings, too,
but it might not be able to process the text (e.g. view diff, change
line endings etc.).

Add an attribute to tell Git what encoding the user expects for a given
file. This way Git can convert the content to its internal UTF-8
representation.

### DIFF
--- a/Documentation/gitattributes.txt
+++ b/Documentation/gitattributes.txt
@@ -312,6 +312,14 @@ number of pitfalls:
   internal contents as UTF-8 and try to convert it to UTF-16 on checkout.
   That operation will fail and cause an error.
 
+- Reencoding content to non-UTF encodings can cause errors as the
+  conversion might not be UTF-8 round trip safe. If you suspect your
+  encoding to not be round trip safe, then add it to
+  `core.checkRoundtripEncoding` to make Git check the round trip
+  encoding (see linkgit:git-config[1]). SHIFT-JIS (Japanese character
+  set) is known to have round trip issues with UTF-8 and is checked by
+  default.
+
 - Reencoding content requires resources that might slow down certain
   Git operations (e.g 'git checkout' or 'git add').
 
--- a/Documentation/gitattributes.txt
+++ b/Documentation/gitattributes.txt
@@ -272,6 +272,86 @@ few exceptions.  Even though...
   catch potential problems early, safety triggers.
 
 
+`working-tree-encoding`
+^^^^^^^^^^^^^^^^^^^^^^^
+
+Git recognizes files encoded in ASCII or one of its supersets (e.g.
+UTF-8, ISO-8859-1, ...) as text files. Files encoded in certain other
+encodings (e.g. UTF-16) are interpreted as binary and consequently
+built-in Git text processing tools (e.g. 'git diff') as well as most Git
+web front ends do not visualize the contents of these files by default.
+
+In these cases you can tell Git the encoding of a file in the working
+directory with the `working-tree-encoding` attribute. If a file with this
+attribute is added to Git, then Git reencodes the content from the
+specified encoding to UTF-8. Finally, Git stores the UTF-8 encoded
+content in its internal data structure (called "the index"). On checkout
+the content is reencoded back to the specified encoding.
+
+Please note that using the `working-tree-encoding` attribute may have a
+number of pitfalls:
+
+- Alternative Git implementations (e.g. JGit or libgit2) and older Git
+  versions (as of March 2018) do not support the `working-tree-encoding`
+  attribute. If you decide to use the `working-tree-encoding` attribute
+  in your repository, then it is strongly recommended to ensure that all
+  clients working with the repository support it.
+
+  For example, Microsoft Visual Studio resources files (`*.rc`) or
+  PowerShell script files (`*.ps1`) are sometimes encoded in UTF-16.
+  If you declare `*.ps1` as files as UTF-16 and you add `foo.ps1` with
+  a `working-tree-encoding` enabled Git client, then `foo.ps1` will be
+  stored as UTF-8 internally. A client without `working-tree-encoding`
+  support will checkout `foo.ps1` as UTF-8 encoded file. This will
+  typically cause trouble for the users of this file.
+
+  If a Git client, that does not support the `working-tree-encoding`
+  attribute, adds a new file `bar.ps1`, then `bar.ps1` will be
+  stored "as-is" internally (in this example probably as UTF-16).
+  A client with `working-tree-encoding` support will interpret the
+  internal contents as UTF-8 and try to convert it to UTF-16 on checkout.
+  That operation will fail and cause an error.
+
+- Reencoding content requires resources that might slow down certain
+  Git operations (e.g 'git checkout' or 'git add').
+
+Use the `working-tree-encoding` attribute only if you cannot store a file
+in UTF-8 encoding and if you want Git to be able to process the content
+as text.
+
+As an example, use the following attributes if your '*.ps1' files are
+UTF-16 encoded with byte order mark (BOM) and you want Git to perform
+automatic line ending conversion based on your platform.
+
+------------------------
+*.ps1		text working-tree-encoding=UTF-16
+------------------------
+
+Use the following attributes if your '*.ps1' files are UTF-16 little
+endian encoded without BOM and you want Git to use Windows line endings
+in the working directory. Please note, it is highly recommended to
+explicitly define the line endings with `eol` if the `working-tree-encoding`
+attribute is used to avoid ambiguity.
+
+------------------------
+*.ps1		text working-tree-encoding=UTF-16LE eol=CRLF
+------------------------
+
+You can get a list of all available encodings on your platform with the
+following command:
+
+------------------------
+iconv --list
+------------------------
+
+If you do not know the encoding of a file, then you can use the `file`
+command to guess the encoding:
+
+------------------------
+file foo.ps1
+------------------------
+
+
 `ident`
 ^^^^^^^
 
--- a/convert.c
+++ b/convert.c
@@ -193,30 +193,30 @@ static enum eol output_eol(enum crlf_action crlf_action)
 	return core_eol;
 }
 
-static void check_safe_crlf(const char *path, enum crlf_action crlf_action,
+static void check_global_conv_flags_eol(const char *path, enum crlf_action crlf_action,
 			    struct text_stat *old_stats, struct text_stat *new_stats,
-			    enum safe_crlf checksafe)
+			    int conv_flags)
 {
 	if (old_stats->crlf && !new_stats->crlf ) {
 		/*
 		 * CRLFs would not be restored by checkout
 		 */
-		if (checksafe == SAFE_CRLF_WARN)
+		if (conv_flags & CONV_EOL_RNDTRP_DIE)
+			die(_("CRLF would be replaced by LF in %s."), path);
+		else if (conv_flags & CONV_EOL_RNDTRP_WARN)
 			warning(_("CRLF will be replaced by LF in %s.\n"
 				  "The file will have its original line"
 				  " endings in your working directory."), path);
-		else /* i.e. SAFE_CRLF_FAIL */
-			die(_("CRLF would be replaced by LF in %s."), path);
 	} else if (old_stats->lonelf && !new_stats->lonelf ) {
 		/*
 		 * CRLFs would be added by checkout
 		 */
-		if (checksafe == SAFE_CRLF_WARN)
+		if (conv_flags & CONV_EOL_RNDTRP_DIE)
+			die(_("LF would be replaced by CRLF in %s"), path);
+		else if (conv_flags & CONV_EOL_RNDTRP_WARN)
 			warning(_("LF will be replaced by CRLF in %s.\n"
 				  "The file will have its original line"
 				  " endings in your working directory."), path);
-		else /* i.e. SAFE_CRLF_FAIL */
-			die(_("LF would be replaced by CRLF in %s"), path);
 	}
 }
 
@@ -268,7 +268,7 @@ static int will_convert_lf_to_crlf(size_t len, struct text_stat *stats,
 static int crlf_to_git(const struct index_state *istate,
 		       const char *path, const char *src, size_t len,
 		       struct strbuf *buf,
-		       enum crlf_action crlf_action, enum safe_crlf checksafe)
+		       enum crlf_action crlf_action, int conv_flags)
 {
 	struct text_stat stats;
 	char *dst;
@@ -298,12 +298,12 @@ static int crlf_to_git(const struct index_state *istate,
 		 * unless we want to renormalize in a merge or
 		 * cherry-pick.
 		 */
-		if ((checksafe != SAFE_CRLF_RENORMALIZE) &&
+		if ((!(conv_flags & CONV_EOL_RENORMALIZE)) &&
 		    has_crlf_in_index(istate, path))
 			convert_crlf_into_lf = 0;
 	}
-	if ((checksafe == SAFE_CRLF_WARN ||
-	    (checksafe == SAFE_CRLF_FAIL)) && len) {
+	if (((conv_flags & CONV_EOL_RNDTRP_WARN) ||
+	     ((conv_flags & CONV_EOL_RNDTRP_DIE) && len))) {
 		struct text_stat new_stats;
 		memcpy(&new_stats, &stats, sizeof(new_stats));
 		/* simulate "git add" */
@@ -316,7 +316,7 @@ static int crlf_to_git(const struct index_state *istate,
 			new_stats.crlf += new_stats.lonelf;
 			new_stats.lonelf = 0;
 		}
-		check_safe_crlf(path, crlf_action, &stats, &new_stats, checksafe);
+		check_global_conv_flags_eol(path, crlf_action, &stats, &new_stats, conv_flags);
 	}
 	if (!convert_crlf_into_lf)
 		return 0;
@@ -1129,7 +1129,7 @@ const char *get_convert_attr_ascii(const char *path)
 
 int convert_to_git(const struct index_state *istate,
 		   const char *path, const char *src, size_t len,
-                   struct strbuf *dst, enum safe_crlf checksafe)
+		   struct strbuf *dst, int conv_flags)
 {
 	int ret = 0;
 	struct conv_attrs ca;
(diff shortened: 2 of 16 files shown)


convert.c
```diff
@@ -256,6 +257,117 @@ static int will_convert_lf_to_crlf(size_t len, struct text_stat *stats,
 
 }
 
+static struct encoding {
+	const char *name;
+	iconv_t *to_git;
+	iconv_t *to_worktree;
+	struct encoding *next;
+} *encoding, **encoding_tail;
+static const char *default_encoding = "utf-8";
```
pluehne: According to the [`iconv_open` `man` pages]((http://man7.org/linux/man-pages/man3/iconv_open.3.html)):

> For the GNU C library, the permitted values are listed by the `iconv --list` command …

This list contains `UTF-8` in uppercase but not lowercase. Could that be an issue?

larsxschneider: AFAIK the encodings are case insensitive.

pluehne: Okay, I wast just wondering.

convert.c
```diff
@@ -256,6 +257,117 @@ static int will_convert_lf_to_crlf(size_t len, struct text_stat *stats,
...
+		return 1;
+	}
+
+	if (!enc->to_git || !enc->to_worktree) {
+		enc->to_git = xcalloc(1, sizeof(iconv_t));
+		enc->to_git = iconv_open(default_encoding, encoding->name);
+		enc->to_worktree = xcalloc(1, sizeof(iconv_t));
+		enc->to_worktree = iconv_open(encoding->name, default_encoding);
+
+		if (enc->to_git == (iconv_t) - 1 ||
```
pluehne: I’d write all of these occurrences either `(iconv_t)(-1)` or at least `(iconv_t)-1` to make clear that `-` is a unary operator here (this looked confusing at a first glance).

convert.c
```diff
@@ -256,6 +257,117 @@ static int will_convert_lf_to_crlf(size_t len, struct text_stat *stats,
...
 
+static struct encoding {
+	const char *name;
+	iconv_t *to_git;
+	iconv_t *to_worktree;
+	struct encoding *next;
+} *encoding, **encoding_tail;
+static const char *default_encoding = "utf-8";
+
+static int encode_to_git(const char *path, const char *src, size_t src_len,
```
pluehne: Some short documentation on the meaning of the return values would be nice.

convert.c
```diff
@@ -969,6 +1081,30 @@ static int ident_to_worktree(const char *path, const char *src, size_t len,
...
 
+static struct encoding *git_path_check_encoding(struct attr_check_item *check)
+{
+	const char *value = check->value;
+	struct encoding *enc;
+
+	if (ATTR_UNSET(value))
+		return NULL;
+
+	for (enc = encoding; enc; enc = enc->next)
```
pluehne: If the code isn’t C98 and that’s not mandated by style otherwise, `struct encoding *enc = encoding` could be put directly into the for loop’s init statement.

convert.c
```diff
@@ -256,6 +257,117 @@ static int will_convert_lf_to_crlf(size_t len, struct text_stat *stats,
...
+#ifndef NO_ICONV
+	char *dst, *re_src;
+	int dst_len, re_src_len;
+
+	if (!enc || (src && !src_len))
+		return 0;
+
+	/*
+	 * Looks like we got called from "would_convert_to_git()".
+	 * Since we have an encoding we must assume that we would encode
```
pluehne: This comment confused me a bit. Suggestion:

> There is no input string or buffer, but an encoding was specified. Apparently, something was intended to be encoded, so report this function call as invalid.

convert.c
```diff
@@ -256,6 +257,117 @@ static int will_convert_lf_to_crlf(size_t len, struct text_stat *stats,
...
+static const char *default_encoding = "utf-8";
+
+static int encode_to_git(const char *path, const char *src, size_t src_len,
+			 struct strbuf *buf, struct encoding *enc)
+{
+#ifndef NO_ICONV
+	char *dst, *re_src;
+	int dst_len, re_src_len;
+
+	if (!enc || (src && !src_len))
```
pluehne: Maybe the following comment would be useful:

> No encoding is specified or the source string is empty. Return successfully, as nothing has to be done.

convert.c
larsxschneider: Good call! 👍

convert.c
```diff
@@ -256,6 +257,117 @@ static int will_convert_lf_to_crlf(size_t len, struct text_stat *stats,
...
+	 * Looks like we got called from "would_convert_to_git()".
+	 * Since we have an encoding we must assume that we would encode
+	 * the content.
+	 */
+	if (!buf && !src) {
+		return 1;
+	}
+
+	if (!enc->to_git || !enc->to_worktree) {
+		enc->to_git = xcalloc(1, sizeof(iconv_t));
```
pluehne: This still looks like a double allocation to me.

First of all, this doesn’t need to be cleared (`xalloc` would also do it), because the pointer is reassigned in the next line.

However, [`iconv_open` allocates the memory for you](http://man7.org/linux/man-pages/man3/iconv_open.3.html) if I read the `man` pages correctly. [Examples](https://www.gnu.org/software/libc/manual/html_node/iconv-Examples.html) also show that allocating this before isn’t necessary.

convert.c
```diff
@@ -256,6 +257,117 @@ static int will_convert_lf_to_crlf(size_t len, struct text_stat *stats,
 
 }
 
+static struct encoding {
+	const char *name;
+	iconv_t *to_git;
```
pluehne: Maybe this could briefly described by a comment like this:

> iconv handle for conversion to Git's canonical character representation

convert.c
```diff
@@ -256,6 +257,117 @@ static int will_convert_lf_to_crlf(size_t len, struct text_stat *stats,
...
+	iconv_t *to_worktree;
```
pluehne: Maybe this could briefly described by a comment like this:

> iconv handle for conversion back to the user-defined character encoding

convert.c
```diff
@@ -256,6 +257,117 @@ static int will_convert_lf_to_crlf(size_t len, struct text_stat *stats,
...
+		enc->to_git = iconv_open(default_encoding, encoding->name);
+		enc->to_worktree = xcalloc(1, sizeof(iconv_t));
+		enc->to_worktree = iconv_open(encoding->name, default_encoding);
+
+		if (enc->to_git == (iconv_t)-1 ||
+		    enc->to_worktree == (iconv_t)-1)
+			error(_("unsupported encoding %s"), encoding->name);
+	}
+
+	/* Do nothing if the encoding is not supported */
```
pluehne: Why not report an error in this case?

convert.c
larsxschneider: `1` indicates successful encoding, `0` is indicates and error or no encoding... I guess that is the standard in this file as similar functions don't explicitly mention it.

convert.c
pluehne: Ah, I see get it now. While `iconv_open` indeed performs an allocation, this is opaque within the `iconv_t` data structure. In other words, your `iconv_t *` is one more level of indirection, which is why manual allocation appears to be necessary.

pluehne: However, `xcalloc` could still be `xalloc`, because the pointer doesn’t need to be nulled (it’s assigned immediately afterward).

I also wonder whether you can’t simply make the `iconv_t *` in the `encoding` struct plain `iconv_t`, assuming that copying `iconv_t` isn’t expensive.

convert.c
larsxschneider: Git wants its variables defined first.

convert.c
larsxschneider: How about:
```
	/*
	 * Looks like we got called from "would_convert_to_git()".
	 * This means Git wants to know if it would encode the content.
	 * Let's answer with "yes" since an encoding was specified.
	 */
```

pluehne: Okay, I get it now 😃. This comment is much better from my point of view!

convert.c
```diff
@@ -256,6 +257,117 @@ static int will_convert_lf_to_crlf(size_t len, struct text_stat *stats,
...
+	/*
+	 * Looks like we got called from "would_convert_to_git()".
+	 * Since we have an encoding we must assume that we would encode
+	 * the content.
+	 */
+	if (!buf && !src) {
+		return 1;
+	}
+
+	if (!enc->to_git || !enc->to_worktree) {
```
pluehne: I’d implement this more defensively and split the two cases into separate if statements. The reason is just that if at some point, `to_git` and `to_worktree` get allocated separately in the future, one of them would get allocated twice and leak for this reason.

convert.c
larsxschneider: How about this?
```
	/*
	 * No encoding is specified or there is nothing to encode.
	 * Tell the caller that the content was not modified.
	 */
```

convert.c
```diff
@@ -256,6 +257,117 @@ static int will_convert_lf_to_crlf(size_t len, struct text_stat *stats,
...
+		enc->to_worktree = xcalloc(1, sizeof(iconv_t));
+		enc->to_worktree = iconv_open(encoding->name, default_encoding);
+
+		if (enc->to_git == (iconv_t)-1 ||
+		    enc->to_worktree == (iconv_t)-1)
+			error(_("unsupported encoding %s"), encoding->name);
+	}
+
+	/* Do nothing if the encoding is not supported */
+	if (enc->to_git == (iconv_t)-1 ||
```
pluehne: This should be `*enc->to_git`.

convert.c
pluehne: That makes it very clear 👍!

convert.c
larsxschneider: Argh. Spot on! Thank you!

convert.c
larsxschneider: Obsolete.

convert.c
larsxschneider: ok!

convert.c
pluehne: Okay, the additional check for null pointers might be unnecessary, because this is checked above.

convert.c
larsxschneider: because we reported the error already above, I will extend the comment with this info

convert.c
larsxschneider: Good call 👍

convert.c
```diff
@@ -256,6 +257,121 @@ static int will_convert_lf_to_crlf(size_t len, struct text_stat *stats,
...
+		 * we would try to re-encode to the original encoding. This
+		 * would fail and we would leave the user with a messed-up
+		 * working tree. Let's try to avoid this by screaming loud.
+		 */
+		die(_("failed to encode '%s' from %s to %s"),
+			path, enc->name, default_encoding);
+
+	/* Encode dst back to ensure no information is lost */
+	re_src = reencode_string_iconv(dst, dst_len, enc->to_worktree, &re_src_len);
+	if (!re_src || strcmp(src, re_src)) {
```
pluehne: Performing this round-trip check in production code could be expensive, I fear. We should check in which cases `iconv` asserts that round-trip conversion is safe, and only allow these cases.

convert.c
```diff
@@ -256,6 +257,121 @@ static int will_convert_lf_to_crlf(size_t len, struct text_stat *stats,
...
+	 * No encoding is specified or there is nothing to encode.
+	 * Tell the caller that the content was not modified.
+	 */
+	if (!enc || (src && !src_len))
+		return 0;
+
+	/*
+	 * Looks like we got called from "would_convert_to_git()".
+	 * This means Git wants to know if it would encode (= modify!)
+	 * the content. Let's answer with "yes" since an encoding was
```
pluehne: There should be a comma before “since” to make clear that this is a conjunction and not preposition.

convert.c
```diff
@@ -256,6 +257,121 @@ static int will_convert_lf_to_crlf(size_t len, struct text_stat *stats,
...
+	 * specified.
+	 */
+	if (!buf && !src)
+		return 1;
+
+	if (enc->to_git == no_conversion) {
```
pluehne: I’m thinking that `no_conversion` might be a bit misleading. For instance, `-1` is returned upon errors by `iconv_open`, which doesn’t imply that “no conversion” is intended.

Another proposal would be `invalid_conversion`. Maybe something like `iconv_t_invalid` might be even more concise (as this really just is an `iconv_t` value and nothing else), however, this might clash with future naming decisions in upstream `iconv`.

t/t0028-encoding.sh
```diff
@@ -0,0 +1,60 @@
...
+
+test_description='encoding conversion via gitattributes'
+
+. ./test-lib.sh
+
+test_expect_success 'setup test repo' '
+
+	text="hallo there!\ncan you read me?" &&
+
+	echo "*.utf16 text encoding=utf-16" >.gitattributes &&
```
pluehne: The spacing around `>` redirects is inconcistent in this test file. I’d personally put spaces before and after each `>`.

convert.c
pluehne: Okay!

convert.c
larsxschneider: 👍

convert.c
larsxschneider: `invalid_conversion` 👍

t/t0028-encoding.sh
larsxschneider: No space after `>` is Git style 😉

pluehne: Okay, but below there are cases where there is no space before `>`, and here, there is one.

t/t0028-encoding.sh
```diff
@@ -0,0 +1,60 @@
...
+	printf "garbage" >t.garbage &&
+	git add t.garbage 2>&1 >/dev/null |
+		grep "error: unsupported encoding" | grep garbage &&
+
+	# cleanup
+	git reset --hard HEAD
+'
+
+test_expect_success 'fail if files with invalid encoding are added' '
+	printf "\0\0h\0a">error.utf16 &&
```
pluehne: See above ☝️.

convert.c
```diff
@@ -256,6 +257,128 @@ static int will_convert_lf_to_crlf(size_t len, struct text_stat *stats,
...
+	if (!re_src || strcmp(src, re_src)) {
+		die(_("encoding '%s' from %s to %s and back is not the same"),
+			path, enc->name, default_encoding);
+	}
+	free(re_src);
+
+	strbuf_attach(buf, dst, dst_len, dst_len + 1);
+	return 1;
+#else
+	error(_("cannot encode '%s' from %s to %s because "
```
pluehne: Why `error` and not `die` here?

convert.c
```diff
@@ -969,6 +1092,32 @@ static int ident_to_worktree(const char *path, const char *src, size_t len,
...
+	struct encoding *enc;
+
+	if (ATTR_UNSET(value))
+		return NULL;
+
+	for (enc = encoding; enc; enc = enc->next)
+		if (!strcmp(value, enc->name))
+			return enc;
+
+	/* Don't re-encode to the default */
```
pluehne: According to the [Chicago Manual of Style](http://www.chicagomanualofstyle.org/16/images/ch07_tab01.pdf#page9) and the Merriam-Webster dictionary, prefixes like `multi`, `anti`, `re`, `non`, etc. shouldn’t be separated with a hyphen.

t/t0028-encoding.sh
larsxschneider: 👍

convert.c
larsxschneider: Well, then a user would not be able to checkout content at all.
This way we can give the user at least the UTF-8 encoded version, no?

pluehne: I see. Maybe add the following comment:

> If iconv is missing, warn the user but return a UTF-8-encoded version of the file to not interrupt checkout.

Documentation/gitattributes.txt
```diff
@@ -146,6 +146,33 @@ Unspecified::
 Any other value causes Git to act as if `text` has been left
 unspecified.
 
+`encoding`
+^^^^^^^^^^
+
+By default Git assumes UTF-8 encoding for text files.  The `encoding`
```
pluehne: Minor nagging about typography: The documentation doesn’t seem to use [English spacing](https://en.wikipedia.org/wiki/History_of_sentence_spacing#French_and_English_spacing) (two spaces between sentences) consistently.

While [many consider double sentence spacing outdated](https://en.wikipedia.org/wiki/History_of_sentence_spacing#Movement_to_single_sentence_spacing), I don’t know what the standard within Git is. This seems to be inconsistent right now, but maybe this should be checked. In case of doubt, I’d suggest modern single sentence spacing.

convert.c
```diff
@@ -969,6 +1099,32 @@ static int ident_to_worktree(const char *path, const char *src, size_t len,
...
+	/* Don't reencode to the default */
+	if (!strcasecmp(value, default_encoding))
+		return NULL;
+
+	enc = xcalloc(1, sizeof(struct convert_driver));
+	enc->name = xstrdup(value);
+	enc->to_git = invalid_conversion;
+	enc->to_worktree = invalid_conversion;
+	*encoding_tail = enc;
+	encoding_tail = &(enc->next);
```
pluehne: What’s the need for the extra parentheses here?

convert.c
```diff
@@ -969,6 +1099,32 @@ static int ident_to_worktree(const char *path, const char *src, size_t len,
 	return 1;
 }
 
+static struct encoding *git_path_check_encoding(struct attr_check_item *check)
```
pluehne: I wouldn’t necessarily call this a `check` function, because it does way more than that. First, the output of the function is the encoding, so it’s more or less a getter function. Second, it’s lazy initialization in that it creates missing encodings on the fly.

An idiom for naming such functions is `get_or_create`, so maybe, this could be renamed to `git_path_get_or_create_encoding` or so.

convert.c
```diff
@@ -256,6 +257,135 @@ static int will_convert_lf_to_crlf(size_t len, struct text_stat *stats,
...
+static struct encoding {
+	const char *name;
+	iconv_t to_git;       /* conversion to Git's canonical encoding (UTF-8) */
+	iconv_t to_worktree;  /* conversion to user-defined encoding */
+	struct encoding *next;
+} *encoding, **encoding_tail;
+static const char *default_encoding = "utf-8";
+static iconv_t invalid_conversion = (iconv_t)-1;
+
+static int encode_to_git(const char *path, const char *src, size_t src_len,
```
pluehne: I wonder whether this should maybe be called `encode_for_git` (as well as `encode_for_worktree`). `encode_to_git` could imply that Git has or is a custom encoding format, while this really is just UTF-8.

Documentation/gitattributes.txt
larsxschneider: I think @gitster prefers that. I dunno.

convert.c
larsxschneider: I guess this follows the other function names in the file ;-)

convert.c
larsxschneider: maybe. but this is kind of in line with the other functions I think

larsxschneider:
Merged as https://github.com/git/git/commit/1ac0ce4d32ab7a3546e7e84a562625576208c7db

PR #446, opened 2017-12-22
Fix urlencode format string on signed char.

j-d-r:
Patch for special chars in password for git-crendential.
Urlencode seems to be used only in credential-store and HTTP proxy, only tested with git-credential.

---

Git credential fails with special char in password.
remote: Invalid username or password.
fatal: Authentication failed for

File ~/.git-credential contains badly urlencoded characters
%ffffffXX%ffffffYY instead of %XX%YY.

Add a cast to an unsigned char to fix urlencode use of %02x
on a char.

Signed-off-by: Julien Dusser <julien.dusser@free.fr>


j-d-r:
Submitgit is too heavy, email sent with git send-email instead.

PR #471, opened 2018-03-16
mergetools: enable tortoisemerge as difftool

volleyms:
TortoiseMerge/TortoiseGitMerge can also be used to view and edit file differences.

This change allows configuring tortoisemerge not only as mergetool but as difftool as well.

### DIFF
--- a/mergetools/tortoisemerge
+++ b/mergetools/tortoisemerge
@@ -1,5 +1,13 @@
-can_diff () {
-	return 1
+diff_cmd () {
+	basename="$(basename "$merge_tool_path" .exe)"
+	if test "$basename" = "tortoisegitmerge"
+	then
+		"$merge_tool_path" \
+			-mine "$REMOTE" -base "$LOCAL"
+	else
+		"$merge_tool_path" \
+			-mine:"$REMOTE" -base:"$LOCAL"
+	fi
 }
 
 merge_cmd () {


submitgit:
@volleyms sent this commit (https://github.com/git/git/compare/0afbf6caa5b16dcfa3074982e5b48e27d452dbbb...0432e450e8e7babc65cdce7b14b7c37205b14464) as a patch to the mailing list with [_submitGit_](https://github.com/rtyley/submitgit) - here on [public-inbox](https://public-inbox.org/git/010201622e3fc0a8-2804475a-8c37-4978-b5dd-b0fb254bfed2-000000@eu-west-1.amazonses.com/), [MARC](http://marc.info/?i=010201622e3fc0a8-2804475a-8c37-4978-b5dd-b0fb254bfed2-000000@eu-west-1.amazonses.com) [](010201622e3fc0a8-2804475a-8c37-4978-b5dd-b0fb254bfed2-000000@eu-west-1.amazonses.com)

dscho:
I could imagine that this mail has not seen any replies because the commit message could be improved, and the [sign off](https://git-scm.com/docs/SubmittingPatches#dco) is missing.

dscho:
@volleyms ping?

dscho:
@volleyms ping?

dscho:
@volleyms gentle ping?

volleyms:
Hi... sorry for not replying so long.

I only wanted to share some small idea... did not know that it is that complicated to contribute... and though I followed the instructions I found and everything was green here, there is nothing happening, and probably because of an IMHO meaningless line in the commit message which I did not see that it was mentioned in the instructions... honestly?

dscho:
@volleyms please don't take it personally that nobody reacted to your contribution yet. Just like you, everybody is busy.

Would you mind amending the commit message with your [`Signed-off-by:` line](https://git-scm.com/docs/SubmittingPatches#sign-off), wrap the paragraph so that the lines are <= 76 columns per line, and send iteration v2?

Even long-time contributors often send half a dozen or more iterations of their patches and patch series, the endeavor here is to make the patches as good as we can _collectively_.

Speaking of patches... I see that the `can_diff` function was removed. I guess most reviewers will be as puzzled about that as I am, so I think it would be a good idea to talk about that in the commit message.

dscho:
> I see that the `can_diff` function was removed. I guess most reviewers will be as puzzled about that as I am, so I think it would be a good idea to talk about that in the commit message.

Hmm. I would have hoped that you find it in you to answer this question.

volleyms:
I don't know where there was a question... but, nevertheless, I will try to do my best as soon as I find the time...

dscho:
> I don't know where there was a question...

It was very implicit. I said "I see that the can_diff function was removed. I guess most reviewers will be as puzzled about that as I am, so I think it would be a good idea to talk about that in the commit message."

What I did not say explicitly is that this puzzles me. I don't understand this. Could you clarify? And then put the clarification into the commit message?

volleyms:
I will try...

dscho:
> I see that the can_diff function was removed.
> [...]
> I don't understand this. Could you clarify? And then put the clarification into the commit message?

@volleyms could you, please?

dscho:
@volleyms ping?

dscho:
@volleyms I really, really, really want to know why it is okay to remove the `can_diff` function in your patch. *Really*. Could you _please_ respond? Pretty please? (I should not even need to beg, you know?)

dscho:
I have no idea how I deserve that silent treatment. This is rude.

volleyms:
No offense, just no time/interest/priority on this, currently.

As far as I could observe, the `can_diff()` method is not necessary. `difftool` and `mergetool` work as expected.

dscho:
> As far as I could observe, the `can_diff()` method is not necessary. `difftool` and `mergetool` work as expected.

That type of reasoning does not make me very confident about the patch.

volleyms:
Fine for me... I use my patch, and others use TortoiseSVN as diff tool only...

SHwareSystemsDev:
It appears that, before an internal diff was implemented, the can_diff() did testing whether an external diff was available. I suspect it wasn't removed in case more reasons a diff might not be doable were going to be tested for, but this hasn't been done yet.

Sent from Yahoo Mail on Android 
 
  On Fri, Oct 25, 2019 at 5:14 PM, Johannes Schindelin<notifications@github.com> wrote:   

I see that the can_diff function was removed.
[...]
I don't understand this. Could you clarify? And then put the clarification into the commit message?


@volleyms could you, please?

dscho:
> It appears that, before an internal diff was implemented, the can_diff() did testing whether an external diff was available. I suspect it wasn't removed in case more reasons a diff might not be doable were going to be tested for, but this hasn't been done yet.

That's a good analysis. With this reasoning, I would actually want to drop that part of the diff, as it would need a good argument to remove the `can_diff` function (and so far, I did not see anything remotely resembling a convincing argument to that end).

SHwareSystemsDev:
You would need to re-add the closing brace, new code uses the original to complete it. Also, a search on whether there are still references to can_diff() in other files may be warranted. If none left then it probably can be deleted anyways, and from declaring header file. If there are some those conditionals possibly could have more optimal code.

 
  On Mon, Dec 30, 2019 at 2:01 PM, Johannes Schindelin<notifications@github.com> wrote:   

It appears that, before an internal diff was implemented, the can_diff() did testing whether an external diff was available. I suspect it wasn't removed in case more reasons a diff might not be doable were going to be tested for, but this hasn't been done yet.


That's a good analysis. With this reasoning, I would actually want to drop that part of the diff, as it would need a good argument to remove the can_diff function (and so far, I did not see anything remotely resembling a convincing argument to that end).

gerhardol:
I see no reason to why 'tortoisemerge' was blocked as a difftool initially. Maybe there were different exe when this was added 10-15 years ago.
It works great, it is my preferred difftool.

I have asked on the TortoiseGit mailinglist if they know the original reason.

That the default tool overrides the explicit configuration is very confusing.

I had done the same patch on my own and will submit a new PR referencing this PR, hopefully getting some explanation of the reasoning.

gerhardol:
A number of related issues related here:

1. tortoisemerge is usable as a difftool, but disabled.
This is the only mergetool configured in this way.
When TortoiseMerge was added in b98c212a9fd97300a8cf39015b8d693f319fd71c in 2009, the commit comment said that a BASE was required. I do not know if it was required then, but is not required now. (TortoiseMerge supports 3-way merge, maybe it was related to that?) I believe I used TortoiseGit and diff at that time, so the commit comment may have been incorrect.
The TortoiseGit maintainer has no recollection on why it was configured this way. 
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/tortoisegit-users/4umDT83oevg
 
This PR fixes this part of the problem, I will try to motivate it better and provide sign-off etc.

2. A tool matching Git builtin will override the user configuration
https://github.com/git/git/blob/main/git-mergetool--lib.sh#L192
If diff.<tool>.cmd (and similar for merge) is set, can_diff() (and can_merge()) should return 0.
https://github.com/git/git/blob/main/git-mergetool--lib.sh#L123 should override the functions.
I may try to make a patch for that too.

3. Merge tools are always listed as available as difftools also if can_read() is defined to 1
https://github.com/git/git/blob/main/git-mergetool--lib.sh#L308
When 1. is fixed, this will not have any effect, will maybe add a comment/TODO

--
Related - where I kind of started

4. difftool --tool-help parsable format
The format is not easy to parse right now, a simpler format would simplify for GUI tools
Some --porcelain format, use --raw also for difftool (and stretch the usage) or rather a new option like --tool-help-raw to give an output like "toolname1\0cmd\ntollname2\0cmd2\n"

PR #472, opened 2018-03-16
Allow use of TLS 1.3. Done during IETF 101 hackathon.

loganaden:


dscho:
This patch made it into `master` as d81b651f56060038a1547f4beb949122533a8165

PR #479, opened 2018-04-04
fix git bash-completion locale bug on macOS

fritz-c:
If there are any variables with non-ASCII characters in the environment displayed by `set`, the following error will be thrown:
```sed: RE error: illegal byte sequence```

This fix ensures that the sed call performs its job without throwing errors.

You can reproduce this bug by adding the following exports to your .bash_profile before the bash-completion script is called:
```sh
export LC_ALL=ja_JP.UTF-8
export HISTTIMEFORMAT="%h月%d日 %H:%M:%S> "
```

Unfortunately I don't have the time to read the 500-line contribution guideline file and add github integrations to send patch files I don't know how to make to a mailing list. So I'm just leaving this pull request here as an FYI.

### DIFF
--- a/contrib/completion/git-completion.bash
+++ b/contrib/completion/git-completion.bash
@@ -282,7 +282,7 @@ __gitcomp ()
 
 # Clear the variables caching builtins' options when (re-)sourcing
 # the completion script.
-unset $(set |sed -ne 's/^\(__gitcomp_builtin_[a-zA-Z0-9_][a-zA-Z0-9_]*\)=.*/\1/p') 2>/dev/null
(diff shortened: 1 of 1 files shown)


iammattcoleman:
An improved fix (which entirely avoids sed for bash shells) has been proposed on the mailing list: https://public-inbox.org/git/20180413103005.671-1-szeder.dev@gmail.com/T/#ma9234b4bd13f1de46cd3c37b3a2c9e8ae3ef9e5f

dscho:
> An improved fix (which entirely avoids `sed` for bash shells) has been proposed on the mailing list: https://public-inbox.org/git/20180413103005.671-1-szeder.dev@gmail.com/T/#ma9234b4bd13f1de46cd3c37b3a2c9e8ae3ef9e5f

And that fix made it into `master`: 94408dc71c0410a385193171cc79d230920b85fa

PR #484, opened 2018-04-16
Bisect first parent patch

tiagonbotelho:

### DIFF
--- a/bisect.c
+++ b/bisect.c
@@ -34,7 +34,7 @@ static const char *term_good;
  * We care just barely enough to avoid recursing for
  * non-merge entries.
  */
-static int count_distance(struct commit_list *entry, int max_parents)
+static int count_distance(struct commit_list *entry, int first_parent_only)
 {
 	int nr = 0;
 
@@ -49,11 +49,10 @@ static int count_distance(struct commit_list *entry, int max_parents)
 		commit->object.flags |= COUNTED;
 		p = commit->parents;
 		entry = p;
-		if (p) {
-			int n = max_parents - 1;
+		if (p && !first_parent_only) {
 			p = p->next;
-			while (p && n-- > 0) {
-				nr += count_distance(p, max_parents);
+			while (p) {
+				nr += count_distance(p, first_parent_only);
 				p = p->next;
 			}
 		}
@@ -83,15 +82,16 @@ static inline void weight_set(struct commit_list *elem, int weight)
 	*((int*)(elem->item->util)) = weight;
 }
 
-static int count_interesting_parents(struct commit *commit, int max_parents)
+static int count_interesting_parents(struct commit *commit, int first_parent_only)
 {
 	struct commit_list *p;
 	int count;
 
-	for (count = 0, p = commit->parents; p && max_parents-- > 0; p = p->next) {
-		if (p->item->object.flags & UNINTERESTING)
-			continue;
-		count++;
+	for (count = 0, p = commit->parents; p; p = p->next) {
+		if (!(p->item->object.flags & UNINTERESTING))
+			count++;
+		if (first_parent_only)
+			break;
 	}
 	return count;
 }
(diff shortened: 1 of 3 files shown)


bisect.c
```diff
@@ -345,7 +348,7 @@ static struct commit_list *do_find_bisection(struct commit_list *list,
 				weight_set(p, weight(q)+1);
 				counted++;
 				show_list("bisection 2 count one",
-					  counted, nr, list);
+						  counted, nr, list, first_parent_only);
```
chriscool: It looks like there is an indent problem above.

dscho:
So this PR really is stalled, then? It is a pity, because as far as I can tell, it stalled another contribution to `git bisect` as well, claiming that the first-parent feature would be implemented quickly.

tiagonbotelho:
@chriscool unfortunately I won't be able to carry this work any further due to time constraints. Sorry I took so long to give any notice on the situation. I'll leave this open for someone to pick up. It shouldn't be far from completion I believe. I do believe this branch is outdated however, since I've wrote a couple of tests for this, which are in the mailing list.

tiagonbotelho:
The updated work can be found [here](https://github.com/tiagonbotelho/git/commit/28cc041f360b3411964f2eb2968d8b4fd1f5833f)

chriscool:
@tiagonbotelho ok, no worries, and thanks for letting us know!

retronym:
I've rebased the patch in my fork: https://github.com/retronym/git/tree/topic/bisect-first-parent as I was interested to try this feature. 

Our project has builds available for all pull request merges, and we can use `git bisect --no-checkout` to test properties of them.

The conflict is in the change to pass a single boolean (`find_all`) rather than the bitset `git_bisect_flags`. My rebase threads an extra parameter through (`first_parent`).

I wasn't able to successfully link a working binary on OSX, I think due to both Clang / GCC being on the path. I'm not planning to pursue this further right now, but thought I'd note what I'd done in case someone with more experience with this project could push it over the line.

chriscool:
@retronym thanks for rebasing the patch and the update!

dscho:
@retronym indeed, thank you!

FWIW this is the thread where @tiagonbotelho submitted the latest iteration: https://public-inbox.org/git/20180828123234.44582-1-tiagonbotelho@hotmail.com/

Essentially, the test needs to be re-done so that it does _not_ ignore all the work that was done in the `setup` test case and instead wastes many cycle to set up a separate commit graph.

I think this would be very easy, given that https://public-inbox.org/git/xmqqbm9mfcf4.fsf@gitster-ct.c.googlers.com/ is almost there.

And I still think that the commit message needs to talk a _lot_ more about this part of the diff, to equip the reader with enough context to allow them to understand why it should be correct:

```diff
	@@ -329,6 +334,11 @@ static struct commit_list *do_find_bisection(struct commit_list *list,
	 	if (0 <= weight(p))
	 		continue;
	 	for (q = p->item->parents; q; q = q->next) {
	+		if ((bisect_flags & BISECT_FIRST_PARENT)) {
	+			if (weight(q) < 0)
	+				q = NULL;
	+			break;
	+		}
	 	if (q->item->object.flags & UNINTERESTING)
	 		continue;
	 	if (0 <= weight(q))
```

It seems that there had been a couple of good hints for such an explanation, along with quite a few pointers how to improve the patch itself, in https://public-inbox.org/git/xmqqlg8qh22q.fsf@gitster-ct.c.googlers.com/

dscho:
@tiagonbotelho gentle ping?

tiagonbotelho:
Can confirm that the link posted by @dscho in [this comment](https://github.com/git/git/pull/484#issuecomment-503596412) was the latest version. It does seem that the feature is very close to completion but as I've mentioned I won't have time to work on it in the foreseeable future 😢 

I will keep myself subscribed to this thread and try to participate in the discussions though

dscho:
I just wish that this would not have held up the `--merges-only` contribution, probably to the point where Harald is not at all interested in contributing anymore, either: we left him hanging for over a year.

dscho:
@tiagonbotelho you could just close this PR, then, if you don't want to block anybody else.

PR #495, opened 2018-05-19
Git stash conversion from shell script to C.

ungps:
Git stash conversion from shell script to C.

@dscho


builtin/stash--helper.c
```diff
@@ -0,0 +1,446 @@
...
+
+	/*
+	 * Reset is overall quite simple, however there is no current public API
+	 * for resetting.
+	 */
+	argv_array_push(&args, "reset");
+	return cmd_reset(args.argc, args.argv, prefix);
+}
+
+static int get_diff_cached_index(struct strbuf *out, struct object_id *c_tree)
```
dscho: Can we please rename this to something else than `get_diff_cached_index()`? We *only* return added files here. Maybe something like `get_newly_staged()` would reflect the purpose better?

builtin/stash--helper.c
```diff
@@ -418,6 +454,79 @@ static int apply_stash(int argc, const char **argv, const char *prefix)
...
+	/*
+	 * This could easily be replaced by get_oid, but currently it will throw a
+	 * fatal error when a reflog is empty, which we can not recover from
+	 */
+	cp.git_cmd = 1;
+	/* Even though --quiet is specified, rev-parse still outputs the hash */
+	cp.no_stdout = 1;
+	argv_array_pushl(&cp.args, "rev-parse", "--verify", "--quiet", NULL);
+	argv_array_pushf(&cp.args, "%s@{0}", ref_stash);
+	ret = run_command(&cp);
```
dscho: Please use the `get_oidf()` function here instead of spawning a separate process to essentially perform the same thing.

builtin/stash--helper.c
```diff
@@ -533,6 +538,35 @@ static int drop_stash(int argc, const char **argv, const char *prefix)
...
+	argc = parse_options(argc, argv, prefix, options,
+			git_stash_helper_pop_usage, 0);
+
+	if (get_stash_info(&info, argc, argv))
+		return -1;
+
+	ret = assert_stash_ref(&info);
+	if (!ret)
+		if ((ret = do_apply_stash(prefix, &info, index)))
+			printf_ln(_("The stash entry is kept in case you need it again."));
```
dscho: Maybe better `if (!ret && do_apply_stash(...) { ret = -1; printf_ln(...) }`?

builtin/stash--helper.c
```diff
@@ -0,0 +1,780 @@
...
+		ret = do_apply_stash(prefix, &info, 1);
+	if (!ret && info.is_stash_ref)
+		ret = do_drop_stash(prefix, &info);
+
+	free_stash_info(&info);
+
+	return ret;
+}
+
+int list_stash(int argc, const char **argv, const char *prefix)
```
dscho: I think you want to mark this function (and `do_store_stash()` and `store_stash()`) as `static` (this is what makes Travis fail on macOS).

dscho:
I will continue reviewing tomorrow ;-)

builtin/stash--helper.c
ungps: I don't think that this is possible at the moment, because there will be a fatal error in case of a empty reflog.

builtin/stash--helper.c
```diff
@@ -689,6 +695,50 @@ static int show_stash(int argc, const char **argv, const char *prefix)
...
+		OPT__QUIET(&quiet, N_("be quiet, only report errors")),
+		OPT_STRING('m', "message", &stash_msg, "message", N_("stash message")),
+		OPT_END()
+	};
+
+	argc = parse_options(argc, argv, prefix, options,
+		git_stash_helper_store_usage, PARSE_OPT_KEEP_UNKNOWN);
+
+	if (argc == 0) {
+		fprintf(stderr, "\"git stash--helper\" requires one <commit> argument\n");
```
tgummerer: I see this string is currently translated, which I think it should continue to be.  Just wrap the string with `_(<string>)` to mark it for translation.  

Also this is missing `store` after `git stash--helper`.

I also wonder if we should just keep this as `git stash` instead of `git stash--helper`, as `stash--helper` feels more like an implementation detail of `git stash` that we're using during the rewrite, but will the go away I assume.

builtin/stash--helper.c
```diff
@@ -689,6 +695,50 @@ static int show_stash(int argc, const char **argv, const char *prefix)
...
+		return -1;
+	}
+
+	w_commit = argv[0];
+
+	if (!stash_msg) {
+		stash_msg  = xstrdup("Created via \"git stash--helper store\".");
+	}
```
tgummerer: omit the unnecessary braces here.

builtin/stash--helper.c
```diff
@@ -603,6 +609,31 @@ static int branch_stash(int argc, const char **argv, const char *prefix)
...
+	struct option options[] = {
+		OPT_END()
+	};
+
+	argc = parse_options(argc, argv, prefix, options,
+		git_stash_helper_list_usage, PARSE_OPT_KEEP_UNKNOWN);
+
+	if (!ref_exists(ref_stash)) {
+		fprintf(stderr, "No stash entries found.");
+	}
```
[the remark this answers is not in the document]
dscho: ... and we also saw reviewers on the Git mailing list changing their mind and saying that this is left to the taste of the developer ;-)

I would prefer without curlies, but I defer to @ungps on this.

builtin/stash--helper.c
```diff
@@ -603,6 +609,31 @@ static int branch_stash(int argc, const char **argv, const char *prefix)
...
+	struct argv_array args = ARGV_ARRAY_INIT;
+	struct option options[] = {
+		OPT_END()
+	};
+
+	argc = parse_options(argc, argv, prefix, options,
+		git_stash_helper_list_usage, PARSE_OPT_KEEP_UNKNOWN);
+
+	if (!ref_exists(ref_stash)) {
+		fprintf(stderr, "No stash entries found.");
```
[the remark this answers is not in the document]
dscho: Right, and a `\n` is missing in the message, too.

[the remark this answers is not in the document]
dscho: BTW I would strongly suggest to leave out the message. `git stash list` is scriptable, after all.

builtin/stash--helper.c
dscho: We just need to make sure that `git stash store` hands off to `stash--helper`, then the end user will never need to know that `stash--helper` even exists, I agree!

dscho:
BTW is sounds to me as if the Travis failure is a spurious problem with their build agent. The same test (`t4015-diff-whitespace.sh`) that [fails in one setup](https://travis-ci.org/git/git/jobs/381858264) succeeds in all the others, *and* that test is unrelated to stash as far as I can tell.

builtin/stash--helper.c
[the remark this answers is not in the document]
tgummerer: Oh, I must have missed that.  `Documentation/CodingGuidelines` still mentions that the style without curlies is the appropriate one, unless whatever is inside of the `if`either spans multiple lines, or there's a following else with curlies.

If we don't care about this anymore, maybe the doc should be updated?  But that is way besides the point of this PR, and probably better discussed on the list anyway, sorry for sidetracking :)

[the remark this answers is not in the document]
ungps: That's true. I am used to always put brackets, but I will be more careful in the future.

ungps:
The (branch)[https://travis-ci.org/ungps/git/builds/381858230] passes all the tests, but the PR doesn't. I tried to debug it, but I couldn't detect the problem.

dscho:
> The branch passes all the tests, but the PR doesn't. I tried to debug it, but I couldn't detect the problem.

It could be a fluke. Does the problem persist when you force-push new changes?

builtin/stash--helper.c
```diff
@@ -634,6 +640,55 @@ static int list_stash(int argc, const char **argv, const char *prefix)
...
+			argv_array_push(&args, "-p");
+	}
+
+	if (!ret) {
+		argv_array_pushl(&args, oid_to_hex(&info.b_commit),
+			oid_to_hex(&info.w_commit), NULL);
+
+		free_stash_info(&info);
+
+		ret = cmd_diff(args.argc, args.argv, prefix);
```
dscho: Just like in the commit before: `cmd_diff()` is not really intended as an API. As such, it might very well not expect to have to clean up open resources or allocated memory, or it might assume that it has to call `setup_git_directory()` itself and simply ignore the `prefix` parameter.

To clarify what that `prefix` and the `run_setup()` is all about:

Some of Git's commands expect to run in a worktree, some of them expect just the presence of a `GIT_DIR` (which can be bare), and some Git commands *might* want to have a worktree or a `GIT_DIR` but might not need any of this. For example, `git diff --no-index <file1> <file2>` works happily outside of any worktree, unattached to any Git repository.

Other commands really *require* a Git repository. They are marked with `RUN_SETUP` in the builtin list in `git.c`. What this does is simply to call `setup_git_directory()`, which historically used `cd("..")` until it found a Git repository, and then returned the relative path to the subdirectory in which it started its search.

For example, if you call `git commit .gitignore` in the subdirectory `Documentation/technical`, `setup_git_directory()` would end up with the top-level directory of the worktree as current directory, and would return `"Documentation/technical/"` as `prefix`, which `cmd_commit()` would then pass on to the code that prefixes `.gitignore` so that you really commit `Documentation/technical/.gitignore`.

Side note: I worked on this recently (I think last year), to allow for `setup_git_directory()` to *not* change the current working directory until the very end (and only if it found the Git repository). But the behavior of `setup_git_directory()` is still the same: discover the Git directory, `cd` to the top-level directory, return the relative path to the previous working directory.

Now, in `cmd_diff()`s case, I think it might actually call `setup_git_directory_gently()`, which will not `die()` if it does not find any Git repository. This is useful e.g. to let `git diff --no-index` use diff settings from the Git config, including the repository config *iff* we found any Git repository.

Back to your use case: I think what we really have to do is to call the diff machinery more directly. Maybe you can figure out from reading the code in `builtin/diff.c` what functions to call? Otherwise, I am fairly certain that there is code in `make_patch()` in `sequencer.c` that should be useful here, too.

builtin/stash--helper.c
```diff
@@ -634,6 +640,55 @@ static int list_stash(int argc, const char **argv, const char *prefix)
...
+	if (!ret && args.argc == 1) {
+		int showStat = 1, showPatch = 0;
+
+		git_config_get_bool("stash.showStat", &showStat);
+		if (showStat)
+			argv_array_push(&args, "--stat");
+
+		git_config_get_bool("stash.showPatch", &showPatch);
+		if (showPatch)
+			argv_array_push(&args, "-p");
```
dscho: This is a pretty verbatim copy of the Unix shell code. In C, however, we can do better: by calling `git_config(<fn>)` where `<fn>` is a callback function that consumes the parsed config values. Please have a look at `git grep -w git_config -- builtin/`'s output for inspiration.

builtin/stash--helper.c
```diff
@@ -634,6 +640,55 @@ static int list_stash(int argc, const char **argv, const char *prefix)
...
+	argv_array_push(&args, "diff");
+
+	for (i = 0; i < argc; ++i) {
+		if (argv[i][0] == '-')
+			argv_array_push(&args, argv[i]);
+		else
+			argv_array_push(&args_refs, argv[i]);
+	}
+
+	ret = get_stash_info(&info, args_refs.argc, args_refs.argv);
```
dscho: If there is no stash, I think we should `die()`, just like `assert_stash_like()` does. That is not, however, what `get_stash_info()` does, so we have to do it here, manually, based on the return value.

builtin/stash--helper.c
```diff
@@ -634,6 +640,55 @@ static int list_stash(int argc, const char **argv, const char *prefix)
...
+	argv_array_clear(&args_refs);
+
+	if (!ret && args.argc == 1) {
```
dscho: I am fairly certain that we should `die()` before this conditional if `ret` is non-zero, so we do not need to test for it here.

However, the way I read `test -z "$FLAGS"` in the shell script, I would assume that we have to test for `if (args.argc == 0)` here (or `if (!args.argc)`, which is kind of the convention in Git's source code).

builtin/stash--helper.c
```diff
@@ -689,6 +695,50 @@ static int show_stash(int argc, const char **argv, const char *prefix)
...
+	const char *stash_msg = NULL;
+	const char *w_commit = NULL;
+	struct option options[] = {
+		OPT__QUIET(&quiet, N_("be quiet, only report errors")),
+		OPT_STRING('m', "message", &stash_msg, "message", N_("stash message")),
+		OPT_END()
+	};
+
+	argc = parse_options(argc, argv, prefix, options,
+		git_stash_helper_store_usage, PARSE_OPT_KEEP_UNKNOWN);
```
dscho: Can you verify that this parsing sees the first option? I ask because I *think* that we need to add the `PARSE_OPT_KEEP_ARGV0` flag [here](https://github.com/ungps/git/blob/50936ea98176b64c674b1c02f87bcd875f0b3c54/builtin/stash--helper.c#L753-L754) to make subsequent `parse_options()` *not* overlook the original `argv[1]`.

builtin/stash--helper.c
```diff
@@ -689,6 +695,50 @@ static int show_stash(int argc, const char **argv, const char *prefix)
...
+	if (!ret) {
+		ret = update_ref(stash_msg, ref_stash, &obj, NULL,
+			REF_FORCE_CREATE_REFLOG, 0);
+	}
+
+	if (ret && !quiet)
+		fprintf_ln(stderr, _("Cannot update %s with %s"), ref_stash, w_commit);
+
+	return ret;
+}
```
dscho: I do not see any reason for this code to live in its own function. Why not just fold it into `store_stash()`? It'll be easier to read, too, as it does not disrupt the flow (you have to read `store_stash()` first, then scroll up to read `do_store_stash()` right now...)

dscho:
Okay, I *finally* came back to review this thing. Nice progress so far!

builtin/stash--helper.c
```diff
@@ -603,6 +609,31 @@ static int branch_stash(int argc, const char **argv, const char *prefix)
...
+	if (!ref_exists(ref_stash)) {
+		fprintf(stderr, "No stash entries found.");
+	}
+
+	argv_array_pushl(&args, "log", "--format=%gd: %gs", "-g", "--first-parent",
+		"-m", NULL);
+	argv_array_pushv(&args, argv);
+	argv_array_push(&args, ref_stash);
+	argv_array_push(&args, "--");
+	ret = cmd_log(args.argc, args.argv, prefix);
```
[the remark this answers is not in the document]
ungps: I added it to the TODO list. I also hope that it will be done by the end of the summer, but even if it is not, there is no problem because I will continue to work on this.

builtin/stash--helper.c
```diff
@@ -634,6 +640,55 @@ static int list_stash(int argc, const char **argv, const char *prefix)
...
+	int i, ret = 0;
+	struct argv_array args = ARGV_ARRAY_INIT;
+	struct argv_array args_refs = ARGV_ARRAY_INIT;
+	struct stash_info info;
+	struct option options[] = {
+		OPT_END()
+	};
+
+	argc = parse_options(argc, argv, prefix, options,
+			git_stash_helper_show_usage, PARSE_OPT_KEEP_UNKNOWN);
```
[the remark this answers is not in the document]
ungps: Actually, I think it is possible not to call `parse_options()`. Its use at the moment is to ignore the first argument. However, I think that `parse_options()` will be necessary when replacing `cmd_diff()` with something more direct.

builtin/stash--helper.c
ungps: Because of `argv_array_push(&args, "diff");` on line 661 `args.argc` will be 1, not 0.

builtin/stash--helper.c
ungps: I think that a better solution for this is to edit `assert_stash_like()` into a void function to be more similar with `assert` from `<assert.h>`. Since `struct stash_info *info` is given as parameter, it can be freed inside the `assert_stash_*()` functions.

builtin/stash--helper.c
ungps: Right.

ungps:
@dscho  thank you a lot for the review. I will submit fix-up patches soon enough.

builtin/stash--helper.c
ungps: I guess that this would be the best way to proceed. In the future, there might be added other config options.

builtin/stash--helper.c
ungps: Theoretically, after the first `parse_options()` inside `cmd_stash__helper()`, the first argument (argv[0]) will be the given subcommand (i.e `store`). After the second `parse_options()` inside `store_stash()`, the first argument (argv[0]) will be the first argument after `store` (i.e for `git stash store foobar`, after the second parse, argv[0] will be `foobar`).

Now, I tested `git stash store foobar` with two printf debug messages before and after `parse_options(...)` call in `store_stash()`. Before it, argv[0] is `store`, and after it is `foobar`.

I guess it is ok. Isn't it?

dscho:
Awesome!

builtin/stash--helper.c
ungps: Thanks for your explanation, @dscho!

In this case, I think there shouldn't be any worries regarding `setup_git_directory()` since using `git stash ...` requires a worktree. (`{ "stash--helper", cmd_stash__helper, RUN_SETUP | NEED_WORK_TREE },`).

The last [commit](https://github.com/git/git/pull/495/commits/0f37f0a6da494a6e1379232d5239eba04d003dc8) I submitted replaces `cmd_diff()` with something more direct.

builtin/stash--helper.c
ungps: I done this while refactoring `show_stash()` to use the diff API, as you suggested [here](https://github.com/git/git/pull/495#discussion_r191661468).

builtin/stash--helper.c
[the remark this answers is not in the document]
dscho: Maybe we should introduce the `parse_options()` call when we need it, i.e. later...

builtin/stash--helper.c
dscho: You might indeed get lucky here in that things work as intended. It is still better to imitate `make_patch()`.

ungps:
Of course! I pushed the version with squashed fixup commits.

builtin/stash--helper.c
```diff
@@ -625,6 +631,69 @@ static int list_stash(int argc, const char **argv, const char *prefix)
...
+			argv_array_push(&args_refs, argv[i]);
+	}
+
+	if (get_stash_info(&info, args_refs.argc, args_refs.argv)) {
+		argv_array_clear(&args);
+		argv_array_clear(&args_refs);
+		return -1;
+	}
+
+	if (args.argc == 1) {
```
dscho: Okay, this really would benefit a comment like `/* no flags so far for the `diff` command */` because I stumbled over it again. This `args.argc == 1` condition is just a bit too magic to be self-explanatory.

builtin/stash--helper.c
```diff
@@ -625,6 +631,69 @@ static int list_stash(int argc, const char **argv, const char *prefix)
...
+		git_config(git_stash_config, NULL);
+		if (show_stat)
+			argv_array_push(&args, "--stat");
+
+		if (show_patch)
+			argv_array_push(&args, "-p");
+	}
+
+	argv_array_pushl(&args, oid_to_hex(&info.b_commit),
+		oid_to_hex(&info.w_commit), NULL);
```
dscho: It might be a bit easier to read if the `oid_to_hex()` calls were aligned, ie. if you broke the line after `&args,`. But that's just my sense of taste, yours may be different, and you're the author, so I defer the judgement to you.

builtin/stash--helper.c
```diff
@@ -625,6 +631,69 @@ static int list_stash(int argc, const char **argv, const char *prefix)
...
+
+	free_stash_info(&info);
```
dscho: Maybe it would make more sense to put this further down, where `&args` is cleared, too. It logically belongs to the clean-up part of this function, even if you could save a tiny bit of memory by releasing the memory earlier.

builtin/stash--helper.c
```diff
@@ -18,6 +18,7 @@ static const char * const git_stash_helper_usage[] = {
 	N_("git stash--helper ( pop | apply ) [--index] [-q|--quiet] [<stash>]"),
 	N_("git stash--helper branch <branchname> [<stash>]"),
 	N_("git stash--helper clear"),
+	N_("git stash--helper store [-m|--message <message>] [-q|--quiet] <commit>"),
```
dscho: Please mention in the commit message that the usage was forgotten in the shell script version (which I think it was, but it *might* be missing on purpose... but then, I can't find it in the commit log), so this is more than a literal conversion, but in a good sense.

builtin/stash--helper.c
```diff
@@ -650,53 +650,67 @@ static int git_stash_config(const char *var, const char *value, void *cb)
...
+	if (argc > 1)
+		usage(git_stash_helper_show_usage);
 
-	argv_array_clear(&args);
+	/* Do the diff thing. */
+	diff_tree_oid(&info.b_commit, &info.w_commit, "", &rev.diffopt);
+	log_tree_diff_flush(&rev);
+
+	free_stash_info(&info);
 	return ret;
```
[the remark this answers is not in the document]
dscho: I guess this no longer applies, and GitHub shows this here because I hit "finish review" after the old commit went away. Please ignore.

builtin/stash--helper.c
```diff
@@ -694,6 +700,43 @@ static int show_stash(int argc, const char **argv, const char *prefix)
...
+	struct option options[] = {
+		OPT__QUIET(&quiet, N_("be quiet, only report errors")),
+		OPT_STRING('m', "message", &stash_msg, "message", N_("stash message")),
+		OPT_END()
+	};
+
+	argc = parse_options(argc, argv, prefix, options,
+		git_stash_helper_store_usage, PARSE_OPT_KEEP_UNKNOWN);
+
+	if (argc == 0) {
```
dscho: Should this not be `if (argc != 1)`? I mean, we are complaining that there is not exactly one `<commit>` argument, so we should not allow *two* (or more)...

dscho: FWIW the shell version agrees:

```sh
test $# = 1 ||
die "$(eval_gettext "\"$dashless store\" requires one <commit> argument")"
```

builtin/stash--helper.c
```diff
@@ -694,6 +700,43 @@ static int show_stash(int argc, const char **argv, const char *prefix)
...
+
+	w_commit = argv[0];
+
+	if (!stash_msg)
+		stash_msg  = xstrdup("Created via \"git stash--helper store\".");
+
+	ret = get_oid(w_commit, &obj);
+	if (!ret) {
+		ret = update_ref(stash_msg, ref_stash, &obj, NULL,
+			REF_FORCE_CREATE_REFLOG, 0);
```
dscho: I wonder whether we really want to pass `0` here instead of `UPDATE_REFS_MSG_ON_ERR`. I think we should pass the latter. Or maybe even better `quiet ? UPDATE_REFS_QUIET_ON_ERR : UPDATE_REFS_MSG_ON_ERR`.

builtin/stash--helper.c
```diff
@@ -10,6 +10,8 @@
 #include "run-command.h"
```
dscho: Personally, I would move this commit and its parent very close to the commit that converts the `show` subcommand. "Close" as in: directly afterwards. I think that would make for a better story to the reader.

builtin/stash--helper.c
```diff
@@ -650,53 +650,67 @@ static int git_stash_config(const char *var, const char *value, void *cb)
 		show_patch = git_config_bool(var, value);
 		return 0;
 	}
+	return git_default_config(var, value, cb);
```
dscho: Whoa. This hunk needs to be moved into the commit introducing `git_stash_config()`, as that commit would not even compile without a final `return`.

builtin/stash--helper.c
```diff
@@ -650,53 +650,67 @@ static int git_stash_config(const char *var, const char *value, void *cb)
...
-		argv_array_clear(&args_refs);
-		return -1;
+			flags++;
 	}
 
-	if (args.argc == 1) {
+	/*
+	 * The config settings are applied only if there are not passed
+	 * any flags.
+	 */
```
dscho: This comment would have helped me understand quicker what is going on if it had been introduced in the commit that converted the `show` subcommand.

dscho: (Read: please move this diff hunk there.)

builtin/stash--helper.c
```diff
@@ -650,53 +650,67 @@ static int git_stash_config(const char *var, const char *value, void *cb)
...
+	if (!flags) {
 		git_config(git_stash_config, NULL);
 		if (show_stat)
-			argv_array_push(&args, "--stat");
-
-		if (show_patch)
-			argv_array_push(&args, "-p");
+			rev.diffopt.output_format |= DIFF_FORMAT_DIFFSTAT;
+		if (show_patch) {
+			rev.diffopt.output_format = ~DIFF_FORMAT_NO_OUTPUT;
```
dscho: Do we need this `~NO_OUTPUT` line also in the `show_stat` case?

builtin/stash--helper.c
```diff
@@ -650,53 +650,67 @@ static int git_stash_config(const char *var, const char *value, void *cb)
...
+	if (!rev.diffopt.output_format) {
+		rev.diffopt.output_format = DIFF_FORMAT_PATCH;
+	}
+	diff_setup_done(&rev.diffopt);
+	rev.diffopt.flags.recursive = 1;
+	setup_diff_pager(&rev.diffopt);
 
-	ret = cmd_diff(args.argc, args.argv, prefix);
+	/*
+	 * We can early return if there was any option not recognised by
```
dscho: "We can return early ..."

And we still have to release the memory (`stash_args`) if we return early.

dscho:
> Excellent progress.

I just checked, you only need to convert `save_stash`, `push_stash` and `create_stash` to make this complete... Nice!

Granted, `push_stash` is the big one. Have you started tackling it yet?

t/t3907-stash-show-config.sh
```diff
@@ -0,0 +1,94 @@
+#!/bin/sh
+
+test_description='Test git stash show configuration.
+
```
dscho: I would leave it at a one-line description, not add two empty lines ;-)

t/t3907-stash-show-config.sh
```diff
@@ -0,0 +1,94 @@
...
+'
+
+. ./test-lib.sh
+
+config_stash_stat_and_patch () {
+	if test "$1" != "unset"
+	then
+		git config stash.showStat "$1"
```
dscho: If you use `test_config`, then the config setting will be automatically reverted after the test case. Not sure whether you want to do that here, but I could imagine that you do....

t/t3907-stash-show-config.sh
```diff
@@ -0,0 +1,94 @@
...
+	if git config --bool stash.showPatch >/dev/null
+	then
+		git config --unset stash.showPatch
+	fi
+}
+
+get_expected_output () {
+	FLAGS=
+
+	if test "$(git config --bool stash.showStat || echo true)" = "true"
```
dscho: I *think* we have some new option for `git config` that defines a default value...

*clicketyclick*

Yep: https://git-scm.com/docs/git-config#git-config---defaultltvaluegt

So: `test true = "$(git config --bool --default true stash.showStat)"` will get you where you want to be.

t/t3907-stash-show-config.sh
```diff
@@ -0,0 +1,94 @@
...
+	then
+		FLAGS=--stat
+	fi
+
+	if test "$(git config --bool stash.showPatch || echo false)" = "true"
+	then
+		FLAGS=${FLAGS}${FLAGS:+ }-p
```
dscho: Shorter: `FLAGS="${FLAGS:+$FLAGS }-p"`

t/t3907-stash-show-config.sh
```diff
@@ -0,0 +1,94 @@
...
+		fi
+
+		unset_config
+		test_expect_success "showStat $i and showPatch $j" '
+			config_stash_stat_and_patch $i $j &&
+			echo 2 >file &&
+			check_output
+		'
+	done
+done
```
dscho: It is tempting to do it in a loop, as it is repetitive code.

However, I would suggest to refrain from doing that as it makes debugging regressions a pain in the back. I'd rather define a couple nice shell functions, and then spell out each invocation individually.

t/t3907-stash-show-config.sh
```diff
@@ -0,0 +1,94 @@
...
+for i in $options
+do
+	for j in $options
+	do
+		if test "$i" = "false" && (test "$j" = "false" || test "$j" = "unset")
+		then
+			test_expect_success "showStat $i and showPatch $j" '
+				config_stash_stat_and_patch $i $j &&
+				echo 2 >file &&
+				get_actual_output &&
```
dscho: How about folding that `echo 2 >file` into `get_actual_output`?

t/t3907-stash-show-config.sh
```diff
@@ -0,0 +1,94 @@
...
+get_actual_output () {
+	git stash &&
+	git stash show >actual
+}
+
+check_output () {
+	get_expected_output &&
+	get_actual_output &&
+	test_cmp expect actual &&
+	rm expect actual
```
dscho: This is not strictly necessary, I think. For debugging, that line will have to be commented out anyway, I have a hunch.

t/t3907-stash-show-config.sh
```diff
@@ -0,0 +1,94 @@
...
+}
+
+options="unset true false"
+
+test_expect_success 'setup' '
+	echo 1 >file &&
+	git add file &&
+	git commit -m initial
```
dscho: Shorter: `test_commit file` (the file name will be `file.t` in that case, though).

t/t3907-stash-show-config.sh
dscho: Okay, so I think I would prefer something like this instead:

```sh
#!/bin/sh

test_description='Test git stash show configuration.'

. ./test-lib.sh

test_expect_success 'setup' '
	test_commit file
'

# takes three parameters:
# 1. the stash.showStat value (or "<unset>")
# 2. the stash.showPatch value (or "<unset>")
# 3. the diff options of the expected output (or "<empty>" for no output)
test_stat_and_patch () {
	if test "<unset>" = "$1"
	then
		git config --unset stash.showStat
	else
		test_config stash.showStat "$1"
	fi &&
	if test "<unset>" = "$2"
	then
		git config --unset stash.showPatch
	else
		test_config stash.showPatch "$2"
	fi &&

	echo 2 >file.t &&
	git diff "$@" >expect &&
	git stash &&
	git stash show >actual &&

	shift &&
	shift &&
	if test "<empty>" = "$1"
	then
		test_must_be_empty actual
	else
		test_cmp expect actual
	fi
}

test_expect_success 'various combinations' '
	test_stat_and_patch "<unset>" "<unset>" &&
	test_stat_and_patch "<unset>" false &&
	test_stat_and_patch "<unset>" true -p &&
	test_stat_and_patch false "<unset>" "<empty>" &&
	test_stat_and_patch false false "<empty>" &&
	test_stat_and_patch false true -p &&
	test_stat_and_patch true "<unset>" --stat &&
	test_stat_and_patch true false --stat &&
	test_stat_and_patch true true --stat -p
'

test_done
```

I could even imagine that that `<empty>` handling is totally unnecessary...

What do you think?

ungps: Thanks, @dscho! Indeed, your approach makes debugging easier. To make it even easier to debug, can't we do a `test_stat_and_patch` per `test_expect_success`?. 

The first two `if`s inside `test_patch_success` should be "unchained" since `git config --unset foobar` returns non-zero value if `foobar` is not set in the first place. I guess that, in this case, a break in `&&`-chain wouldn't be a problem, right? I mean, if `git config` doesn't work as intended, then it would be reported by `t1300-config.sh`

I made some more little changes to it (call one `test_stat_and_patch` per `test_expect_sucesss`, moved `shift && shift` before `git diff $@` and dropped `<empty>` handling since it is not necessary)  and force-pushed it. Do you think it is good enough now?

tgummerer: > The first two ifs inside test_patch_success should be "unchained" 

Does that make it past chainlint?  I guess it might because this is being done in a function?  You could also consider using [`test_might_fail`](https://github.com/git/git/blob/master/t/README#L717), which is happy with both successes and failures, but will catch segfaults.

mtrout:
New email mtrout1111@gmail.com

builtin/stash.c
```diff
@@ -0,0 +1,1601 @@
...
+}
+
+static int save_stash(int argc, const char **argv, const char *prefix)
+{
+	int keep_index = -1;
+	int patch_mode = 0;
+	int include_untracked = 0;
+	int quiet = 0;
+	int ret = 0;
+	char *stash_msg = NULL;
```
dscho: How about making this a `const char *` (to avoid casting later)?

builtin/stash.c
```diff
@@ -0,0 +1,1601 @@
...
+	git_config(git_diff_basic_config, NULL);
+
+	argc = parse_options(argc, argv, prefix, options, git_stash_usage,
+			     PARSE_OPT_KEEP_UNKNOWN | PARSE_OPT_KEEP_DASHDASH);
+
+	index_file = get_index_file();
+	strbuf_addf(&stash_index_path, "%s.stash.%" PRIuMAX, index_file,
+		    (uintmax_t)pid);
+
+	if (argc == 0)
```
dscho: I think in Git's source code, we prefer `(!argc)` in such cases, but if you feel strongly, it is okay to leave `(argc == 0)`.

builtin/stash.c
```diff
@@ -0,0 +1,1601 @@
...
+			    N_("include ignore files"), 2),
+		OPT_STRING('m', "message", &stash_msg, N_("message"),
+			   N_("stash message")),
+		OPT_END()
+	};
+
+	if (argc)
+		argc = parse_options(argc, argv, prefix, options,
+				     git_stash_push_usage,
+				     0);
```
dscho: Oh, I see. You pass `0, NULL` from one callsite, so my comment did actually not apply, sorry!

dscho:
Oh, another thing before sending off v10: our [coccinelle scripts identified code](https://travis-ci.org/git/git/jobs/453712231) that should be expressed differently:

```diff
diff -u -p a/builtin/stash.c b/builtin/stash.c
--- a/builtin/stash.c
+++ b/builtin/stash.c
@@ -406,8 +406,8 @@ static int do_apply_stash(const char *pr
 		return error(_("cannot apply a stash in the middle of a merge"));
 
 	if (index) {
-		if (!oidcmp(&info->b_tree, &info->i_tree) ||
-		    !oidcmp(&c_tree, &info->i_tree)) {
+		if (oideq(&info->b_tree, &info->i_tree) ||
+		    oideq(&c_tree, &info->i_tree)) {
 			has_index = 0;
 		} else {
 			struct strbuf out = STRBUF_INIT;
@@ -441,7 +441,7 @@ static int do_apply_stash(const char *pr
 	o.branch1 = "Updated upstream";
 	o.branch2 = "Stashed changes";
 
-	if (!oidcmp(&info->b_tree, &c_tree))
+	if (oideq(&info->b_tree, &c_tree))
 		o.branch1 = "Version stash was based on";
 
 	if (quiet)
```

Could you fold these changes in, please?

builtin/stash.c
```diff
@@ -0,0 +1,1602 @@
...
+
+	if (!stash_msg_buf->len) {
+		strbuf_release(stash_msg_buf);
+		strbuf_addf(stash_msg_buf, "WIP on %s", msg.buf);
+	} else {
+		struct strbuf temp_buf = STRBUF_INIT;
+		strbuf_addf(&temp_buf, "On %s: %s", branch_name,
+			    stash_msg_buf->buf);
+		strbuf_release(stash_msg_buf);
+		*stash_msg_buf = temp_buf;
```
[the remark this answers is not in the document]
dscho: Come to think of it, a much, much easier implementation of `strbuf_insertf()` would be this:

```c
void strbuf_vinsertf(struct strbuf *sb, size_t pos, const char *fmt, va_list ap)
{
	int len, len2;
	char save;
	va_list cp;

	if (pos > sb->len)
		die("`pos' is too far after the end of the buffer");
	va_copy(cp, ap);
	len = vsnprintf(sb->buf + sb->len, 0, fmt, cp);
	va_end(cp);
	if (len < 0)
		BUG("your vsnprintf is broken (returned %d)", len);
	if (!len)
		return; /* nothing to do */
	if (unsigned_add_overflows(sb->len, len))
		die("you want to use way too much memory");
	strbuf_grow(sb, len);
	memmove(sb->buf + pos + len, sb->buf + pos, sb->len - pos);
	/* vsnprintf() will append a NUL, overwriting one of our characters */
	save = sb->buf[pos + len];
	len2 = vsnprintf(sb->buf + pos, sb->alloc - sb->len, fmt, ap);
	sb->buf[pos + len] = save;
	if (len2 != len)
		BUG("your vsnprintf is broken (returns inconsistent lengths)");
	strbuf_setlen(sb, sb->len + len);
}

void strbuf_insertf(struct strbuf *sb, size_t pos, const char *fmt, ...)
{
	va_list ap;
	va_start(ap, fmt);
	strbuf_vinsertf(sb, pos, fmt, ap);
	va_end(ap);
}
```

In other words, rather than appending the formatted text and then rotating it into place, it makes more sense to make room at `pos` first, and then write into that location directly.

The only problem this approach could have is when `vsnprintf()` writes too many bytes, but that is a bug in `vsnprintf()`, then, and for such buggy implementations, Git already has its `compat/` replacement.

dscho:
This made it into `master` via e36adf712251.

PR #497, opened 2018-05-23
rebase: minimal skeleton builtin

prertik:
To rewrite git-rebase.sh into a C builtin, a minimal implement of skeleton or
framework which holds the basic usage and creates a base for further
modification. This allows us to keep on depending upon the git-rebase.sh until
builtin/rebase2.c is ready.

This skeletal will be ultimately be re-written to further accommodate the
crucial rebase features and with the development of this builtin the work of
rebase--helper gets obsolete with every commit.

Signed-off-by: Pratik Karki <predatoramigo@gmail.com>


dscho:
It does not really make sense to call it `rebase--helper2`: first of all, the `*--helper` commands are *not* end-user facing, while `rebase2` is. Second, they are all called `*--helper`, and *none* of them is `*--helper2`.

After sleeping over this, however, I am even convinced that we have to not only go the non--helper route, but actually imitate the way I did it with the difftool in be8a90e59ce4c7603207a8255284fdbbffff1a2e:

- rename `git-rebase.sh` to `git-legacy-rebase.sh`,
- adjust all the files referencing `git-rebase`, such as `.gitignore`, `Makefile` etc,
- add the minimal draft of the `rebase` builtin that for now calls `use_builtin_rebase(void)` to determine whether to hand off to `git legacy-rebase` or not.

This has the rather dramatic advantage that you can switch the default from legacy to non-legacy and *very* quickly encounter bugs by running the test suite.

prertik:
This is my second calling to @stefanbeller to give his opinion on all this. @chriscool  and @dscho, can we come to a generally accepted technique so I can progress further? I'm getting really really confused. Both of your techniques are good for me.

dscho:
I am sorry for the confusion!

Re: how to proceed: I have a rather strong opinion about going the same route as `difftool` now, because you basically face the same problems as I did when I converted that command: you are converting from top-down, as your main project is to convert the option parsing. And it would not make sense at all to move that to a helper, as you would have to call that helper from the script (helpers are *only* called from scripts, they are implementation details that the user never should need to care about), transferring all relevant information to the helper, then somehow parsing the output of the helper back into shell data structures. And all of that would be done completely in vain because at the end of the project, this code would become obsolete again.

Besides, changing the --helper would necessarily cause tons of conflicts with the fellow GSoC project by @agrn. We should very much avoid that.

So I am rather strongly suggesting that @chriscool should agree with me that the --helper approach is not appropriate here.

Instead, let's imitate https://github.com/git/git/commit/be8a90e59ce4c7603207a8255284fdbbffff1a2e here: rename `git-rebase.sh` to `git-legacy-rebase.sh`, and introduce a builtin `builtin/rebase.c` that inspects a config option (opt-in) to hand off to the legacy rebase. When that option is turned off explicitly, we continue with the builtin rebase.

Then we can build up the interactive rebase incrementally: the first step could even be to implement essentially these shell commands: `. git-rebase--am && git_rebase_am`. Then, we would implement enough of the command-line parsing to support `git rebase <base>`. Then, we should pass in enough of the environment variables to the shell script snippet to make this work.

In general, that would be wrong, of course, but would it not be satisfying to see the first draft of the builtin rebase to perform its duties correctly?

And from there, we can add all of the functionality, one by one, until it all works. And we can even have a running statistic how many of the test cases in Git's test suite still fail with the builtin rebase, to gage how far along we are.

chriscool:
> So I am rather strongly suggesting that @chriscool should agree with me that the --helper approach is not appropriate here.

Ok I agree with you now.

sea51930:
Don't pass on it again. It took two months and nothing happened. Life was therefore a failure. FUCK was wasted on the e-mail. There wasn’t any substantial meaning.

2018-05-25 20:18 GMT+08:00 Christian Couder <notifications@github.com<mailto:notifications@github.com>>:

So I am rather strongly suggesting that @chriscool<http://link.gmreg5.net/x/c?c=2414898&l=61cd499e-6d7e-45de-8fb3-1c013b463369&r=616d6606-e2af-4c27-9d22-2c3719041a47> should agree with me that the --helper approach is not appropriate here.

Ok I agree with you now.

prertik:
@dscho & @chriscool, glad we came to an understanding. I'm very happy now and I believe the way @dscho guided is the best way. Adding @stefanbeller to fill up his notifications on this pull request. 😉 
[This](https://youtu.be/VZxr5e9W9eg?t=16) will perfectly demonstrate how glad I am.

prertik:
@sea51930 I don't understand... 😕

stefanbeller:
@prertik 
Sorry for chiming in late here, I was AFK for a while and then had so many loose ends to look at. I should have announced that prior to that.

@dscho wrote:
> Instead, let's imitate be8a90e here: rename git-rebase.sh to git-legacy-rebase.sh, and introduce a builtin builtin/rebase.c that inspects a config option (opt-in) to hand off to the legacy rebase. When that option is turned off explicitly, we continue with the builtin rebase.

I think that is a good way to go, as it has proven to be successful. Another (although slightly different) example for the "skeleton + add features on top" is found in the series that was merged in 7aa2da616208a51ec51ce3639fe6abd9631b74db, so the skeleton starts at 73c2779f421fe1eaead6f1c07a0e134a9c17d6db, and then eventually 783d7e865ec8f6190f8d3abe3ab72a9410d611f1 removed the shell code. (That series was also a GSoC student)

dscho:
@stefanbeller Yep, that's essentially what I imitated in `difftool`. I chose to go for a config setting instead of an environment variable, though, as it made it easier for me to let Git for Windows' users opt-in to use the builtin difftool early (and report bugs and make it robust).

builtin/rebase.c
```diff
@@ -0,0 +1,54 @@
...
+		if (sane_execvp(path, (char **)argv) < 0)
+			die_errno("could not exec %s", path);
+
+		return 0;
+	}
+	prefix = setup_git_directory();
+	trace_repo_setup(prefix);
+	setup_work_tree();
+
+	die("NEEDS WORK");
```
dscho: That's okay, but I think it would be better to use either `"TODO"` or `"TBD"` (for "To Be Done") here.

Now, let's plan for the next steps because we are really close to the exciting part! I am fairly certain that you can *already* implement enough code to run `git rebase <base>` in a relatively small second commit: As I mentioned a couple of times via mail and IRC (i.e. too far from the code to be useful), what we need to do here is essentially emulate https://github.com/prertik/git/blob/41531d0ba88cfe20e870d57ee437d0fcd7d05064/git-legacy-rebase.sh#L203-L222 which kind of boils down to

```sh
. git-rebase--$type
git_rebase__$type${preserve_merges:+__preserve_merges}
```

This is very simple shell scripting that we can call ourselves *very easily* by imitating what `rebase -i`'s `exec` command does: https://github.com/prertik/git/blob/41531d0ba88cfe20e870d57ee437d0fcd7d05064/sequencer.c#L2501-L2509

Note: I chose that `exec` example because it not only demonstrates how to execute a Unix shell script snippet, but also how to pass environment variables to it (which you will need to do later, quite a few variables even, because the `git-rebase--<type>` backends have to consume pretty much all of the parsed command-line options in the form of shell variables.

Please do take the time, though, to implement this environment variable passing using `argv_array_push()` and `argv_array_pushf()`, as demonstrated e.g. here: https://github.com/prertik/git/blob/41531d0ba88cfe20e870d57ee437d0fcd7d05064/column.c#L377-L394

By the way, it's not quite the truth that we can get away with implementing code to call *just* those two Unix shell commands: there is a function `finish_rebase()` which also needs to be implemented. However, in the first run, it does not need to handle the autostash. And the `rm -f` command corresponds to the POSIX function `unlink()`, the `rm -rf` to `remove_dir_recursively()`, and the most concise example how to run `git gc --auto` is [in `builtin/merge.c`](https://github.com/prertik/git/blob/41531d0ba88cfe20e870d57ee437d0fcd7d05064/builtin/merge.c#L408-L416).

To support `git -c rebase.useBuiltin=true rebase <base>`, I think that a very simple "option parsing" would be enough: `if (argc != 2) die("Usage: %s <base>", argv[0]);`.

The autostash part can be handled in a later patch, and should definitely leverage the [`apply_autostash()`](https://github.com/prertik/git/blob/41531d0ba88cfe20e870d57ee437d0fcd7d05064/sequencer.c#L2913-L2955) function that was introduced to support finishing interactive rebases in the sequencer; to make that work with the `type=am` mode, you will want to use [`get_dir()`](https://github.com/prertik/git/blob/41531d0ba88cfe20e870d57ee437d0fcd7d05064/sequencer.c#L197-L202) and teach it about the `rebase-apply` directory, and you will want to make `apply_autostash()` non-static (but keep `get_dir()` static).

I really think that we are very close to having something that is both exciting ("It can run a (super-simple) rebase already!") as well as helpful, as it leads the way to using the `t34*.sh` suite of tests to assess how far along this project is. You could, for example, append the line `git config rebase.useBuiltin true` to the body of the `test_create_repo` function in `t/test-lib-functions.sh`, and then run `make -C t t3400-rebase.sh` to see whether any test case succeeds ;-)

builtin/rebase.c
```diff
@@ -0,0 +1,54 @@
...
+*/
+static int use_builtin_rebase(void) {
+	struct child_process cp = CHILD_PROCESS_INIT;
+	struct strbuf out = STRBUF_INIT;
+	int ret;
+
+	argv_array_pushl(&cp.args,
+			 "config", "--bool", "rebase.usebuiltin", NULL);
+	cp.git_cmd = 1;
+	if (capture_command(&cp, &out, 28))
```
chriscool: I wonder if `out`should be released before returning in case capture_command() failed.

Also I wonder why 28 is used.

prertik: Oh no.... At first I was trying to options provided by rebase into the builtin. There were about 28 options provided in rebase. But, that was not the approach to do it and I failed to do so. I think I mistakenly left 28 there.

sea51930:
#497

PR #503, opened 2018-06-06
 rebase--interactive: rewrite the edit-todo functionality to C

agrn:
This rewrites the edit-todo functionality from shell to C. This is part of the effort to rewrite interactive rebase in C.

The relevant commit is the last one.

### DIFF
--- a/builtin/rebase--helper.c
+++ b/builtin/rebase--helper.c
@@ -13,12 +13,12 @@ static const char * const builtin_rebase_helper_usage[] = {
 int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 {
 	struct replay_opts opts = REPLAY_OPTS_INIT;
-	unsigned flags = 0, keep_empty = 0, rebase_merges = 0, write_edit_todo = 0;
+	unsigned flags = 0, keep_empty = 0, rebase_merges = 0;
 	int abbreviate_commands = 0, rebase_cousins = -1;
 	enum {
 		CONTINUE = 1, ABORT, MAKE_SCRIPT, SHORTEN_OIDS, EXPAND_OIDS,
 		CHECK_TODO_LIST, SKIP_UNNECESSARY_PICKS, REARRANGE_SQUASH,
-		ADD_EXEC, APPEND_TODO_HELP
+		ADD_EXEC, APPEND_TODO_HELP, EDIT_TODO
 	} command = 0;
 	struct option options[] = {
 		OPT_BOOL(0, "ff", &opts.allow_ff, N_("allow fast-forward")),
@@ -28,8 +28,6 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 		OPT_BOOL(0, "rebase-merges", &rebase_merges, N_("rebase merge commits")),
 		OPT_BOOL(0, "rebase-cousins", &rebase_cousins,
 			 N_("keep original branch points of cousins")),
-		OPT_BOOL(0, "write-edit-todo", &write_edit_todo,
-			 N_("append the edit-todo message to the todo-list")),
 		OPT_CMDMODE(0, "continue", &command, N_("continue rebase"),
 				CONTINUE),
 		OPT_CMDMODE(0, "abort", &command, N_("abort rebase"),
@@ -50,6 +48,9 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 			N_("insert exec commands in todo list"), ADD_EXEC),
 		OPT_CMDMODE(0, "append-todo-help", &command,
 			    N_("insert the help in the todo list"), APPEND_TODO_HELP),
+		OPT_CMDMODE(0, "edit-todo", &command,
+			    N_("edit the todo list during an interactive rebase"),
+			    EDIT_TODO),
 		OPT_END()
 	};
 
@@ -90,6 +91,8 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 	if (command == ADD_EXEC && argc == 2)
 		return !!sequencer_add_exec_commands(argv[1]);
 	if (command == APPEND_TODO_HELP && argc == 1)
-		return !!append_todo_help(write_edit_todo, keep_empty);
+		return !!append_todo_help(0, keep_empty);
+	if (command == EDIT_TODO && argc == 1)
+		return !!edit_todo_list(flags);
 	usage_with_options(builtin_rebase_helper_usage, options);
 }
--- a/builtin/rebase--helper.c
+++ b/builtin/rebase--helper.c
@@ -3,6 +3,7 @@
 #include "config.h"
 #include "parse-options.h"
 #include "sequencer.h"
+#include "rebase-interactive.h"
 
 static const char * const builtin_rebase_helper_usage[] = {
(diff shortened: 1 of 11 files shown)


builtin/rebase--helper.c
```diff
@@ -49,6 +47,9 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 			N_("insert exec commands in todo list"), ADD_EXEC),
 		OPT_CMDMODE(0, "append-todo-help", &command,
 			    N_("insert the help in the todo list"), APPEND_TODO_HELP),
+		OPT_CMDMODE(0, "edit-todo", &command,
```
chriscool: In the commit message you say "a new flag, `edit-todo`, is added" but it looks like the "flag" is a command mode so "a new command mode, `--edit-todo`, is added" might be better.

builtin/rebase--helper.c
```diff
@@ -89,6 +90,8 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 	if (command == ADD_EXEC && argc == 2)
 		return !!sequencer_add_exec_commands(argv[1]);
 	if (command == APPEND_TODO_HELP && argc == 1)
-		return !!append_todo_help(edit_todo, keep_empty);
+		return !!append_todo_help(0, keep_empty);
```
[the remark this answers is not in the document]
agrn: Yes, it’s used in edit_todo_list(): https://github.com/git/git/pull/503/commits/9f0514cb045e8569d2bc770e94e174cf25588224#diff-79231f0693f84f3951daeea17065aad9R4408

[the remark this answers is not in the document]
chriscool: I tend to think that a separate commit would be better.

builtin/rebase--helper.c
```diff
@@ -12,12 +12,12 @@ static const char * const builtin_rebase_helper_usage[] = {
 int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 {
 	struct replay_opts opts = REPLAY_OPTS_INIT;
-	unsigned flags = 0, keep_empty = 0, rebase_merges = 0, edit_todo = 0;
```
chriscool: In the commit message you say "`edit-todo`, is added, and the `write-edit-todo` is removed" which is confusing when we see the above line.

It might have been better to use `write_edit_todo` instead of `edit_todo` in the first place.

sequencer.c
```diff
@@ -4386,6 +4386,34 @@ int append_todo_help(unsigned edit_todo, unsigned keep_empty)
...
+	const char *todo_file = rebase_path_todo();
+	FILE *todo;
+
+	if (strbuf_read_file(&buf, todo_file, 0) < 0)
+		return error(_("could not read '%s'."), todo_file);
+
+	strbuf_stripspace(&buf, 1);
+	todo = fopen_or_warn(todo_file, "w");
+	if (!todo)
+		return 1;
```
chriscool: Maybe call `strbuf_release(&buf)` in case of error.

builtin/rebase--helper.c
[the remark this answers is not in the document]
chriscool: Actually it looks like we still need the first argument of append_todo_help() below.

sequencer.c
```diff
@@ -4386,6 +4386,34 @@ int append_todo_help(unsigned edit_todo, unsigned keep_empty)
...
+
+	strbuf_write(&buf, todo);
+	fclose(todo);
+	strbuf_release(&buf);
+
+	transform_todos(flags | TODO_LIST_SHORTEN_IDS);
+	append_todo_help(1, 0);
+
+	launch_editor(todo_file, NULL, NULL);
```
chriscool: What happens if launching the editor fails? Does launch_editor() die or return an error? Which will be the exit code of git-rebase--interactive.sh?

builtin/rebase--helper.c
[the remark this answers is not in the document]
chriscool: Sorry I did not see your comment above when writing the 2 comments above.

sequencer.c
agrn: `launch_editor()` returns an error, but it won’t be handled, so the return code will be 0.

chriscool: The original shell script was:

```
		git_sequence_editor "$todo" ||
			die "$(gettext "Could not execute editor")" 
```

so the exit code would not have been 0 with the shell code.

agrn: Also, `transform_todos()` returns an error, but its errors were ignored by git-rebase--interactive.sh. Should I handle them?

agrn: Same question for `append_todo_help()`.

chriscool: I think for now it is ok if the errors that were ignored are still ignored.

builtin/rebase--helper.c
agrn: I just rerolled this patch to rename the variable.

agrn:
I just updated this branch.

sequencer.c
dscho: Or maybe it would be an obvious bonus of the rewrite ;-)

sequencer.c
```diff
@@ -4386,6 +4386,37 @@ int append_todo_help(unsigned edit_todo, unsigned keep_empty)
...
 }
 
+int edit_todo_list(unsigned flags)
+{
+	struct strbuf buf = STRBUF_INIT;
+	const char *todo_file = rebase_path_todo();
+	FILE *todo;
+
+	if (strbuf_read_file(&buf, todo_file, 0) < 0)
+		return error(_("could not read '%s'."), todo_file);
```
chriscool: Maybe use error_errno() instead of error().

chriscool:
The updated branch looks good to me. I just commented about maybe using error_errno() instead of error() in one place.

chriscool:
Yeah I think you can send it!

dscho:
Looks good to me, too!

agrn:
Following [Phillip’s feedback on the mailing list](https://public-inbox.org/git/3bfd3470-4482-fe6a-2cd9-08311a0bbaac@talktalk.net/), I added a new funtion to launch the sequence editor. Any thoughts ?

editor.c
```diff
@@ -95,3 +95,21 @@ int launch_editor(const char *path, struct strbuf *buffer, const char *const *en
...
+{
+	const char *editor = getenv("GIT_SEQUENCE_EDITOR");
+
+	if (!editor)
+		git_config_get_string_const("sequence.editor", &editor);
+	if (!editor)
+		return launch_editor(path, buffer, env);
+
+	return launch_specified_editor(editor, path, buffer, env);
+}
```
chriscool: Maybe have a git_sequence_editor() function that finds the right editor for a sequence and then launch_sequence_editor() could just call `launch_specified_editor(git_sequence_editor(), path, buffer, env);`

chriscool:
About "This is done to benefit from the existing code." in the commit message, I don't think it provides much information, so either you should remove the sentence or you should explain how exactly it benefits existing code.

Maybe you could explain that as part of the rewrite in C of the functionality to edit the todo list we need to launch the sequence editor and so on.

agrn:
This should be better.

dscho:
@agrn these patches have made it into `master` via b49ef560ed66, haven't they?

PR #504, opened 2018-06-07
WIP: Add GET_OID_GENTLY

ungps:


tree-walk.c
```diff
@@ -547,13 +562,25 @@ int get_tree_entry(const struct object_id *tree_oid, const char *name, struct ob
 		retval = -1;
 	} else {
 		struct tree_desc t;
-		init_tree_desc(&t, tree, size);
-		retval = find_tree_entry(&t, name, oid, mode);
+		if (gently) {
+			if (init_tree_desc_gently(&t, tree, size)) {
+				free(tree);
+				return -1;
```
tgummerer: I think it would be better to set `retval` to -1 here, and then `goto done`.  If at some point this function grows, and we allocate some more memory before here, when we make the change we would probably forget to free it.  Although I see we do things the same way in line 557-558, so maybe that's not a valid concern.

tree-walk.c
```diff
@@ -521,19 +528,27 @@ static int find_tree_entry(struct tree_desc *t, const char *name, struct object_
...
 			return 0;
 		}
-		return get_tree_entry(oid, name + entrylen, result, mode);
+		return get_tree_entry_gently(oid, name + entrylen, result, mode, gently);
 	}
 	return -1;
 }
 
-int get_tree_entry(const struct object_id *tree_oid, const char *name, struct object_id *oid, unsigned *mode)
+int get_tree_entry_gently(const struct object_id *tree_oid, const char *name, struct object_id *oid, unsigned *mode, int gently)
```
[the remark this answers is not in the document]
ungps: I guess that the best way would be to do as `Documentation/CodingGuidelines` states. In a future patch (since it is not related to `GET_OID_GENTLY`), I will fix the existing tiny mistakes regarding code styling.

I will edit this.

tree-walk.c
ungps: It makes sense. I am not sure what is git org opinion on using `goto` statements. In most cases, such a statement would make the code more unreadable; but I think in this case it would be good alternative and actually make the code future-proof.

sha1-name.c
```diff
@@ -1582,36 +1598,58 @@ static void diagnose_invalid_index_path(int stage,
...
 
+static const char *resolve_error = "dummy";
 
-static char *resolve_relative_path(const char *rel)
+static char *resolve_relative_path_gently(const char *rel, int gently)
 {
-	if (!starts_with(rel, "./") && !starts_with(rel, "../"))
+	if (!starts_with(rel, "./") && !starts_with(rel, "../")) {
 		return NULL;
+	}
```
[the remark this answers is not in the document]
ungps: Right. Old habits die hard.

sha1-name.c
```diff
@@ -802,8 +803,13 @@ static int get_oid_basic(const char *str, int len, struct object_id *oid,
...
 		}
-		if (read_ref_at(real_ref, flags, at_time, nth, oid, NULL,
-				&co_time, &co_tz, &co_cnt)) {
+
+		ret = read_ref_at(real_ref, flags, at_time, nth, oid, NULL,
+				&co_time, &co_tz, &co_cnt);
+		if (ret == -1) {
+			return -1;
+		}
+		if (ret) {
```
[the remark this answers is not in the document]
ungps: I guess it should be guarded against the die, just to be safe and make code more easy to be maintained.

tree-walk.c
tgummerer:
```
$ git grep goto | wc -l
1032
```
While there may be a few false positives in there, from a quick look it looks like mostly these are `goto` statements.  So I'd say it's fairly widespread, and I do feel like it would make sense to use it here.

dscho: It *is* fairly common to use `goto` in Git's source code to simplify resource management (`free`, `close`) in error code paths.

ungps:
I changed the name of the commits to be more alike to the format you suggested. I think that I should also squash `WIP: tree-walk: add two new gentle helpers` and `WIP: tree-walk: add one new gentle helper` int one commit.. What do you say?

About the rebase, I do not remember what I did (I bet it was something stupid), but now it's all good! :+1:

dscho:
> I changed the name of the commits to be more alike to the format you suggested.

Thanks!

> I think that I should also squash WIP: tree-walk: add two new gentle helpers and WIP: tree-walk: add one new gentle helper into one commit.. What do you say?

That sounds like a good idea.

> About the rebase, I do not remember what I did (I bet it was something stupid), but now it's all good! 👍

:-) I still surprise myself at times with entertaining rebase results...

BTW how about running this through `git rebase --autosquash`? By now, I think I forgot enough about this PR that I would like to start a fresh review from scratch, sound okay?

ungps:
Done!

ungps:
Done!

dscho:
The test failure looks like some connectivity issue on Travis' side ("No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.")

cache.h
```diff
@@ -1314,12 +1314,14 @@ struct object_context {
...
 #define GET_OID_RECORD_PATH     0200
 #define GET_OID_ONLY_TO_DIE    04000
+#define GET_OID_GENTLY		  010000
 
 #define GET_OID_DISAMBIGUATORS \
 	(GET_OID_COMMIT | GET_OID_COMMITTISH | \
 	GET_OID_TREE | GET_OID_TREEISH | \
 	GET_OID_BLOB)
 
+extern int get_oid_gently(const char *str, struct object_id *oid, int gently);
```
dscho: My instinctive reaction was that the `gently` is correct in the function name, but it would be `gentle` in the parameter name.

It seems that Git's source code agrees, by 6:3 votes (compare with `git grep -c ', int gentle'` vs `git grep -c ', int gently'`).

sha1-name.c
```diff
@@ -1460,10 +1468,16 @@ int strbuf_check_branch_ref(struct strbuf *sb, const char *name)
...
  * notably "xyz^" for "parent of xyz"
  */
-int get_oid(const char *name, struct object_id *oid)
+int get_oid_gently(const char *name, struct object_id *oid, int gently)
 {
 	struct object_context unused;
-	return get_oid_with_context(name, 0, oid, &unused);
+	if (gently)
+		gently = GET_OID_GENTLY;
+	return get_oid_with_context(name, gently, oid, &unused);
```
dscho: Or put it directly into the call:

```c
return get_oid_with_context(name, gently ? GET_OID_GENTLY : 0, oid, &unused);
```

(It a good practice to avoid changing parameters, in particular when you change their role (from a Boolean to a bitfield of flags, in this case)).

sha1-name.c
```diff
@@ -1513,24 +1527,29 @@ int get_oid_blob(const char *name, struct object_id *oid)
...
 		    filename, object_name_len, object_name);
+	}
 	if (is_missing_file_error(errno)) {
 		char *fullname = xstrfmt("%s%s", prefix, filename);
-
+		if (gently)
+			return -1;
 		if (!get_tree_entry(tree_oid, fullname, &oid, &mode)) {
 			die("Path '%s' exists, but not '%s'.\n"
 			    "Did you mean '%.*s:%s' aka '%.*s:./%s'?",
```
dscho: I wonder whether we need those error messages (as warnings in the gentle case). Maybe note this in the commit message as something we might want to consider to do differently?

sha1-name.c
```diff
@@ -1566,11 +1587,14 @@ static void diagnose_invalid_index_path(int stage,
...
 		ce = active_cache[pos];
 		if (ce_namelen(ce) == namelen &&
-		    !memcmp(ce->name, filename, namelen))
+		    !memcmp(ce->name, filename, namelen)) {
+			if (gently)
+				return -1;
 			die("Path '%s' is in the index, but not at stage %d.\n"
 			    "Did you mean ':%d:%s'?",
 			    filename, stage,
 			    ce_stage(ce), filename);
```
dscho: Likewise, this message seems pretty useful, also in the gentle case.

sha1-name.c
```diff
@@ -1582,31 +1606,52 @@ static void diagnose_invalid_index_path(int stage,
...
 		if (ce_namelen(ce) == fullname.len &&
-		    !memcmp(ce->name, fullname.buf, fullname.len))
+		    !memcmp(ce->name, fullname.buf, fullname.len)) {
+			if (gently)
+				return -1;
 			die("Path '%s' is in the index, but not '%s'.\n"
 			    "Did you mean ':%d:%s' aka ':%d:./%s'?",
 			    fullname.buf, filename,
 			    ce_stage(ce), fullname.buf,
 			    ce_stage(ce), filename);
```
dscho: And also here, I could imagine that the error message might need to be printed even when returning -1.

OTOH this might merit turning `gently` into `flags` so that it can be gentle and/or quiet. Dunno. What do you think, @ungps?

sha1-name.c
```diff
@@ -1582,36 +1598,58 @@ static void diagnose_invalid_index_path(int stage,
...
 {
-	if (!starts_with(rel, "./") && !starts_with(rel, "../"))
+	if (!starts_with(rel, "./") && !starts_with(rel, "../")) {
 		return NULL;
+	}
 
-	if (!is_inside_work_tree())
+	if (!is_inside_work_tree()) {
+		if (gently) {
+			return (char*) resolve_error;
```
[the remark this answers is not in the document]
dscho: The contract of this function is to return `NULL` if nothing was resolved. I think we should keep this behavior, and output the message instead.

sha1-name.c
dscho: Oh! I just saw that there is already a `GET_OID_QUIETLY`! So maybe we want to accept a parameter `flags` here, and also fix the issue where the `die()` prints something despite `flags & GET_OID_QUIETLY` possibly being non-zero.

I think I would do this, as a separate patch that I would insert before this one: add the `int flags` parameter and inserting `if (flags & GET_OID_QUIETLY) exit(128);` before the `die()` call.

sha1-name.c
```diff
@@ -1582,31 +1606,52 @@ static void diagnose_invalid_index_path(int stage,
...
+	}
 
 	strbuf_release(&fullname);
+	return 0;
 }
 
+static const char *resolve_error = "dummy";
 
-static char *resolve_relative_path(const char *rel)
+static char *resolve_relative_path_gently(const char *rel, int gently)
```
dscho: As you introduce the `, int gently` parameter, I think it is incorrect to rename the function to `*_gently()`: it is only gentle if that flag is set. So if it was up to me, I would simply keep the old function name. (And not even add the `gentle` parameter, but simply accept the `flags` unchanged, then test `flags & GET_OID_GENTLY`.)

sha1-name.c
```diff
@@ -1729,7 +1783,7 @@ static int get_oid_with_context_1(const char *name,
 					diagnose_invalid_oid_path(prefix,
 								   filename,
 								   &tree_oid,
-								   name, len);
+								   name, len, 0);
```
dscho: That's surprising. Why pass `0` even if the `flags` bitfield contains `GET_OID_GENTLY`?

sha1-name.c
```diff
@@ -1715,7 +1766,10 @@ static int get_oid_with_context_1(const char *name,
 			const char *filename = cp+1;
 			char *new_filename = NULL;
 
-			new_filename = resolve_relative_path(filename);
+			new_filename = resolve_relative_path_gently(filename,
+				flags & GET_OID_GENTLY);
+			if (new_filename == resolve_error)
```
dscho: Oh, I see!

And now I remember that I even suggested this strategy: use a bogus return value to indicate an error. Forget what I said earlier about returning a `NULL`. But we need to add a code comment here, as well as where we return that `(char *)resolve_error` to say that this return value is not intended to be used other than to test for an error.

tree-walk.c
```diff
@@ -501,7 +503,13 @@ static int find_tree_entry(struct tree_desc *t, const char *name, struct object_
...
 		oid = tree_entry_extract(t, &entry, mode);
 		entrylen = tree_entry_len(&t->entry);
-		update_tree_entry(t);
+
+		if (gently) {
+			if (update_tree_entry_gently(t))
+				return -1;
+		} else {
+			update_tree_entry(t);
+		}
```
dscho: Maybe invert the logic to make it easier to read?

```c
if (!gently)
    update_tree_entry(t);
else if (update_tree_entry_gently(t))
    return -1;
```

tree-walk.c
```diff
@@ -491,7 +491,9 @@ struct dir_state {
 	struct object_id oid;
 };
 
-static int find_tree_entry(struct tree_desc *t, const char *name, struct object_id *result, unsigned *mode)
+static int find_tree_entry_gently(struct tree_desc *t, const char *name,
```
dscho: Same comment as before: the parameter `gently` decides whether this function is gentle or not. So it would probably make more sense to *not* rename the function (this applies to all of the renamed functions in this patch).

tree-walk.c
```diff
@@ -521,19 +529,28 @@ static int find_tree_entry(struct tree_desc *t, const char *name, struct object_
...
-	tree = read_object_with_reference(tree_oid, tree_type, &size, &root);
+	if (gently) {
+		struct object_info oi = OBJECT_INFO_INIT;
+
+		if (oid_object_info_extended(the_repository, tree_oid, &oi, 0) < 0)
+			return -1;
+		tree = *oi.contentp;
+	} else {
+		tree = read_object_with_reference(tree_oid, tree_type, &size, &root);
+	}
```
dscho: A good rule to follow is to handle the easier code block first. This not only resembles natural human behavior 😄 but it also makes it easier to follow the logic for the occasional reader:

```c
if (!gently)
	tree = read_object_with_reference(tree_oid, tree_type, &size, &root);
else {
	struct object_info oi = OBJECT_INFO_INIT;

	if (oid_object_info_extended(the_repository, tree_oid, &oi, 0) < 0)
		return -1;
	tree = *oi.contentp;
}
```

tree-walk.c
```diff
@@ -547,13 +564,26 @@ int get_tree_entry(const struct object_id *tree_oid, const char *name, struct ob
...
-		init_tree_desc(&t, tree, size);
-		retval = find_tree_entry(&t, name, oid, mode);
+		if (gently) {
+			if (init_tree_desc_gently(&t, tree, size)) {
+				free(tree);
+				return -1;
+			}
+		} else {
+			init_tree_desc(&t, tree, size);
+		}
```
dscho: Here again, the simpler arm of the `if ... else` should come first.

tree-walk.c
```diff
@@ -547,13 +564,26 @@ int get_tree_entry(const struct object_id *tree_oid, const char *name, struct ob
...
+		retval = find_tree_entry_gently(&t, name, oid, mode, gently);
 	}
 	free(tree);
 	return retval;
 }
 
+int get_tree_entry(const struct object_id *tree_oid, const char *name,
```
dscho: I guess this function is used in too many places (23, if I counted correctly) to change the signature: You would have to add tons of `, 0`, making the patch harder to read. So I'd be okay with renaming `get_tree_entry()` (but I would suggest describing the reason in the commit message).

tree-walk.c
```diff
@@ -600,9 +630,20 @@ enum follow_symlinks_result get_tree_entry_follow_symlinks(struct object_id *tre
...
+				    &current_tree_oid, &oi, 0) < 0) {
+					retval = MISSING_OBJECT;
+					goto done;
+				}
+				tree = *oi.contentp;
+			} else {
+				tree = read_object_with_reference(&current_tree_oid,
+								  tree_type, &size,
+								  &root);
+			}
```
dscho: Once again: simpler conditional block first.

dscho: (I'll not repeat this for the rest of the file, it really applies to all if...else conditionals ;-))

tree-walk.c
```diff
@@ -752,6 +825,12 @@ enum follow_symlinks_result get_tree_entry_follow_symlinks(struct object_id *tre
 	return retval;
 }
 
+enum follow_symlinks_result get_tree_entry_follow_symlinks(struct object_id *tree_oid, const char *name, struct object_id *result, struct strbuf *result_path, unsigned *mode)
```
dscho: There are only three callers of this function, currently. I think we can change the signature to accept that `gentle` flag.

sha1-name.c
```diff
@@ -1773,12 +1773,12 @@ static int get_oid_with_context_1(const char *name,
 			if (new_filename)
```
dscho: commit subject: `make ... more gentle` (it's an adjective in this form, not an adverb).

sha1-name.c
```diff
@@ -1773,12 +1773,12 @@ static int get_oid_with_context_1(const char *name,
...
-				ret = get_tree_entry_follow_symlinks(&tree_oid,
+				ret = get_tree_entry_follow_symlinks_gently(&tree_oid,
 					filename, oid, &oc->symlink_path,
-					&oc->mode);
+					&oc->mode, flags & GET_OID_GENTLY);
 			} else {
-				ret = get_tree_entry(&tree_oid, filename, oid,
-						     &oc->mode);
+				ret = get_tree_entry_gently(&tree_oid, filename, oid,
+						     &oc->mode, flags & GET_OID_GENTLY);
```
dscho: It would make sense to invert the commit order, I guess, to make lower-level functions optionally gentle first, and only then use that in subsequent commits to make callers more gentle.

(I know, this is not the way it was developed, but it sounds more logical to me, as a reader of the patches.)

dscho:
@ungps I guess this is still an ongoing project, eh?

dscho:
@ungps ping?

ungps:
> @ungps ping?

Sorry for late reply; I was caught with exams and wasn't so active online. But yes, I hope I will resume the work in a week.

dscho:
> I hope I will resume the work in a week.

Excellent!

dscho:
@ungps ping?

ungps:
Hello.

It will take more time than I thought. I started an internship on 1st of July and my schedule is pretty messed up (mostly because of me, not of the internship). Also, I find this issue to be a little bit difficult to work on because it takes a lot time to explore the codebase (go a long way back in the caller graph, through all callees and understand what happens there). I need to find a big window of free time instead of trying to work "fragmented" because I lose the flow in between.

dscho:
Makes sense. So let's keep this PR open.

dscho:
@ungps are you still planning on working on this? Otherwise, I think it would be okay to simply close this PR.

ungps:
Hi @dscho .
For the moment I think it is better to just close it since I do not know exactly when I will find enough time. I hope that I will be able to work as soon as possible.

PR #506, opened 2018-06-13
rebase--interactive: rewrite reflog-related operation in C

agrn:
This rewrites setup_reflog_action() and checkout_onto() from shell to C. The relevant commits are the last three.

For setup_reflog_action(), I need to pass $switch_to.  For
checkout_onto(), I need to pass $onto, $onto_name, and $orig_head.

Now, for $onto and $onto_name, I thought I would call the parameters
`--onto` and `--onto-name` (and I don’t know what to set as a help
message).  For $switch_to and $orig_head, I set `--commit-id`, and I’m
not really proud of it.

Any ideas?


agrn:

builtin/rebase--helper.c
```diff
@@ -50,6 +52,10 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 		OPT_CMDMODE(0, "edit-todo", &command,
 			    N_("edit the todo list during an interactive rebase"),
 			    EDIT_TODO),
+		OPT_CMDMODE(0, "setup-reflog", &command,
+			    N_("setup the reflog action"), SETUP_REFLOG),
+		OPT_STRING(0, "commit-id", &commit_id, NULL,
+			   N_("the commit to checkout")),
```
chriscool: I would say that if setup-reflog should always be passed a commit id, then we may assume that the commit id always follow `--setup-reflog` and we don't need a special `--commit-id` option. See how it works for `--add-exec-commands`.

builtin/rebase--helper.c
```diff
@@ -93,5 +99,7 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 		return !!append_todo_help(0, keep_empty);
 	if (command == EDIT_TODO && argc == 1)
 		return !!edit_todo_list(flags);
+	if (command == SETUP_REFLOG)
```
chriscool: It looks like there is no check for the number of arguments that SETUP_REFLOG expects.

git-rebase--interactive.sh
```diff
@@ -72,6 +72,7 @@ collapse_todo_ids() {
 
 # Switch to the branch in $into and notify it in the reflog
 checkout_onto () {
+	comment_for_reflog start
```
chriscool: From the commit message this is added because "$GIT_SEQUENCE_EDITOR is not longer set at the first call of
checkout_onto()". So I would say that this change is related to the previous commit that does not set $GIT_SEQUENCE_EDITOR anymore at the first call of checkout_onto() and should be moved in that previous commit.

sequencer.c
```diff
@@ -3135,6 +3135,34 @@ static const char *reflog_message(struct replay_opts *opts,
...
+
+	return run_command_silent(&cmd, !verbose);
+}
+
+int setup_reflog_action(struct replay_opts *opts, const char *switch_to,
+			int verbose)
+{
+	const char *action;
+
+	if (strlen(switch_to) > 0) {
```
chriscool: We tend to use `if (*switch_to)` or maybe `if (switch_to && *switch_to)`.

sequencer.h
```diff
@@ -100,6 +100,9 @@ int update_head_with_reflog(const struct commit *old_head,
 void commit_post_rewrite(const struct commit *current_head,
 			 const struct object_id *new_head);
 
+int setup_reflog_action(struct replay_opts *opts, const char *switch_to,
```
chriscool: I would suggest "commit_id" instead of "switch_to".

sequencer.c
```diff
@@ -3135,6 +3135,34 @@ static const char *reflog_message(struct replay_opts *opts,
...
+	cmd.git_cmd = 1;
+
+	argv_array_push(&cmd.args, "checkout");
+	argv_array_push(&cmd.args, commit);
+	argv_array_pushf(&cmd.env_array, GIT_REFLOG_ACTION "=%s", action);
+
+	return run_command_silent(&cmd, !verbose);
+}
+
+int setup_reflog_action(struct replay_opts *opts, const char *switch_to,
```
chriscool: I would suggest "commit_id" instead of "switch_to".

chriscool: Or maybe "commit" to be consistent with checkout_base_commit().

git-rebase--interactive.sh
agrn: It’s already in the commit that removes setup_reflog_action()

builtin/rebase--helper.c
agrn: Oh, it also looks like --checkout-onto always takes $orig_head, $onto, and $onto_name… so I will do something like that here, too. Sorry for the noise.

sequencer.c
```diff
@@ -766,6 +766,32 @@ N_("you have staged changes in your working tree\n"
 #define VERIFY_MSG  (1<<4)
 #define CREATE_ROOT_COMMIT (1<<5)
 
+static int run_command_silent(struct child_process *cmd,
+			      unsigned silent)
+{
+	if (silent) {
```
chriscool: Maybe you could just `return run_command(cmd);` in the !silent case first. This would decrease indentation in the rest of the function.

agrn:
I just updated my branch.

sequencer.c
```diff
@@ -861,21 +879,7 @@ static int run_git_commit(const char *defmsg, struct replay_opts *opts,
...
-				      NULL, 0,
-				      &buf, 0);
-		if (rc)
-			fputs(buf.buf, stderr);
-		strbuf_release(&buf);
-		return rc;
-	}
-
-	return run_command(&cmd);
+	return run_command_silent(&cmd, is_rebase_i(opts) && !(flags & EDIT_MSG));
```
chriscool: Another possibility is to remove the silent argument from run_command_silent() and to have:

```
if (is_rebase_i(opts) && !(flags & EDIT_MSG))
        return run_command_silent(&cmd);
else
        return run_command(&cmd);
```

builtin/rebase--helper.c
```diff
@@ -98,5 +101,7 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 		return !!edit_todo_list(flags);
 	if (command == SETUP_REFLOG && argc > 1)
 		return !!setup_reflog_action(&opts, argv[1], verbose);
+	if (command == CHECKOUT_ONTO && argc > 3)
```
chriscool: What happens if argc is 5 or 6 for example?

sequencer.c
agrn: True. It’s probably less confusing.

git-rebase--interactive.sh
```diff
@@ -176,7 +168,8 @@ EOF
 
 	git rebase--helper --check-todo-list || {
 		ret=$?
-		checkout_onto
+		git rebase--helper --checkout-onto "$onto_name" "$onto" \
+		    $orig_head ${verbose:+--verbose}
```
chriscool: Maybe also quote $orig_head

git-rebase--interactive.sh
```diff
@@ -186,7 +179,8 @@ EOF
 	onto="$(git rebase--helper --skip-unnecessary-picks)" ||
 	die "Could not skip unnecessary pick commands"
 
-	checkout_onto
+	git rebase--helper --checkout-onto "$onto_name" "$onto" $orig_head \
+	    ${verbose:+--verbose}
```
chriscool: Maybe also quote $orig_head

builtin/rebase--helper.c
```diff
@@ -98,5 +101,7 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
...
+		return !!checkout_onto(&opts, argv[1], argv[2], argv[3], verbose);
```
chriscool: The commit message still talks about 'two new flags: “onto-name” and “onto”'. It looks like it needs an update.

builtin/rebase--helper.c
agrn: It still runs. Is this a real problem? We just ignore them.

sequencer.c
```diff
@@ -3162,6 +3162,26 @@ int setup_reflog_action(struct replay_opts *opts, const char *commit,
...
+int checkout_onto(struct replay_opts *opts,
+		  const char *onto_name, const char *onto,
+		  const char *orig_head, unsigned verbose)
+{
+	struct object_id oid;
+	const char *action = reflog_message(opts, "start", "checkout %s", onto_name);
+
+	if (get_oid(orig_head, &oid))
+		/* TODO: translation? */
+		die("%s: not a valid SHA1", orig_head);
```
chriscool: Where does this error message comes from? At least it needs s/SHA1/OID/.

builtin/rebase--helper.c
chriscool: Yeah, but I think we prefer to die in this case as it means that there is probably an issue in the calling shell script.

chriscool: So the `argc > 3` should probably be `argc == 4`.

agrn: (sorry, when argc is 5, it’s when rebase--helper was called with the `--verbose` flag)

agrn: Right.

chriscool: I think that the function that parses options removes the flags it interprets.

sequencer.c
agrn: It comes from here: https://github.com/git/git/blob/master/builtin/update-ref.c#L414-L415

update-ref was originally called by checkout_onto().

agrn:
Updated.

sequencer.c
```diff
@@ -766,6 +766,27 @@ N_("you have staged changes in your working tree\n"
 #define VERIFY_MSG  (1<<4)
 #define CREATE_ROOT_COMMIT (1<<5)
 
+static int run_command_silent(struct child_process *cmd)
```
stefanbeller: Maybe rename it to run_command_silent_if_successful or silent_on_success ?

sequencer.c
```diff
@@ -766,6 +766,27 @@ N_("you have staged changes in your working tree\n"
...
+{
+	struct strbuf buf = STRBUF_INIT;
+	int rc;
+
+	/* hide stderr on success */
+	cmd->stdout_to_stderr = 1;
+	cmd->err = -1;
```
stefanbeller: From the comment of pipe_command:

 * The fields of "cmd" should be set up as they would for a normal run_command
 * invocation. But note that there is no need to set the in, out, or err
 * fields; pipe_command handles that automatically.

I don't think you need to set cmd.err here

stefanbeller: Oh, before this was done to signal if we need to have the silen_on_success invocation, but now it is not needed any more as the signaling is done otherwise.

sequencer.c
chriscool: Ok, I still think "OID" or "object id" are better than SHA1 these days.

agrn: OK. Do I wrap this in a `_()`?

dscho: Yes, please!

sequencer.c
```diff
@@ -3162,6 +3162,25 @@ int setup_reflog_action(struct replay_opts *opts, const char *commit,
...
+	if (get_oid(orig_head, &oid))
+		die(_("%s: not a valid OID"), orig_head);
+
+	if (checkout_base_commit(opts, onto, verbose, action)) {
+		apply_autostash(opts);
+		sequencer_remove_state(opts);
+		die(_("could not detach HEAD"));
+	}
+
+	return update_ref(NULL, "ORIG_HEAD", &oid, NULL, 0, UPDATE_REFS_DIE_ON_ERR);
```
chriscool: Is there a reason why UPDATE_REFS_DIE_ON_ERR is used? I am not sure the original would have died if the ref update failed.

agrn: It’s used in update-ref.c here: https://github.com/git/git/blob/master/builtin/update-ref.c#L440-L442

chriscool: Yeah, I think you are right.

chriscool:
This LGTM now. Thanks!

agrn:
I forgot to remove `comment_for_reflog()`. Is this good for the list?

sequencer.c
dscho: I think in this case, we do not want to `die()`, though. We're in a library function that is supposed to be called from wherever, not just from code paths that want to die with a potentially confusing error message. `UPDATE_REFS_MSG_ON_ERR` would appear more appropriate in this case, to me. Looking at the [`update_refs()` code](https://github.com/git/git/blob/68372c88794aba15f853542008cda39def768372/refs.c#L1105-L1120), I am even more certain of this.

agrn: Right. I see `update_refs()` is also used elsewhere in sequencer.c, and they all use `UPDATE_REFS_MSG_ON_ERR`.

agrn:
I updated my branch. Travis seems to have some problems these days :/

chriscool:
Yeah, it looks like Travis had some issues. Maybe you can try again the tests that failed?

agrn:
Done!

dscho:
@agrn these commits were applied and integrated into `master` via b49ef560ed66, weren't they?

PR #507, opened 2018-06-15
WIP: make apply_autostash available to use for rebase builtin

prertik:
The approach is to teach apply_autostash about rebase_apply directory
so that it can use from the rebase builtin.

Signed-off-by: Pratik Karki <predatoramigo@gmail.com>


prertik:
How about this?

chriscool:
I agree with @dscho and @stefanbeller.

PR #511, opened 2018-06-16
rebase --root: fix `reword` on a root commit

dscho:
Todd Zullinger reported this bug in https://public-inbox.org/git/20180615043111.GS3094@zaya.teonanacatl.net/: when calling `git rebase --root` and trying to reword the root commit's message, a BUG is reported.

This fixes that.

IMO the bug fix is trivial enough to qualify for inclusion into v2.18.0, still.


dscho:
Continuing here: https://github.com/gitgitgadget/git/pull/3

PR #512, opened 2018-06-17
Fix PROMPT_COMMAND overwriting default behavior

JBallin:
Currently removes shell_session_history_check on login and update_terminal_cwd at each prompt

### DIFF
--- a/contrib/completion/git-prompt.sh
+++ b/contrib/completion/git-prompt.sh
@@ -24,6 +24,10 @@
 #          will show username, at-sign, host, colon, cwd, then
 #          various status string, followed by dollar and SP, as
 #          your prompt.
+#	   NOTE: To mimic terminal's default $PROMPT_COMMAND 
+#          (history check on login and update tab title to current dir):
+#            shell_session_history_check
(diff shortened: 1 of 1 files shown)


JBallin:
Upon [further research](https://unix.stackexchange.com/questions/450492/what-does-shell-session-history-check-do/450493#450493), I've discovered that `shell_session_history_check` isn't necessary for all users and may only apply to newer macs. Not calling it won't cause any issues and may just cause confusion. Voting to leave as is.

Note that the proper way would be `PROMPT_COMMAND='shell_session_history_check; ...'` instead of calling it outside of the `PROMPT_COMMAND` (gets called too early). The function will automatically remove itself from the `PROMPT_COMMAND`.

JBallin:
Upon [further research](https://unix.stackexchange.com/questions/450492/what-does-shell-session-history-check-do/450493#450493), I've discovered that `shell_session_history_check` isn't necessary for all users and may only apply to newer macs. Not calling it won't cause any issues and may just cause confusion. Voting to leave as is.

Note that the proper way would be `PROMPT_COMMAND='shell_session_history_check; ...'` instead of calling it outside of the `PROMPT_COMMAND` (gets called too early). The function will automatically remove itself from the `PROMPT_COMMAND`.

PR #513, opened 2018-06-21
rebase--interactive: rewrite complete_action() in C

agrn:
This rewrites complete_action() from shell to C. The relevant commits are the last two.


agrn:
@dscho @stefanbeller @chriscool

git-rebase--interactive.sh
```diff
@@ -127,54 +127,6 @@ init_revisions_and_shortrevisions () {
...
-
-$comment_char $(eval_ngettext \
-	"Rebase \$shortrevisions onto \$shortonto (\$todocount command)" \
-	"Rebase \$shortrevisions onto \$shortonto (\$todocount commands)" \
-	"$todocount")
-EOF
-	git rebase--helper --append-todo-help ${keep_empty:+--keep-empty}
-
-	has_action "$todo" ||
-		return 2
```
chriscool: Why isn't there code for the above 2 lines in the function you wrote?

git-rebase--interactive.sh
```diff
@@ -127,54 +127,6 @@ init_revisions_and_shortrevisions () {
...
-
-	cp "$todo" "$todo".backup
-	collapse_todo_ids
-	git_sequence_editor "$todo" ||
-		die_abort "$(gettext "Could not execute editor")"
-
-	has_action "$todo" ||
-		return 2
```
chriscool: Here also why isn't there code for the above 2 lines in the function you wrote?

sequencer.c
```diff
@@ -4374,11 +4377,9 @@ int check_todo_list(void)
 	return res;
 }
 
-int append_todo_help(unsigned edit_todo, unsigned keep_empty)
+static void append_todo_help(unsigned edit_todo, unsigned keep_empty,
+			     struct strbuf *buf)
```
chriscool: I think it would be easier to review if changes to this function were in a separate commit.

admin7770:
Danke ;)

Christian Couder <notifications@github.com> schrieb am Fr., 22. Juni 2018,
09:26:

sequencer.c
```diff
@@ -4479,7 +4473,7 @@ static int rewrite_file(const char *path, const char *buf, size_t len)
 }
 
 /* skip picking commits whose parents are unchanged */
-int skip_unnecessary_picks(void)
+static int skip_unnecessary_picks(const char **output_oid)
```
chriscool: Maybe changes to this function could also go to another commit.

admin7770:
Ok sorry

Christian Couder <notifications@github.com> schrieb am Fr., 22. Juni 2018,
09:32:

builtin/rebase--helper.c
```diff
@@ -41,21 +43,17 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 			N_("expand commit ids in the todo list"), EXPAND_OIDS),
 		OPT_CMDMODE(0, "check-todo-list", &command,
 			N_("check the todo list"), CHECK_TODO_LIST),
-		OPT_CMDMODE(0, "skip-unnecessary-picks", &command,
-			N_("skip unnecessary picks"), SKIP_UNNECESSARY_PICKS),
```
chriscool: Maybe removel of useless modes could be done in another commit.

chriscool:
There is "rebase--helper: remove checkout-onto" after the Signed-off-by at the end of the commit that rewrites complete_action().

git-rebase--interactive.sh
agrn: Return code 2 is a special case for interactive rebase. It tells git-rebase.sh to apply the autostash, to remove the state dir and to die with the message “nothing to do” ([see here](https://github.com/git/git/blob/master/git-rebase.sh#L215-L220)). I decided to rewrite this case in C, so instead of `return 2`, there is [this](https://github.com/git/git/pull/513/files/d77c4414835ec158c87d863e849f1c5f9c322a6d..0b4b1c0fb3cd9bbc80f1914ebbebf2a81b38886d#diff-79231f0693f84f3951daeea17065aad9R4615).

After the call to `--append-todo-help`, I noticed that the todo-list has at least one action: if the todo file is empty, then complete_action() writes a noop there. So I removed this specific check from my code.

agrn: It also reminds me that I forgot to remove the handling of return code 2 in git-rebase.sh.

agrn: Oh no, it would break git-rebase--preserve-merges.sh.

git-rebase--interactive.sh
agrn: Same as above, except here it is actually handled by my code [here](https://github.com/git/git/pull/513/files/d77c4414835ec158c87d863e849f1c5f9c322a6d..0b4b1c0fb3cd9bbc80f1914ebbebf2a81b38886d#diff-79231f0693f84f3951daeea17065aad9R4613).

admin7770:
Thank you :)

agrn <notifications@github.com> schrieb am Fr., 22. Juni 2018, 11:36:

git-rebase--interactive.sh
chriscool: Ok, please explain all that in the commit message.

admin7770:
Ok Work Tal you

Christian Couder <notifications@github.com> schrieb am Fr., 22. Juni 2018,
11:45:

agrn:
Updated. The relevant commits are the last four.

chriscool:
Yeah, this looks better to me now.

In the commit message I would suggest to start with the more general information and to have details toward the end.

agrn:
This should be better, then

chriscool:
Yeah, this looks good to me! Thanks!

dscho:
@agrn this PR should be addressed by b49ef560ed66, yes?

PR #517, opened 2018-07-03
clarify --no-hardlinks in repo shrinking checklist

jonorthwash:

### DIFF
--- a/Documentation/git-filter-branch.txt
+++ b/Documentation/git-filter-branch.txt
@@ -428,7 +428,9 @@ to clone, that keeps your original intact.
 
 * Clone it with `git clone file:///path/to/repo`.  The clone
   will not have the removed objects.  See linkgit:git-clone[1].  (Note
-  that cloning with a plain path just hardlinks everything!)
+  that cloning with a plain path just hardlinks everything!  Also note
+  that simply using `--no-hardlinks` with a plain path will not work
+  either.)
 
 If you really don't want to clone it, for whatever reasons, check the
 following points instead (in this order).  This is a very destructive


dscho:
@jonorthwash thank you for your contribution. Maybe you want to also suggest what to do instead? (I guess prefixing the plain path with `file://` should work...) It is always nicer to say what works than only what does not work.

Also, your commit is missing a good commit message and the [`Signed-off-by:` footer](https://git-scm.com/docs/SubmittingPatches#dco).

jonorthwash:
> It is always nicer to say what works than only what does not work.

Agreed.  Looking back at my PR now I'm sad that I didn't put that, since I have no idea anymore what I ended up doing that worked.

Perhaps someone who knows better could build on my commit?  If it can be accepted and built on, I'm happy to update the PR with the Signed-off-by footer.  Or, if you're sure `file://` will work, I can add that too...

dscho:
> if you're sure `file://` will work

I am indeed pretty sure, yet would still suggest that you give it a quick try, and then update your PR (and then send it to the Git mailing list via either [GitGitGadget](https://gitgitgadget.github.io/), [submitGit](https://submitgit.herokuapp.com/) or [manually](https://github.com/git-for-windows/git/blob/master/CONTRIBUTING.md#submit-your-patch).

dscho:
> would still suggest that you give it a quick try

Did you?

jonorthwash:
Was this merged (or the documentation otherwise updated)?

I never tried the `file://` approach because I've not needed to do this since then, but if it works, it makes sense for it to be added too!

jonorthwash:
Also note that the documentation already says to use `file://`.  My patch just adds one additional clarification to the list of things that won't work!  I don't think there's any reason to state again that `file://` is the right approach?

dscho:
> Was this merged (or the documentation otherwise updated)?

No. I asked for clarification on at least one issue, and did not receive any feedback. Now, I like to be given the silent treatment as much as the next person, which is pretty much not at all. I also figured that this meant that you were absolutely not interested in continuing with this PR.

To continue, I would _highly_ advise to give my suggestions at least a brief whirl, to _at least_ include in the commit message what you learned.

After that, as I pointed out, you will need to send the patch to the mailing list for review. You can use [GitGitGadget](https://gitgitgadget.github.io), [submitGit](https://submitgit.herokuapp.com) or [send it manually](https://github.com/git-for-windows/git/blob/master/CONTRIBUTING.md#submit-your-patch).

PR #518, opened 2018-07-04
rebase--interactive: rewrite init_revisions_and_shortrevisions() in C

agrn:
This rewrites init_revisions_and_shortrevisions() in C. The relevant commit is the last one.

### DIFF
--- a/builtin/rebase--helper.c
+++ b/builtin/rebase--helper.c
@@ -97,7 +97,7 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 	if (command == ADD_EXEC && argc == 2)
 		return !!sequencer_add_exec_commands(argv[1]);
 	if (command == APPEND_TODO_HELP && argc == 1)
-		return !!append_todo_help(0, keep_empty);
+		return !!append_todo_help_to_file(0, keep_empty);
 	if (command == EDIT_TODO && argc == 1)
 		return !!edit_todo_list(flags);
 	if (command == PREPARE_BRANCH && argc == 2)
--- a/builtin/rebase--helper.c
+++ b/builtin/rebase--helper.c
@@ -13,12 +13,12 @@ static const char * const builtin_rebase_helper_usage[] = {
 int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 {
 	struct replay_opts opts = REPLAY_OPTS_INIT;
-	unsigned flags = 0, keep_empty = 0, rebase_merges = 0, write_edit_todo = 0;
+	unsigned flags = 0, keep_empty = 0, rebase_merges = 0;
 	int abbreviate_commands = 0, rebase_cousins = -1;
 	enum {
 		CONTINUE = 1, ABORT, MAKE_SCRIPT, SHORTEN_OIDS, EXPAND_OIDS,
 		CHECK_TODO_LIST, SKIP_UNNECESSARY_PICKS, REARRANGE_SQUASH,
-		ADD_EXEC, APPEND_TODO_HELP
+		ADD_EXEC, APPEND_TODO_HELP, EDIT_TODO
 	} command = 0;
 	struct option options[] = {
 		OPT_BOOL(0, "ff", &opts.allow_ff, N_("allow fast-forward")),
@@ -28,8 +28,6 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 		OPT_BOOL(0, "rebase-merges", &rebase_merges, N_("rebase merge commits")),
 		OPT_BOOL(0, "rebase-cousins", &rebase_cousins,
 			 N_("keep original branch points of cousins")),
-		OPT_BOOL(0, "write-edit-todo", &write_edit_todo,
-			 N_("append the edit-todo message to the todo-list")),
 		OPT_CMDMODE(0, "continue", &command, N_("continue rebase"),
 				CONTINUE),
 		OPT_CMDMODE(0, "abort", &command, N_("abort rebase"),
@@ -50,6 +48,9 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 			N_("insert exec commands in todo list"), ADD_EXEC),
 		OPT_CMDMODE(0, "append-todo-help", &command,
 			    N_("insert the help in the todo list"), APPEND_TODO_HELP),
+		OPT_CMDMODE(0, "edit-todo", &command,
+			    N_("edit the todo list during an interactive rebase"),
+			    EDIT_TODO),
 		OPT_END()
 	};
 
@@ -90,6 +91,8 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 	if (command == ADD_EXEC && argc == 2)
 		return !!sequencer_add_exec_commands(argv[1]);
 	if (command == APPEND_TODO_HELP && argc == 1)
-		return !!append_todo_help(write_edit_todo, keep_empty);
+		return !!append_todo_help(0, keep_empty);
+	if (command == EDIT_TODO && argc == 1)
+		return !!edit_todo_list(flags);
 	usage_with_options(builtin_rebase_helper_usage, options);
 }
--- a/builtin/rebase--helper.c
+++ b/builtin/rebase--helper.c
@@ -18,7 +18,8 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 	enum {
 		CONTINUE = 1, ABORT, MAKE_SCRIPT, SHORTEN_OIDS, EXPAND_OIDS,
 		CHECK_TODO_LIST, SKIP_UNNECESSARY_PICKS, REARRANGE_SQUASH,
-		ADD_EXEC, APPEND_TODO_HELP, EDIT_TODO, PREPARE_BRANCH
+		ADD_EXEC, APPEND_TODO_HELP, EDIT_TODO, PREPARE_BRANCH,
+		CHECKOUT_ONTO
 	} command = 0;
 	struct option options[] = {
 		OPT_BOOL(0, "ff", &opts.allow_ff, N_("allow fast-forward")),
@@ -54,6 +55,8 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 			    EDIT_TODO),
 		OPT_CMDMODE(0, "prepare-branch", &command,
 			    N_("prepare the branch to be rebased"), PREPARE_BRANCH),
+		OPT_CMDMODE(0, "checkout-onto", &command,
+			    N_("checkout a commit"), CHECKOUT_ONTO),
 		OPT_END()
 	};
 
@@ -99,5 +102,7 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 		return !!edit_todo_list(flags);
 	if (command == PREPARE_BRANCH && argc == 2)
 		return !!prepare_branch_to_be_rebased(&opts, argv[1]);
+	if (command == CHECKOUT_ONTO && argc == 4)
+		return !!checkout_onto(&opts, argv[1], argv[2], argv[3]);
 	usage_with_options(builtin_rebase_helper_usage, options);
 }
--- a/builtin/rebase--helper.c
+++ b/builtin/rebase--helper.c
@@ -18,7 +18,7 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 	enum {
 		CONTINUE = 1, ABORT, MAKE_SCRIPT, SHORTEN_OIDS, EXPAND_OIDS,
 		CHECK_TODO_LIST, SKIP_UNNECESSARY_PICKS, REARRANGE_SQUASH,
-		ADD_EXEC, APPEND_TODO_HELP, EDIT_TODO
+		ADD_EXEC, APPEND_TODO_HELP, EDIT_TODO, PREPARE_BRANCH
 	} command = 0;
 	struct option options[] = {
 		OPT_BOOL(0, "ff", &opts.allow_ff, N_("allow fast-forward")),
@@ -28,6 +28,7 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 		OPT_BOOL(0, "rebase-merges", &rebase_merges, N_("rebase merge commits")),
 		OPT_BOOL(0, "rebase-cousins", &rebase_cousins,
 			 N_("keep original branch points of cousins")),
+		OPT__VERBOSE(&opts.verbose, N_("be verbose")),
 		OPT_CMDMODE(0, "continue", &command, N_("continue rebase"),
 				CONTINUE),
 		OPT_CMDMODE(0, "abort", &command, N_("abort rebase"),
@@ -51,6 +52,8 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 		OPT_CMDMODE(0, "edit-todo", &command,
 			    N_("edit the todo list during an interactive rebase"),
 			    EDIT_TODO),
+		OPT_CMDMODE(0, "prepare-branch", &command,
+			    N_("prepare the branch to be rebased"), PREPARE_BRANCH),
 		OPT_END()
 	};
 
@@ -94,5 +97,7 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 		return !!append_todo_help(0, keep_empty);
 	if (command == EDIT_TODO && argc == 1)
 		return !!edit_todo_list(flags);
+	if (command == PREPARE_BRANCH && argc == 2)
+		return !!prepare_branch_to_be_rebased(&opts, argv[1]);
 	usage_with_options(builtin_rebase_helper_usage, options);
 }
--- a/builtin/rebase--helper.c
+++ b/builtin/rebase--helper.c
@@ -3,6 +3,7 @@
 #include "config.h"
 #include "parse-options.h"
 #include "sequencer.h"
+#include "rebase-interactive.h"
 
 static const char * const builtin_rebase_helper_usage[] = {
 	N_("git rebase--helper [<options>]"),
@@ -12,12 +13,12 @@ static const char * const builtin_rebase_helper_usage[] = {
 int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 {
 	struct replay_opts opts = REPLAY_OPTS_INIT;
-	unsigned flags = 0, keep_empty = 0, rebase_merges = 0;
+	unsigned flags = 0, keep_empty = 0, rebase_merges = 0, write_edit_todo = 0;
 	int abbreviate_commands = 0, rebase_cousins = -1;
 	enum {
 		CONTINUE = 1, ABORT, MAKE_SCRIPT, SHORTEN_OIDS, EXPAND_OIDS,
 		CHECK_TODO_LIST, SKIP_UNNECESSARY_PICKS, REARRANGE_SQUASH,
-		ADD_EXEC
+		ADD_EXEC, APPEND_TODO_HELP
 	} command = 0;
 	struct option options[] = {
 		OPT_BOOL(0, "ff", &opts.allow_ff, N_("allow fast-forward")),
@@ -27,6 +28,8 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 		OPT_BOOL(0, "rebase-merges", &rebase_merges, N_("rebase merge commits")),
 		OPT_BOOL(0, "rebase-cousins", &rebase_cousins,
 			 N_("keep original branch points of cousins")),
+		OPT_BOOL(0, "write-edit-todo", &write_edit_todo,
+			 N_("append the edit-todo message to the todo-list")),
 		OPT_CMDMODE(0, "continue", &command, N_("continue rebase"),
 				CONTINUE),
 		OPT_CMDMODE(0, "abort", &command, N_("abort rebase"),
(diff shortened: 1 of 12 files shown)


builtin/rebase--helper.c
```diff
@@ -73,18 +78,59 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
...
+		if (upstream) {
+			struct object_id upstream_oid;
+
+			get_oid(upstream, &upstream_oid);
+			shortupstream = find_unique_abbrev(&upstream_oid, DEFAULT_ABBREV);
+			strbuf_addf(&revisions, "%s...%s", upstream, head_hash);
+			strbuf_addf(&shortrevisions, "%s..%s", shortupstream, shorthead);
+		} else {
+			strbuf_addf(&revisions, "%s...%s", onto, head_hash);
+			strbuf_add(&shortrevisions, shorthead, strlen(shorthead));
```
chriscool: Maybe xstrfmt() could be used, but I am not sure it would improve anything :-)

agrn:
Apparently that would do the same thing, but I think I prefer using xstrfmt() 
as we write once in the buffer, then access directly to its content.  That 
would be clearer, in my opinion.

agrn:
Ouch, there is a warning I didn’t noticed in my code.

builtin/rebase--helper.c
```diff
@@ -73,18 +79,59 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
...
+	if (command == MAKE_SCRIPT && !upstream && squash_onto) {
+		FILE *squash_onto_file;
+		int ret;
+
+		squash_onto_file = fopen(path_squash_onto(), "w");
+		if (squash_onto_file == NULL)
+			return error_errno(_("could not open '%s'"), path_squash_onto());
+
+		ret = fputs(squash_onto, squash_onto_file);
+		fclose(squash_onto_file);
```
dscho: This loses the newline (not sure it matters, but it is a change in behavior, and I don't think we have to). Also: Don't we already have helpers for this?

*clicketyclick*

`write_file(path, fmt, ...)`. So the code would be this short:

```c
if (command == MAKE_SCRIPT && !upstream && squash_onto &&
    write_file(path_squash_onto(), "%s\n", squash_onto) < 0)
        return error_errno(_("could not write '%s'"), path_squash_onto());
```

builtin/rebase--helper.c
```diff
@@ -73,18 +79,59 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
...
+			get_oid(upstream, &upstream_oid);
+			shortupstream = find_unique_abbrev(&upstream_oid, DEFAULT_ABBREV);
+			strbuf_addf(&revisions, "%s...%s", upstream, head_hash);
+			strbuf_addf(&shortrevisions, "%s..%s", shortupstream, shorthead);
+		} else {
+			strbuf_addf(&revisions, "%s...%s", onto, head_hash);
+			strbuf_add(&shortrevisions, shorthead, strlen(shorthead));
+		}
+	}
+
```
dscho: In my opinion, it would make the code more readable to have this refactored out into its own function, which is called precisely before it is needed, i.e. directly before the calls to `sequencer_*()`.

This would also make it easier to take custody of the memory. And *then* `xstrfmt()` would make a ton of sense. Something like

```c
static int get_revision_ranges(const char *upstream, const char *onto, char **head_hash, char **revisions, char **shortrevisions)
{
...
}
```

Then, you would call that function in the below code, with variables local to that scope of the conditional block, assign the return value of the call into the `sequencer_*()` function to a local variable, release the memory, return the value, clap your hands, done.

builtin/rebase--helper.c
```diff
@@ -73,18 +79,59 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
...
 	if (command == CONTINUE && argc == 1)
 		return !!sequencer_continue(&opts);
 	if (command == ABORT && argc == 1)
 		return !!sequencer_remove_state(&opts);
-	if (command == MAKE_SCRIPT && argc > 1)
-		return !!sequencer_make_script(stdout, argc, argv, flags);
+	if (command == MAKE_SCRIPT && (argc == 1 || argc == 2)) {
```
dscho: Why restrict this to almost the opposite of `argc > 1`?

builtin/rebase--helper.c
```diff
@@ -73,18 +79,59 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
...
+		const char *restrict_rev = (argc == 2) ? argv[1] : NULL;
+		const char *f_argv[3] = {argv[0], revisions.buf, restrict_rev};
```
dscho: That is not really readable. `f_argv`? Shouldn't the `argv` list be `NULL` terminated? I have to admit that it took me a while to wrap my head around this...

Also, unless my memory went bad, we have to add the `restrict_revision` (why change the name? That's unnecessary) with a prefixed `^` to make it negative.

You can avoid all that by using an `argv_array`. Please do so.

builtin/rebase--helper.c
```diff
@@ -97,9 +144,9 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 		return !!edit_todo_list(flags);
 	if (command == PREPARE_BRANCH && argc == 2)
 		return !!prepare_branch_to_be_rebased(&opts, argv[1]);
-	if (command == COMPLETE_ACTION && argc == 6)
-		return !!complete_action(&opts, flags, argv[1], argv[2], argv[3],
-					 argv[4], argv[5], autosquash, keep_empty);
+	if (command == COMPLETE_ACTION && argc == 3)
+		return !!complete_action(&opts, flags, shortrevisions.buf, argv[1], onto,
+					 head_hash, argv[2], autosquash, keep_empty);
```
dscho: So the short revisions are only needed here? The we also should only generate it here. You can bring this about by teaching aforementioned function to handle `shortrevisions != NULL` and only doing the "short dance" in that case.

git-rebase--interactive.sh
```diff
@@ -87,16 +70,16 @@ git_rebase__interactive () {
...
 
-	init_revisions_and_shortrevisions
-
 	git rebase--helper --make-script ${keep_empty:+--keep-empty} \
 		${rebase_merges:+--rebase-merges} \
 		${rebase_cousins:+--rebase-cousins} \
-		$revisions ${restrict_revision+^$restrict_revision} >"$todo" ||
+		${upstream:+--upstream "$upstream"} ${onto:+--onto "$onto"} \
+		${squash_onto:+--squash-onto "$squash_onto"} \
+		${restrict_revision+^$restrict_revision} >"$todo" ||
```
dscho: Oh, so that is the piece that I was missing. You changed the signature of the `--make-script` call!

Maybe it would make sense to split this commit into two, even, to first implement the part that is needed by `--make-script`, and the second commit that takes care of the needs of `--complete-action` (most notably, the `shortrevisions` part).

Would make it easier to review, methinks.

dscho:
@agrn excellent progress!

builtin/rebase--helper.c
agrn: write_file() does not return an error code. I’m changing this accordingly.

dscho: Yes, you're right. @peff changed this a while ago. In the exact wrong direction. ☹️

peff: To be fair, the return code was a red herring; it would always die anyway. You can be mad at me for removing the `_gently` form, though.

builtin/rebase--helper.c
agrn: I’m also dissatisfied with `f_argv`, do you have a better name to suggest?

agrn:
I updated my branch

builtin/rebase--helper.c
dscho: I am not mad at you. I am sad that some of us worked so hard to libify Git, and that this change (undoing some of that effort) was so happily applied.

Maybe in the future, your aims are more aligned with mine, now that you see where it would have been very nice indeed to already have a properly "gentle" function.

This `die()`ing left and right is just no good. It is a really sloppy way to develop code.

peff: I'm not sure I disagree with the original removal. The gentle version never had any callers and after 4 years was removed. I'm totally happy for you to re-add it if you have an actual caller, but I have no regrets about removing dead code.

builtin/rebase--helper.c
dscho: Maybe `rev_list_args` or `rev_args`.

builtin/rebase--helper.c
```diff
@@ -83,8 +111,28 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
...
+		ret = get_revision_ranges(upstream, onto, &head_hash, &revisions);
+		if (ret)
+			return ret;
+
+		argv_array_push(&f_argv, argv[0]);
+		argv_array_push(&f_argv, revisions);
+		if (argc == 2)
+			argv_array_push(&f_argv, argv[1]);
+
+		ret = !!sequencer_make_script(stdout, f_argv.argc, f_argv.argv, flags);
```
dscho: Okay, I made up my mind: `make_script_args` is my favorite alternative name for `f_argv`.

builtin/rebase--helper.c
```diff
@@ -83,8 +111,28 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
...
+		struct argv_array f_argv = ARGV_ARRAY_INIT;
+
+		if (!upstream && squash_onto)
+			write_file(path_squash_onto(), "%s\n", squash_onto);
+
+		ret = get_revision_ranges(upstream, onto, &head_hash, &revisions);
+		if (ret)
+			return ret;
+
+		argv_array_push(&f_argv, argv[0]);
```
dscho: What is `argv[0]` at this point? Would it not make most sense to name it in a local variable?

builtin/rebase--helper.c
```diff
@@ -83,8 +111,28 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
...
+		argv_array_push(&f_argv, revisions);
+		if (argc == 2)
+			argv_array_push(&f_argv, argv[1]);
```
dscho: `restrict_revision` is the only thing that is not parsed via parse options. And that is the reason why the code here talks about `argv[1]` which is not as obvious as `restrict_revision` would be...

dscho:
Thank you! It might be just me, but I have to admit that this split into the two commits made this a much more pleasant read to me, as everything became a lot more obvious.

builtin/rebase--helper.c
dscho: In light of all the code churn caused by having to libify everything, carefully combing through code to avoid those pesky `exit()` and `die()` calls (doing this in the `deprecate-grafts` patch series took me well over three times as long as to come up with the original patch series!!!), I think you will agree that a better idea is to keep the gentle version, and to convert the non-gentle callers (because they should most likely have been gentle in the first place, that's the problem your review uncovered, not that the gentle version wasn't necessary: it was, and still is, necessary.)

builtin/rebase--helper.c
```diff
@@ -10,22 +10,36 @@ static GIT_PATH_FUNC(path_squash_onto, "rebase-merge/squash-onto")
...
+			*revisions = xstrfmt("%s...%s", upstream, *head_hash);
+		if (shortrevisions) {
+			shortupstream = find_unique_abbrev(&upstream_oid, DEFAULT_ABBREV);
+			*shortrevisions = xstrfmt("%s..%s", shortupstream, shorthead);
+		}
+	} else {
+		if (revisions)
+			*revisions = xstrfmt("%s...%s", onto, *head_hash);
+		if (shortrevisions)
+			*shortrevisions = xstrndup(shorthead, strlen(shorthead));
```
chriscool: Why not just `*shortrevisions = xstrdup(shorthead)`?

builtin/rebase--helper.c
```diff
@@ -10,22 +10,36 @@ static GIT_PATH_FUNC(path_squash_onto, "rebase-merge/squash-onto")
...
 
 		get_oid(upstream, &upstream_oid);
-		*revisions = xstrfmt("%s...%s", upstream, *head_hash);
-	} else
-		*revisions = xstrfmt("%s...%s", onto, *head_hash);
+
+		if (revisions)
+			*revisions = xstrfmt("%s...%s", upstream, *head_hash);
+		if (shortrevisions) {
+			shortupstream = find_unique_abbrev(&upstream_oid, DEFAULT_ABBREV);
```
chriscool: shortupstream is used only inside the current if clause, so it could be declared here too.

builtin/rebase--helper.c
dscho: Wait, is that even pointing out the right issue? Since `revisions` is essentially `$onto..$head_hash`, `shortrevisions` should be `$short_onto..$short_head`.

/me goes and looks

Hmpf. The code agrees: https://github.com/git/git/blob/v2.18.0/git-rebase--interactive.sh#L898

I think this is a bug in `git-rebase--interactive.sh`...

builtin/rebase--helper.c
dscho: > I think that converting some die() calls into error() calls when no other libifying effort is made could be seen as unnecessary code churn and rejected.

Are you trying to excuse yourself from the task? I remember that the builtin `am` suffered from many a `die()`, and I would not be surprised if it still did.

I know it is hard work, and I understand that you don't want to do it if you do not even see the value of it.

Obviously, it would be much easier to stay sloppy and just call `die()`. And leave the user puzzled what happened here, as the code that calls `die()` is unlikely to have the bigger picture of what the user tried to do.

For example, if your builtin `am` runs as part of a rebase, a regular Git user who is not familiar with Git's code base would probably go "ummm" when seeing something about `am` written in an error message: "What is this 'am' thing? I called a *rebase* for crying out loud!"

So I understand that it is so much easier from the developer's point of view to just forgo proper error handling. Quick and dirty. And make a mess of the user experience.

I encourage you to try to understand this reasoning, so that your own code can become better at this.

dscho: > It is also very hard to predict a long time in advance if and when we will need to libify a piece of code.

I forgot to point out how ludicrous this statements premise is: you seem to think that it is okay to be sloppy by calling `die()`, as long as you think the code is not deep down in the call chain. Forgetting that every useful piece of code will be reused sooner rather than later. So unless you think that your code is not useful, you should write it in a way that it can be reused easily.

builtin/rebase--helper.c
agrn: argv[0] contains "rebase--helper". I think a comment would be more appropriate than a variable.

builtin/rebase--helper.c
agrn: Same as above, I think a comment would be more appropriate than a variable.

builtin/rebase--helper.c
agrn: As far as I can see, `$shortrevisions` was introduced with d911d1465d ([see here](https://github.com/git/git/commit/d911d1465d134cbd126e531a544f33b13f4e855a#diff-0f15aff45d5dd346465c35597a5f274eR659)), and it was already like that here. Changing its value to `$short_onto..$short_head` does not seem to break any test, though.

builtin/rebase--helper.c
dscho: No. While it is true that code cannot be self-documenting all the time, this is a prime example where it *can*. And experience tells us quite loudly that code comments are *prone* to become stale. So a code comment here would be less appropriate than a variable that documents clearly what this does.

Besides, I have a hunch that this `argv[0]` is ignored by `make_script()`, and therefore maybe the true fix is to avoid it altogether (read: change both caller and callee *not* to require a bogus `argv[0]` in the first place).

builtin/rebase--helper.c
dscho: Same as above, you are wrong here. A code comment is *always* more prone to become stale than real code, as the code comment is ignored by the compiler, and the code simply is not.

Plus, I have no idea why you insist on passing the `restrict_revision` as a non-option. It would make everything so much clearer, and solve the current problem, if you simply did it via `--restrict-revision=...` instead of playing games with positional parameters.

builtin/rebase--helper.c
dscho: It should be a strong hint that *I myself*, during a review, stumbled over this. So please do expect the same question to come up on the Git mailing list. And the best way to prepare for this is to either dig into history to find out whether there is any good reason for this, or whether this is a bug. In the latter case, fix it. In the former, document it in the commit message. That's what commit messages are for, after all, to answer important question about the patch such as this one, that are not readily answered by merely looking at the diff.

agrn: [Well, you tried to change this a while ago](https://public-inbox.org/git/alpine.DEB.1.00.0901260031460.14855@racer/). Apart from that, I did not really found something interesting on the list.

Also, I wanted to point out that this specific code is executed when the user call a `rebase --root` (empty upstream means `rebase --root`), [and that `$shortonto` is only used in the help message](https://github.com/agrn/git/blob/ag/init-revisions-c/sequencer.c#L4512-L4515). So, if I make a `rebase -i --root`, I have something like that:

```
pick f203597 First commit
pick 7855fc1 changes
…
pick 5488a93 more changes
pick fb6bb71 last commit

# Rebase fb6bb71 onto c9e8c18 (447 commands)
#        ^            ^~~~~~~ $shortonto
#        ~~~~~~~ $shortrevisions
```

I find this a bit unclear.

dscho: "a while ago"... *snickers*... 9.5 years ago. I did not remember, I must confess.

But now I understand why `shortrevisions` is defined that way: in the `--root` case, there is no `..` way to specify the commit range, sadly. The `$onto..$head`, as `revisions` does it, is actually incorrect (although it does not matter in practice, as nobody would use an `onto` that is reachable from `head` in the `--root` case: that would cause essentially a non-`--root` rebase).

Still, I feel uneasy about all of this discussion *not* being distilled into a small, clarifying paragraph in the commit message. The issue *is* confusing, after all.

agrn:
I updated the branch again.

This renames `f_argv`, uses an option to pass `restrict_revision`, drops `argv[0]`, and improves (in my opinion) `get_revision_ranges()`.

builtin/rebase--helper.c
```diff
@@ -38,10 +60,6 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 				ABORT),
 		OPT_CMDMODE(0, "make-script", &command,
 			N_("make rebase script"), MAKE_SCRIPT),
-		OPT_CMDMODE(0, "shorten-ids", &command,
-			N_("shorten commit ids in the todo list"), SHORTEN_OIDS),
-		OPT_CMDMODE(0, "expand-ids", &command,
-			N_("expand commit ids in the todo list"), EXPAND_OIDS),
```
dscho: I just noticed that this removal is not accompanied by any removal of shell script code that would call those subcommands... Should this not be squashed into an earlier commit, then?

dscho:
> I updated the branch again.

Thank you! I find the updated version very readable.

builtin/rebase--helper.c
agrn: Indeed, it should be part of the previous commit.

builtin/rebase--helper.c
```diff
@@ -37,7 +55,7 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 	int abbreviate_commands = 0, rebase_cousins = -1, ret;
 	const char *head_hash = NULL, *onto = NULL, *restrict_revisions = NULL,
 		*squash_onto = NULL, *upstream = NULL;
-	char *revisions = NULL;
+	char *revisions = NULL, *shortrevisions = NULL;
```
chriscool: It seems to me that the declarations of revisions and shortrevisions could be moved below inside the if clauses where these variables are used.

agrn:
I updated the branch

chriscool:
It looks great to me! Thanks!

agrn:
Hi, I just updated my branch to update skip_unnecessary_picks() (9ac8b9d), fixed some flaws Junio pointed out in complete_action() (``rebase -i`` would not abort if the todo list had only commented actions, a560b56), and added a test case for this (7a0068b1e).

Please tell me what you think about those changes!

builtin/rebase--helper.c
```diff
@@ -90,8 +90,13 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 		return !!transform_todos(flags);
 	if (command == CHECK_TODO_LIST && argc == 1)
 		return !!check_todo_list();
-	if (command == SKIP_UNNECESSARY_PICKS && argc == 1)
-		return !!skip_unnecessary_picks();
+	if (command == SKIP_UNNECESSARY_PICKS && argc == 1) {
+		struct object_id oid;
+		ret = skip_unnecessary_picks(&oid);
```
chriscool: Maybe `ret` could be declared here, as it looks like it is used only in this if clause.

sequencer.c
```diff
@@ -4471,6 +4475,101 @@ int skip_unnecessary_picks(struct object_id *output_oid)
...
+
+	if (parse_insn_buffer(buf->buf, &todo_list)) {
+		todo_list_release(&todo_list);
+		return error(_("unusable todo list: '%s'"), todo_file);
+	}
+
+	if (count_commands(&todo_list) == 0) {
+		apply_autostash(opts);
+		sequencer_remove_state(opts);
+		todo_list_release(&todo_list);
```
chriscool: Maybe we should also release `buf` here.

sequencer.c
```diff
@@ -4471,6 +4475,101 @@ int skip_unnecessary_picks(struct object_id *output_oid)
...
+		return 1;
+
+	if (cmd && *cmd)
+		sequencer_add_exec_commands(cmd);
+
+	if (strbuf_read_file(buf, todo_file, 0) < 0)
+		return error_errno(_("could not read '%s'."), todo_file);
+
+	if (parse_insn_buffer(buf->buf, &todo_list)) {
+		todo_list_release(&todo_list);
```
chriscool: Maybe we should release `buf` here.

agrn: No need to do so. ``buf`` is part of ``todo_list``, and is released by ``todo_list_release()``.

chriscool:
I am not sure about the memory management in case of errors in a560b56.

sequencer.c
chriscool: Ok, sorry about the noise.

builtin/rebase--helper.c
agrn: I fixed this in my new pull request.

dscho:
@agrn this has been integrated into `master` via b49ef560ed66, amirite?

PR #522, opened 2018-07-16
rebase--interactive: rewrite write_basic_state() and init_basic_state() in C

agrn:
This rewrites ``write_basic_state()`` and ``init_basic_state()`` in C. The relevant commits are the last two.

### DIFF
--- a/builtin/rebase--helper.c
+++ b/builtin/rebase--helper.c
@@ -97,7 +97,7 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 	if (command == ADD_EXEC && argc == 2)
 		return !!sequencer_add_exec_commands(argv[1]);
 	if (command == APPEND_TODO_HELP && argc == 1)
-		return !!append_todo_help(0, keep_empty);
+		return !!append_todo_help_to_file(0, keep_empty);
 	if (command == EDIT_TODO && argc == 1)
 		return !!edit_todo_list(flags);
 	if (command == PREPARE_BRANCH && argc == 2)
--- a/builtin/rebase--helper.c
+++ b/builtin/rebase--helper.c
@@ -13,12 +13,12 @@ static const char * const builtin_rebase_helper_usage[] = {
 int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
 {
 	struct replay_opts opts = REPLAY_OPTS_INIT;
-	unsigned flags = 0, keep_empty = 0, rebase_merges = 0, write_edit_todo = 0;
+	unsigned flags = 0, keep_empty = 0, rebase_merges = 0;
 	int abbreviate_commands = 0, rebase_cousins = -1;
 	enum {
 		CONTINUE = 1, ABORT, MAKE_SCRIPT, SHORTEN_OIDS, EXPAND_OIDS,
 		CHECK_TODO_LIST, SKIP_UNNECESSARY_PICKS, REARRANGE_SQUASH,
-		ADD_EXEC, APPEND_TODO_HELP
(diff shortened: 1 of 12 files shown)


builtin/rebase--helper.c
```diff
@@ -88,13 +92,23 @@ int cmd_rebase__helper(int argc, const char **argv, const char *prefix)
...
 			   N_("squash onto")),
 		OPT_STRING(0, "upstream", &upstream, N_("upstream"),
 			   N_("the upstream commit")),
+		OPT_STRING(0, "head-name", &head_name, N_("head-name"), N_("head name")),
+		OPT_STRING('S', "gpg-sign", &opts.gpg_sign, N_("gpg-sign"),
+			   N_("GPG-sign commits")),
+		OPT_STRING(0, "strategy", &opts.strategy, N_("strategy"),
+			   N_("rebase strategy")),
+		OPT_STRING(0, "strategy-opts", &raw_strategies, N_("strategy-opts"),
+			   N_("strategy option")),
```
chriscool: Shouldn't this be "strategy options" (with an "s" at the end of "options"?

builtin/rebase--helper.c
```diff
@@ -44,6 +47,25 @@ static int get_revision_ranges(const char *upstream, const char *onto,
...
 }
 
+static int init_basic_state(struct replay_opts *opts, const char *head_name,
+			    const char *onto, const char *orig_head)
+{
+	struct stat sb;
+	FILE *interactive;
+
+	if (stat(path_state_dir(), &sb) && mkdir_in_gitdir(path_state_dir()))
+		return error(_("could not create temporary %s"), path_state_dir());
```
chriscool: Maybe using error_errno() would be better as I am not sure that mkdir_in_gitdir() would output an error message.

builtin/rebase--helper.c
```diff
@@ -44,6 +47,25 @@ static int get_revision_ranges(const char *upstream, const char *onto,
...
+
+	delete_reflog("REBASE_HEAD");
+
+	interactive = fopen(path_interactive(), "w");
+	if (!interactive)
+		return error(_("could not mark as interactive"));
```
chriscool: Here also maybe use error_errno().

agrn:
I also replaced “rebase--interactive” by “rebase -i” in the commit messages.

sequencer.h
```diff
@@ -116,3 +116,7 @@ int prepare_branch_to_be_rebased(struct replay_opts *opts, const char *commit);
 void print_commit_summary(const char *prefix, const struct object_id *oid,
 			  unsigned int flags);
 #endif
+
+void parse_strategy_opts(struct replay_opts *opts, char *raw_opts);
+int write_basic_state(struct replay_opts *opts, const char *head_name,
+		      const char *onto, const char *orig_head);
```
chriscool: I wonder if write_basic_state() really needs to be public in sequencer.{h,c}. Do you have plan to reuse it?

chriscool:
Except for my last comment, this looks good to me! Thanks!

sequencer.h
agrn: No, I did it to have access to ``rebase_path_*()``.

chriscool: Ok, then.

agrn: Perhaps it would be better to make those we need public, like I did with the todo-list file?

chriscool: Yeah, that's another possible solution. Maybe you can just send everything as it is and wait for possible comments about this.

dscho:
@agrn this PR has been integrated via b49ef560ed66, correct?

PR #524, opened 2018-07-23
rebase--interactive: turn git-rebase--interactive into a builtin

agrn:
This turns git-rebase--interactive into a builtin. The revelant commits are the last four.


builtin/rebase--interactive2.c
```diff
@@ -70,12 +70,13 @@ int cmd_rebase__interactive(int argc, const char **argv, const char *prefix)
...
+	const char *onto = NULL, *restrict_revision = NULL,
 		*squash_onto = NULL, *upstream = NULL, *head_name = NULL,
 		*switch_to = NULL, *cmd = NULL, *onto_name = NULL;
-	char *raw_strategies = NULL, *revisions = NULL, *shortrevisions = NULL;
-	struct argv_array make_script_args = ARGV_ARRAY_INIT;
-	FILE *todo_list;
+	char *raw_strategies = NULL;
+	enum {
+		CONTINUE = 1, SKIP, EDIT_TODO, SHOW_CURRENT_PATCH
+	} command = 0;
```
chriscool: Maybe the enum could also have NONE=0, like:

+	enum {
+		NONE=0, CONTINUE, SKIP, EDIT_TODO, SHOW_CURRENT_PATCH
+	} command = 0;

builtin/rebase--interactive2.c
```diff
@@ -131,57 +139,79 @@ int cmd_rebase__interactive(int argc, const char **argv, const char *prefix)
...
+		ret = sequencer_continue(&opts);
+	} else if (command == EDIT_TODO)
+		ret = edit_todo_list(flags);
+	else if (command == SHOW_CURRENT_PATCH) {
+		struct child_process cmd = CHILD_PROCESS_INIT;
+
+		cmd.git_cmd = 1;
+		argv_array_pushl(&cmd.args, "show", "REBASE_HEAD", "--", NULL);
+		ret = run_command(&cmd);
+	} else {
```
chriscool: This else clause is used when `command == 0` right? I think it would be clearer to have a NONE enum value as I suggested above and to have `else if (command == NONE)` here.

Actually maybe the whole `if (...) else if (...) ...` could be replaced by a `switch (command) { case NONE: ... case CONTINUE: ... }`.

builtin/rebase--interactive2.c
```diff
@@ -131,57 +139,79 @@ int cmd_rebase__interactive(int argc, const char **argv, const char *prefix)
...
+		FILE *todo_list;
+
+		ret = prepare_branch_to_be_rebased(&opts, switch_to);
+		if (ret)
+			return !!ret;
+
+		ret = get_revision_ranges(upstream, onto, &head_hash,
+					  &revisions, &shortrevisions);
+		if (ret)
+			return !!ret;
```
chriscool: When this returns, the returned value is always 1, right? In this case maybe it is clearer to have `return 1` instead of `return !!ret`.

chriscool:
Maybe the strategy to get to a git-rebase--interactive in C could be described a little better in the commits messages of the last 4 commits.

It's not easy to understand why things are done like this.

chriscool:
Usually the git-$command--helper.c file is renamed into git-$command.c at the end of a rewrite in C series but that's not what happens here, so a bit more explanation would be welcome.

agrn:
I just updated my branch. Aside from code and commit message changes, I moved rebase--helper’s remaining modes to rebase--interactive.c.

Also, I’m not sure if the switch/case is properly formatted.

agrn:
This should be better

builtin/rebase--interactive2.c
```diff
@@ -131,57 +139,83 @@ int cmd_rebase__interactive(int argc, const char **argv, const char *prefix)
...
 
-	ret = get_revision_ranges(upstream, onto, &head_hash,
-				  &revisions, &shortrevisions);
-	if (ret)
-		return !!ret;
+		rerere_clear(&merge_rr);
+	case CONTINUE:
+		ret = sequencer_continue(&opts);
+		break;
+	}
```
chriscool: I think this brace should be after the call to rerere_clear() above.

chriscool: Also maybe a `break;` is missing after the call to rerere_clear(), otherwise maybe a `/* fall through */`comment should be added after the call to rerere_clear().

builtin/rebase--interactive.c
```diff
@@ -160,6 +172,15 @@ int cmd_rebase__interactive(int argc, const char **argv, const char *prefix)
...
 	}
+	case SHORTEN_OIDS:
+	case EXPAND_OIDS:
+		return !!transform_todos(flags);
+	case CHECK_TODO_LIST:
+		return !!check_todo_list();
+	case REARRANGE_SQUASH:
+		return !!rearrange_squash();
+	case ADD_EXEC:
+		return !!sequencer_add_exec_commands(cmd);
```
chriscool: In the EDIT_TODO case above there is:

```
ret = edit_todo_list(flags);
break;
```

instead of:

```
```

so it feels that the code could be more consistent by doing things in the same way everywhere.

agrn:
Updated

builtin/rebase--interactive2.c
chriscool: I am not sure about the style. Let's leave it like this and see what people on the list say.

chriscool:
This looks great to me now!

dscho:
@agrn this has been integrated via b49ef560ed66, right?

PR #526, opened 2018-08-01
sequencer: refactor functions that manipulate the todo-list

agrn:
This refactors most functions that manipulate the todo-list not to read it from the filesystem. The relevant commits are the last nine.

I’m not sure that the new functions have a good name…

### DIFF
--- a/sequencer.c
+++ b/sequencer.c
@@ -55,8 +55,7 @@ static GIT_PATH_FUNC(rebase_path, "rebase-merge")
  * file and written to the tail of 'done'.
  */
 GIT_PATH_FUNC(rebase_path_todo, "rebase-merge/git-rebase-todo")
-static GIT_PATH_FUNC(rebase_path_todo_backup,
-		     "rebase-merge/git-rebase-todo.backup")
+GIT_PATH_FUNC(rebase_path_todo_backup, "rebase-merge/git-rebase-todo.backup")
 
 /*
  * The rebase command lines that have already been processed. A line
--- a/sequencer.c
+++ b/sequencer.c
@@ -4505,61 +4505,95 @@ int sequencer_make_script(struct repository *r, FILE *out,
  * Add commands after pick and (series of) squash/fixup commands
  * in the todo list.
  */
-int sequencer_add_exec_commands(struct repository *r,
-				const char *commands)
+static void todo_list_add_exec_commands(struct todo_list *todo_list,
+					struct string_list *commands)
 {
-	const char *todo_file = rebase_path_todo();
-	struct todo_list todo_list = TODO_LIST_INIT;
-	struct strbuf *buf = &todo_list.buf;
-	size_t offset = 0, commands_len = strlen(commands);
-	int i, insert;
+	struct strbuf *buf = &todo_list->buf;
+	size_t base_offset = buf->len;
+	int i, insert, nr = 0, alloc = 0;
+	struct todo_item *items = NULL, *base_items = NULL;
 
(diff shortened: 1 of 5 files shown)


sequencer.c
```diff
@@ -4314,111 +4314,26 @@ int transform_todos(unsigned flags)
...
-}
-
-define_commit_slab(commit_seen, unsigned char);
-/*
- * Check if the user dropped some commits by mistake
- * Behaviour determined by rebase.missingCommitsCheck.
- * Check if there is an unrecognized command or a
- * bad SHA-1 in a command.
- */
 int check_todo_list(void)
```
chriscool: Maybe renaming check_todo_list() to check_todo_list_from file() would make things clearer.

sequencer.c
```diff
@@ -4314,111 +4314,26 @@ int transform_todos(unsigned flags)
...
 {
-	enum missing_commit_check_level check_level = get_missing_commit_check_level();
-	struct strbuf todo_file = STRBUF_INIT;
-	struct todo_list todo_list = TODO_LIST_INIT;
-	struct strbuf missing = STRBUF_INIT;
-	int advise_to_edit_todo = 0, res = 0, i;
-	struct commit_seen commit_seen;
-
-	init_commit_seen(&commit_seen);
+	struct todo_list old = TODO_LIST_INIT, new = TODO_LIST_INIT;
```
chriscool: Nit: I would call the variable `old_todo` and `new_todo` instead of just `old` and `new`.

sequencer.c
```diff
@@ -4710,18 +4700,42 @@ int rearrange_squash(void)
...
 
 	clear_commit_todo_item(&commit_todo);
+
+	if (todo_list_parse_insn_buffer(todo_list->buf.buf, todo_list))
+		BUG("unusable todo list");
+
+	return 0;
+}
+
+int rearrange_squash(void)
```
chriscool: Maybe this could be called rearrange_squash_in_todo_file() or something like that.

chriscool:
There are a few nits I noticed especially about naming functions that operate on files. Otherwise it looks good to me.

sequencer.c
```diff
@@ -4324,7 +4314,31 @@ int transform_todos(unsigned flags)
...
+}
+
+int transform_todos(unsigned flags)
+{
+	const char *todo_file = rebase_path_todo();
+	struct todo_list todo_list = TODO_LIST_INIT;
+	int i;
+
+	if (strbuf_read_file(&todo_list.buf, todo_file, 0) < 0)
+		return error(_("could not read '%s'."), todo_file);
```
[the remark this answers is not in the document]
agrn: I don’t understand why you think this is not related to the current patch, this code was part of the original ``transform_todos()``.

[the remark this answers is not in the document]
chriscool: I mean that the current patch doesn't introduce the error) call, so it is a separate issue (from the issue the current patch addresses) to change error() into error_errno(), which means that changing error() into error_errno() should be done in another patch.

agrn:
I updated my branch

dscho:
For the record: the CI tests failed yesterday because of that [flaky t5570-osx problem that is already fixed in `pu`](https://github.com/git/git/compare/master...gitgitgadget:jk/no-sigpipe-during-network-transport). I re-ran the CI test and it passed (as flaky tests usually do when run again).

dscho:
@agrn this has been integrated via e62e68d35949, right?

agrn:
Yes.  I’m closing this.

PR #547, opened 2018-10-27
DiffHighlight.pm: Use correct /dev/null for UNIX and Windows

webstech:
Use File::Spec->devnull() for output redirection to avoid messages when Windows version of Perl is first in path.  The message 'The system cannot find the path specified.' is displayed each time git is run to get colors.


dscho:
@webstech since this has made it to `master` (see https://github.com/git/git/commit/6804ba3a58ff2999533af28ca5af3901e30fd1f2), maybe you want to close this here PR?

PR #550, opened 2018-11-05
Fix utf 16 working tree encoding

AdRiAnIlloO:
Currently:

*	Using UTF-16 or UTF-32 value on `working-tree-encoding` causes files with that encoding which also have BOM to later be assigned a BE BOM (0xFEFF) once Git converts them out from UTF-8 as stored internally, thus breaking initial LE BOM (0xFFFE) for projects which require it.

*	Using explicit BE/LE BOM instead in `working-tree-encoding`, is prohibited when the designated files start with a BOM (only for the declared UTF-16/UTF-32 encoding), disabling human-readable diffs (simply informing of "binary type"), and completely blocking commit operations, along with an error message. This behaviour happens as Git respects this part of Unicode's standard: http://unicode.org/faq/utf_bom.html#bom10

In this PR:

*	Git only prohibites the opposite BOM than the one in `working-tree-encoding` (e.g. if declared LE, then it denies BE BOM presence within the associated file, of the declared UTF-16/UTF-32). This way the user can now make Git operations which were previously impossible, with the only requisite being to match the endianness of `working-tree-encoding` attribute with the associated file/s. Furthermore, Git won't change the endianness to BE default after internal operations take place. The only downside is this violates the Unicode standard referenced earlier (in the equal BOM cases).

*	Note that the behaviour when user sets `working-tree-encoding` value to exactly "UTF-16" or "UTF-32" (so without endianness) is unaffected. Git continues to require BOM presence in that case.

Testing can be done with the script that first commit adds in `./t` directory. The comments there should be relatively easy to understand. There is only one test case that should pass after having the second commit, but should fail if discarding that commit. I added the test script in a commit different than the one with the fixes for that reason (to be easily comparable).

### DIFF
--- a/t/t0028-working-tree-encoding.sh
+++ b/t/t0028-working-tree-encoding.sh
@@ -11,7 +11,10 @@ test_expect_success 'setup test files' '
 
 	text="hallo there!\ncan you read me?" &&
 	echo "*.utf16 text working-tree-encoding=utf-16" >.gitattributes &&
+	echo "*bebom.utf16be text working-tree-encoding=utf-16be" >>.gitattributes &&
+	echo "*lebom.utf16le text working-tree-encoding=utf-16le" >>.gitattributes &&
 	printf "$text" >test.utf8.raw &&
+	printf "\xEF\xBB\xBF$text" >test.bom.utf8.raw && # To deal w/BOM files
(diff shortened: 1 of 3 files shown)


dscho:
@AdRiAnIlloO it seems that your PR has stalled a little... are you still interested in this?

PR #551, opened 2018-11-08
Add info that Windows supports in CI

joaothallis:
Is written that Travis CI is not supported for Windows.

A version with support for Windows has already been released: https://blog.travis-ci.com/2018-10-11-windows-early-release.


joaothallis:
@sea51930 what I meant?

PR #558, opened 2018-11-21
pack-redundant: optimize for memory and cpu time

sunchao9:
When call pack-redundant witch "--all", if there are too much
local packs and too much redundant objects within them, the
**minimize** step may eat up the memory and cpu because it
would produces a huge-big-size "pll" list in the heap.

Change the strategy of "cmp_two_packs" could solve this: At
first sort all the local packs from smallest to biggest. Then
if two local packs have the same objects, do not remove it from
"unique_objects" in both packs' "pack_list", but just the one
which has bigger size.

With the strategy for "cmp_two_packs" above, the unique set in
the local packs is the collection with smallest size. And
**minimize** will need not to get the permutations for the missing
objects from the unique set, which need less memory and cpu time.

Signed-off-by: sunchao <sunchao9@huawei.com>

### DIFF
--- a/builtin/pack-redundant.c
+++ b/builtin/pack-redundant.c
@@ -247,11 +247,14 @@ static struct pack_list * pack_list_difference(const struct pack_list *A,
 	return ret;
 }
 
+// because smaller packs are in the left of the list, we could just
+// remove the redundant objects in the right pack, after cmp finished,
+// the left packs whose unique_objects size is zero is redundant packs.
 static void cmp_two_packs(struct pack_list *p1, struct pack_list *p2)
 {
 	unsigned long p1_off = 0, p2_off = 0, p1_step, p2_step;
 	const unsigned char *p1_base, *p2_base;
-	struct llist_item *p1_hint = NULL, *p2_hint = NULL;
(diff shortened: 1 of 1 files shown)


sunchao9:
We could reproduct the case that `git pack-redundant --all` will eat up cpu-time by this script:

the following script create the repository which contains lots of redundant packs:

```
#!/bin/bash

git init --bare base
pushd base
git config --local gc.auto 0
popd
git clone base 1

for ((i=0; i<1000; i++)) {
    pushd 1 >/dev/null
    date +%s%N >> $i
    git add .
    git commit -sm "$i"
    git push origin HEAD:master
    popd >/dev/null

    pushd base/objects/pack >/dev/null
    git repack

    if test "x$(($i%2))" = "x0"; then
        git repack -a
        name=`ls -lt  | head -n 2 | tail -n 1 | cut -d' ' -f 9`
        touch ${name%.idx}.keep
    fi
    popd >/dev/null
}
```

at last, we could run `git pack-redundant --all` in `base`, and we could see that the command will exit unnormally "killed".

PR #561, opened 2018-12-01
Adds documentation for project configuration file

Sanjo:
I had the problem that I wanted to distribute some Git configurations for the repository.

I have documented the solution that I have found by trying out some things here:
https://medium.com/@Sanjo/distributing-the-git-config-of-a-repository-with-all-users-of-the-repository-99f40350afa4

I have also found out that it isn't explicitly documented that you can have a project scoped .gitconfig that can be committed. I have only seen that it is documented that you can have a global and a OS user scoped .gitconfig. See https://git-scm.com/docs/git-config#FILES.

This is my proposal for documenting it. But I have noticed that there is [this](https://github.com/git/git/blob/7068cbc4abac53d9c3675dfba81c1e97d25e8eeb/Documentation/git-config.txt#L291-L293) which somehow is related but I haven't understood it completely what it should mean.

### DIFF
--- a/Documentation/git-config.txt
+++ b/Documentation/git-config.txt
@@ -291,6 +291,9 @@ $GIT_DIR/config::
 $GIT_DIR/config.worktree::
 	This is optional and is only searched when
 	`extensions.worktreeConfig` is present in $GIT_DIR/config.
+	
+.gitconfig::
+	Project-specific configuration file.	
 
 If no further options are given, all reading options will read all of these
 files that are available. If the global or the system-wide configuration


dscho:
Given that it would pose a serious security risk to allow a tracked `.gitconfig` to be used by Git, I am not sure what this PR tries to get at. Are you saying that this works? Because it doesn't, and it shouldn't.

(Apart from that, sending people elsewhere to piece together information bit by bit is not going to go over well. Instead of that Medium link, or at least *before* that link, you should summarize the idea.)

Sanjo:
Sorry, there was some confusion on my side. I already had deleted the Medium article. Closing this PR.

PR #563, opened 2018-12-08
submodule: Add --reset-hard option for git submodule update

yarikoptic:
This patch adds a --reset-hard option for the update command to hard
reset submodule(s) to the gitlink for the corresponding submodule in
the superproject.  This feature is desired e.g. to be able to discard
recent changes in the entire hierarchy of the submodules after running

   git reset --hard PREVIOUS_STATE

in the superproject which leaves submodules in their original state,
and

   git reset --hard --recurse-submodules PREVIOUS_STATE

would result in the submodules being checked into detached HEADs.

As in the original  git reset --hard  no checks or any kind of
safe-guards against jumping into some state which was never on the
current branch is done.

must_die_on_failure is not set to  yes to mimic behavior of a update
--checkout strategy, which would leave user with a non-clean state
immediately apparent via  git status  so an informed decision/actions
could be done manually.

Signed-off-by: Yaroslav Halchenko <debian@onerussian.com>

I will send it now also to the list for the review


robbyoconnor:
> Git Source Code Mirror - This is a publish-only repository and all pull requests are ignored. Please follow Documentation/SubmittingPatches procedure for any of your improvements.

robbyoconnor:
> Git Source Code Mirror - This is a publish-only repository and **all pull requests are ignored. Please follow Documentation/SubmittingPatches procedure for any of your improvements.**

Emphasis added.

Documentation/git-submodule.txt
```diff
@@ -358,6 +358,16 @@ the submodule itself.
 	If the key `submodule.$name.update` is set to `rebase`, this option is
 	implicit.
 
+--reset-hard::
+	This option is only valid for the update command.
+	Hard reset current state to the commit recorded in the	superproject.
+    If this option is given, the submodule's HEAD will not get detached
```
stefanbeller: Are you mixing tabs and spaces for indentation here?

If you want to make this a new paragraph in the man page, you can separate the new paragraph with a '+' and then not indenting it at all. (See the --reference NOTE below as that uses the same style)

Documentation/git-submodule.txt
```diff
@@ -358,6 +358,16 @@ the submodule itself.
...
+    if it was not detached before. Note that, like with a regular
+    git reset --hard  no safe-guards are in place to prevent jumping
```
stefanbeller: Could you quote `git reset --hard` such that it is highlighted? (Also double space after that command?)

Documentation/git-submodule.txt
yarikoptic: emacs played a trick on me. thanks for spotting, fixed now

Documentation/git-submodule.txt
yarikoptic: thanks, fixed

Also added to this PR an additional improvement (IMHO) to the tests, will follow up on the mailing list now

robbyoconnor:
This is unlikely to ever be merged -- look at the README ;)

yarikoptic:
@robbyoconnor  I am just using this PR for testing and possible review/comments... patches are sent to the list

dscho:
@yarikoptic I suppose this has moved to https://public-inbox.org/git/20181211040839.17472-1-debian@onerussian.com/#r? If so, is it possible that this project has stalled a little?

yarikoptic:
yes, it has moved to the list and yes it has stalled -- I wanted to beautify the tests and never got back to it.  I hope to do so in a few weeks.  I will close here
Cheers!

PR #574, opened 2019-02-10
Fix: Check flag "--ignore-errors" on lstat() call

qusielle:
"git add --ignore-errors" command fails immediately when lstat return error (for example, when it can't read the file), despite the ignore errors' flag is enabled.

### DIFF
--- a/read-cache.c
+++ b/read-cache.c
@@ -791,7 +791,7 @@ int add_to_index(struct index_state *istate, const char *path, struct stat *st, 
 int add_file_to_index(struct index_state *istate, const char *path, int flags)
 {
 	struct stat st;
-	if (lstat(path, &st))
+	if (lstat(path, &st) && !(flags & ADD_CACHE_IGNORE_ERRORS))
 		die_errno(_("unable to stat '%s'"), path);
 	return add_to_index(istate, path, &st, flags);
 }


read-cache.c
```diff
@@ -791,7 +791,7 @@ int add_to_index(struct index_state *istate, const char *path, struct stat *st,
 int add_file_to_index(struct index_state *istate, const char *path, int flags)
 {
 	struct stat st;
-	if (lstat(path, &st))
+	if (lstat(path, &st) && !(flags & ADD_CACHE_IGNORE_ERRORS))
```
dscho: Is this really a recoverable error? I would claim that it is an *unrecoverable* error, as a failed `lstat()` call leaves `st` uninitialized, yet it is used two lines further down. That would introduce a major bug.

dscho: @qusielle Ping!

dscho:
@qusielle gentle ping?

qusielle:
@dscho 
I'm really sorry to keep you waiting so long...
There could be files that triggers an error on stat(), when other files proceed correctly.
In my case, I'm experimenting with git based backup running under Cygwin and some target files have utf-8 (issue does not reproduce with usual latin/numeric only names) long names (200+ utf chars). Windows can handle them, but all operations on them failed under Cygwin. For example, I created (by Windows Explorer) the file with 220 `й` letters, then in Cygwin:
```
$ echo -n 'йййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййй' | wc -c  # check the real size
440

$ ll
ls: cannot access 'ййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййй'$'\320': No such file or directory
-????????? ? ?              ?        ?            ? 'ййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййй'$'\320'

$ ls й*
ls: cannot access 'ййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййй'$'\320': No such file or directory

$ stat й*
stat: cannot stat 'ййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййй'$'\320': No such file or directory
```
It's okay for me to skip these problematic files, but official Git terminates entire `git add` command when such files come up. But with this patch I have the desired behavior:
```
$ git add --ignore-errors .
error: ййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййййй can only add regular files, symbolic links or git-directories
```
(everything else added successfully)

So, it's like that in my case.
Thank you for your interest and I apologize for not replying so long.

dscho:
> I'm really sorry to keep you waiting so long...

Don't worry. This is _your_ contribution, and it takes however long it takes.

> So, it's like that in my case.

That is a pretty good explanation why this patch is needed. How about amending the commit message to add that explanation? That would convince people on the Git mailing list, I think.

Also, you probably want to amend the commit message so that the first line looks more like the first lines of other commits' messages, i.e. in your case something like

```
add: respect `--ignore-errors` when `lstat()` reports errors
```

Further, please make sure that your commit message wraps at <= 76 columns per line. Also do make sure to add your [sign off](https://git-scm.com/docs/SubmittingPatches#dco).

dscho:
@qusielle gentle ping?

dscho:
@qusielle are you still interested in this patch? If not, please close this PR.

dscho:
@qusielle another gentle ping? It would be nice to have this fix, and you're almost there.

qusielle:
@dscho, thanks a lot for guiding me.
I created a pull request in Gitgitgadget: https://github.com/gitgitgadget/git/pull/432
Hope I have done everything right.

dscho:
Excellent, thank you! Do you want to close this here PR, then?

qusielle:
@dscho 
Thank you again for your patience.
You are doing very great stuff for all the git users.
Yes, I will close this PR.

PR #582, opened 2019-03-12
Fix 'make test' preprocessing on Solaris 11.3 and friends

noloader:
* Provide variables AWK, SED, GREP
* Use tools from /usr/gnu/bin

Also see https://www.spinics.net/lists/git/msg355130.html .


noloader:
Defective, closing...

PR #585, opened 2019-03-22
Syntax, grammar, and language convention fixes

DjMorgul:
Added opening exclamation and question marks.

Applied the convention of translating "tree-ish" as "árbol-ismo". Might not be perfect, but it was already used like that mostly, and the other way it was translated as, "parte-del-árbol", was confusing.

Applied the convention of using "por defecto" all around, rather than "predeterminado" (it was only used a few times, and it takes longer space), or "default" (makes no sense to use that, when "por defecto" is a popular and 100% Spanish-language way of saying that).

Still a lot of work pending to be done, specially on deciding and applying the same language conventions all around, rather than mixing them up (as it's done right now). For example:
- Will we refer to the user in formal (usted) way, or informal (tú)? Right now it's mixed up, both ways used in different places. Beware of the places where it was translated as infinitive, and imperative. English language isn't explicit on whether it's using infinitive or imperative forms, specially when no subject was specified.
- Should we translate "string" as "cadena" (or better not, to not confuse it with the translation of "chain" as "cadena")?
- Should we translate "branch" as "rama", "tag" as "etiqueta"? Right now it seems to be mixed up.
- Isn't is too confusing to translate "file" as either "archivo" or "carpeta" based on context (using "archivo" to refer to a file system file, and "carpeta" to refer to a file within an archive), when file system "file" is being translated as "archivo"? It might be an acceptable patch for American Spanish (because they usually don't know use nor understand the word "fichero"), but in Spain's Spanish we can use "fichero" as a direct translation of "file", and we can still use "archivo" to refer to "archive", and so we can use "carpeta" to only mean "folder" (and never for meaning a file within an archive). We can make a clean translation in Spain's Spanish, without having to guess the meaning based on context, just needing to learn the convention that "archivo" will only be used to refer to archives, and not to files. Because we can use the words "archivo" and "fichero" indifferently to refer to files, but we're also used to seeing them with a bit differing meaning, such as in "ficheros dentro de un archivo comprimido" ("files within a compressed archive").

It would be very valuable to create language varieties for different Spanish speaking regions in the world, which tend to have their own language conventions, and it doesn't always look good enough to put all of them in the same box, much less mixing their differing popular conventions up on these translations.

### DIFF
--- a/po/es.po
+++ b/po/es.po
@@ -143,7 +143,7 @@ msgstr "--cached está fuera de un repositorio"
 #: apply.c:825
 #, c-format
 msgid "Cannot prepare timestamp regexp %s"
-msgstr "No se puede preparar una marca de tiempo para la expresión regular %s"
+msgstr "No se puede preparar expresión regular de marca de tiempo %s"
 
 #: apply.c:834
 #, c-format
@@ -239,7 +239,7 @@ msgstr "parche corrupto en la línea %d"
 #: apply.c:1841
 #, c-format
 msgid "new file %s depends on old contents"
-msgstr "nuevo archivo %s depende en contenidos viejos"
+msgstr "nuevo archivo %s depende de contenidos viejos"
 
 #: apply.c:1843
 #, c-format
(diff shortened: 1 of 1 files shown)


dscho:
The commit message leaves this unclear, so I'll spell it out here: this patch is about the Spanish translation.

The translations of Git are maintained in a separate repository: https://github.com/git-l10n/git-po/. Please rebase to `master` and re-open your PR there (it will not languish there as long as in this unmonitored repository, promise!).

PR #587, opened 2019-03-22
Make stashing nothing exit 1

keith:
In the case there are no files to stash, but the user asked to stash, we
should exit 1 since the stashing failed.

### DIFF
--- a/git-stash.sh
+++ b/git-stash.sh
@@ -318,7 +318,7 @@ push_stash () {
 	if no_changes "$@"
 	then
 		say "$(gettext "No local changes to save")"
-		exit 0
+		exit 1
 	fi
 
 	git reflog exists $ref_stash ||


submitgit:
@keith sent this commit (https://github.com/git/git/compare/041f5ea1cf987a4068ef5f39ba0a09be85952064...b880926df10d7bab0188c50427e1fdc951f3a19d) as a patch to the mailing list with [_submitGit_](https://github.com/rtyley/submitgit) - here on [public-inbox](https://public-inbox.org/git/01020169a7ad6af3-ad50e2d1-19fb-46eb-b397-759f8d579e8b-000000@eu-west-1.amazonses.com/), [MARC](http://marc.info/?i=01020169a7ad6af3-ad50e2d1-19fb-46eb-b397-759f8d579e8b-000000@eu-west-1.amazonses.com) [](01020169a7ad6af3-ad50e2d1-19fb-46eb-b397-759f8d579e8b-000000@eu-west-1.amazonses.com)

keith:
Mailing list folks don't want this

dscho:
> Mailing list folks don't want this

To clarify: making it the default is probably what seemed undesirable to the reviewers (myself included). If you still want the functionality, you could add it conditional on a command-line option.

Maxim-Mazurok:
I also think that it should return `1`, because `git stash pop` actually returns `1` when there are no stashes to pop. So, in this case:
```
git stash # < return code 0, but no files to stash
git stash pop # < return code 1, because previous command actually "failed"
```
I found this after debugging git automation and thought that it's not right.

@dscho what do you think, is it a valid argument?

dscho:
It does not matter what _I_ think. This should be discussed on the mailing list. I suspect that we might end up with the deprecation phase I described above.

Maxim-Mazurok:
Well, I wanted to get your opinion before reopening this discussion. Does it make sense at all?

dscho:
As I said, my opinion does not matter. Forming consensus on the Git mailing list does.

Maxim-Mazurok:
It does matter to me, but I can see why you might not want to publicly post your opinion. No problem, will try to get some opinions from the mailing list, thanks!

dscho:
No, I have no problem stating my opinion publicly.

What I have a problem with is the discussion _here_. It would be useful a discussion to have on the Git mailing list, unlike the discussion over here, which next to nobody sees, so it's very, very little "bang for the buck".

PR #590, opened 2019-03-31
Tried to make slash-rules more readable

iwasherefirst2:
As discussed in the mailing list, I hope this make the edited paragraphs more readable and clear.

Here are some posts that show what kind of problems people had to understand this part of the docs:
https://github.com/git/git-scm.com/issues/1332
https://stackoverflow.com/a/41761521/2311074
https://stackoverflow.com/questions/50028104/claryfing-gitignore-documentation

### DIFF
--- a/Documentation/gitignore.txt
+++ b/Documentation/gitignore.txt
@@ -103,9 +103,9 @@ PATTERN FORMAT
    then the pattern will match in all directories. In other words,
    `foo/` will match `/bar/foo/` and `foo` will match `/bar/foo`.
 
- - The character "`*`" matches anything except "`/`".
-   For example, "foo/*.json" matches
-   "foo/test.json" but not "foo/bar/test.json".
+ - The character "`*`" matches anything except a non trailing slash "`/`".
+   For example, "foo/*" matches
+   "foo/test.json" and "foo/bar/" but not "foo/bar/test.json".
    The character "`?`" matches any one character except "`/`".
    The character "`[]`" matches one character in a selected range.
    See fnmatch(3) and the FNM_PATHNAME flag for a more detailed
--- a/Documentation/gitignore.txt
+++ b/Documentation/gitignore.txt
@@ -89,23 +89,26 @@ PATTERN FORMAT
    Put a backslash ("`\`") in front of the first "`!`" for patterns
    that begin with a literal "`!`", for example, "`\!important!.txt`".
 
- - If the pattern ends with a slash, it is removed for the
-   purpose of the following description, but it would only find
-   a match with a directory.  In other words, `foo/` will match a
-   directory `foo` and paths underneath it, but will not match a
+ - If the pattern ends with a slash, it would only find
+   a match with a directory. In other words, `foo/` will match a
+   directory `foo` anywhere, but will not match a
    regular file or a symbolic link `foo` (this is consistent
    with the way how pathspec works in general in Git).
 
- - If the pattern does not contain a slash '/', Git treats it as
-   a shell glob pattern and checks for a match against the
-   pathname relative to the location of the `.gitignore` file
-   (relative to the toplevel of the work tree if not from a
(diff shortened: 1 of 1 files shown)


Documentation/gitignore.txt
```diff
@@ -89,23 +89,26 @@ PATTERN FORMAT
...
    regular file or a symbolic link `foo` (this is consistent
    with the way how pathspec works in general in Git).
 
- - If the pattern does not contain a slash '/', Git treats it as
-   a shell glob pattern and checks for a match against the
-   pathname relative to the location of the `.gitignore` file
-   (relative to the toplevel of the work tree if not from a
-   `.gitignore` file).
+ - If the pattern contains a slash '/' other then a trailing slash,
+   it is always considered from the root. In other words, `foo/bar`
```
PhilipOakley: I also suspect that "root" maybe a confused location. In Linux/*nix, root is a special place, while on Windows (Git-for-) root either doesn't exist (MS view), or could be anywhere (each bash has its own) depending on view point.

I think it actually means the location of the .gitignore file (and we can have many, repeated at many levels).

The whole .gitignore doc has a long history of confusion...

iwasherefirst2: This is true. Would you think its better to write "[...] is always considered from the location of the `.gitignore` file"?

PhilipOakley: 'It' is certainly worth suggesting. Though maybe "_the pattern_ is always considered from the `.gitignore` file location".

Also note that Git PRs, by them selves, are ignored. Once the checks are OK, one can send them upstream by either send-email, or via gitgitgadget.

I believe the documentation failure maybe that there has bee an update of the toolset. see https://public-inbox.org/git/20190329123520.27549-6-szeder.dev@gmail.com/

iwasherefirst2: Thank you for your help. Does it mean I have to create a fork at https://github.com/gitgitgadget/gitgitgadget, and create a PR with the same content and this PR? And about the documentation error, does it mean I simply have to wait till it gets fixed?

PhilipOakley:
If you are able, you can do the (documentation) checks yourself by using the SDK to have a local build system and directly check that the output formats well in both man view and pdf/html view. If the SDK then direct submission via format-patch/send-email may be quickest. (do you already have the SDK?)

I don't see that there is much to worry about for the CI failure in this case. Interacting with the upstream is probably more important.

iwasherefirst2:
@PhilipOakley I am not sure what you mean with SDK here. Couln't really get anything on google. I will try to push it to the gitgitgat repository.

PhilipOakley:
Hi @iwasherefirst2, The SDK (Software Development Kit) for https://gitforwindows.org/ is linked from the "[Contribute](https://gitforwindows.org/#contribute)" button on that page, should you wish to try that route. The [FAQ ](https://github.com/git-for-windows/git/wiki/FAQ)link takes you to the wiki pages (user editable content) with more details. But go with what works for you.

dscho:
I guess that this PR is now abandoned in favor of https://public-inbox.org/git/20190602090415.5889-1-admin@in-ici.net/#r. Correct, @iwasherefirst2?

iwasherefirst2:
> I guess that this PR is now abandoned in favor of https://public-inbox.org/git/20190602090415.5889-1-admin@in-ici.net/#r. Correct, @iwasherefirst2?

Yes. I thought it doesn't matter if this is closed or not because of
> Git Source Code Mirror - This is a publish-only repository and all pull requests are ignored. 
but I will close it now.

dscho:
> I thought it doesn't matter if this is closed or not because of
> 
> > Git Source Code Mirror - This is a publish-only repository and all pull requests are ignored.
> > but I will close it now.

True, true, but it does run the Azure Pipeline now, and I actually started looking at these PRs ;-) So I'd like only truly active PRs to be open. Thanks for closing this!

PR #605, opened 2019-05-03
Fix "Not a valid object name HEAD0" w/ p4 unshelve

mdymike:
git p4 unshelve fails with "fatal: Not a valid object name HEAD0" and "Command failed: git cat-file commit HEAD^0" on certain systems e.g. git version 2.21.0.windows.1 / python 2.7.16

It seems that the python pOpen implementation drops the ^ character when invoked using a string instead of an array as first argument.

Solution is to use the array format of passing the command to fOpen, which is preferred (see https://docs.python.org/2/library/subprocess.html) and is used in other parts of this code anyway.

### DIFF
--- a/git-p4.py
+++ b/git-p4.py
@@ -737,7 +737,7 @@ def extractLogMessageFromGitCommit(commit):
 
     ## fixme: title is first line of commit, not 1st paragraph.
     foundTitle = False
-    for log in read_pipe_lines("git cat-file commit %s" % commit):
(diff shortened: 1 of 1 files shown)


cjsquibb:
I ran into this issue today, and I appreciate the fix.

mdymike:
Hey - per that message from git I have resubmitted this pull request on gitgadget https://github.com/gitgitgadget/git/pull/183

In the meantime, it's easy to tweak that git-p4.py file yourself on disk to fix the problem immediately

PR #617, opened 2019-06-26
Fix bug in P4 Unshelve

ybeapps:
The correct syntax for parent commit with numeric argument is HEAD~{n} and not HAED^{n}

### DIFF
--- a/git-p4.py
+++ b/git-p4.py
@@ -3966,7 +3966,7 @@ class P4Unshelve(Command):
         """
 
         for parent in (range(65535)):
-            log = extractLogMessageFromGitCommit("{0}^{1}".format(starting_point, parent))
+            log = extractLogMessageFromGitCommit("{0}~{1}".format(starting_point, parent))
             settings = extractSettingsGitLog(log)
             if 'change' in settings:
                 return settings


dscho:
Thank you for contributing to Git!

Apart from the typo in the commit message, which would be easy to fix, I am actually not so sure that this patch fixes a bug instead of introducing one.

The notation `HEAD~3` does not refer to the 3rd parent of the current tip commit: it refers to the (first) parent of the (first) parent of the (first) parent of the tip commit.

Usually, when we talk about "2nd parent" in Git, we really refer to the commit merged into a merge commit In rare cases, Git users decide that they want to merge more than one branch (we call this an "octopus" merge), in which case there are more than two parent commits, and the "2nd parent" would merely refer to the first branch that was merged, out of the many that were.

Take for example https://github.com/git/git/commit/90d79d71910415387590a733808140e770382b2f. This is a merge commit. It has two parents: https://github.com/git/git/commit/f9089e8491fdf50d941f071552872e7cca0e2e04 and https://github.com/git/git/commit/fc7e03aace379c3d6b00ac977447f115421324fe. You can refer to them as 90d79d71910415387590a733808140e770382b2f^1 (or short: 90d79d71910415387590a733808140e770382b2f^) and 90d79d71910415387590a733808140e770382b2f^2. Regular commits do not have a second parent.

If you use `~` instead of `^`, it means that you follow the first-parent history. `HEAD~1` is the first parent (i.e. identical to `HEAD^`), `HEAD~2` is the (first) parent of `HEAD~1`, etc. And `HEAD~100` is the (first) parent of `HEAD~99`.

Therefore, `HEAD~1` is equivalent to `HEAD^1`, but `HEAD~2` is _not_ equivalent to `HEAD^2`.

So I am afraid that I think this here patch gets this wrong...

ybeapps:
This is not a solution. There is another one already accepted in [git-for-windows](https://github.com/git-for-windows).

Thanks @dscho for the great explanation

PR #618, opened 2019-06-28
l10n: de.po: Fix German translation

derphilipp:
Fix translation error of "complete" => "vollständig" instead of "unvollständig"


dscho:
Good point.

Would you mind opening a PR at https://github.com/git-l10n/git-po instead? That's where the translations of Git's messages are maintained.

dscho:
Oh, I see you opened https://github.com/ruester/git-po-de/pull/2 already, which is even better. Thank you!

PR #620, opened 2019-07-01
git-prompt: improve cherry-pick/revert detection

phillipwood:
Fix the detection of in-progress cherry-picks and reverts when the user commits or resets a conflict resolution in the middle of a sequence of picks/reverts.


phillipwood:
I should have opened this on gitgitgadget/git instead

PR #650, opened 2019-09-22
Turned the extern integer fetch_if_missing into field of struct repository 

dev-shubham1:
Hi GIT contributors !
I am Shubham Ghule.I have resolved the problem of extern integer fetch_if_missing by putting it in  the structure repository as a field.I have achieved this by declaring an object named repo of structure repository and including alloc.h file where fetch_if_missing was used.
I'm looking forward to your review.
Special thanks to @dscho. 
with regards,
Shubham.

### DIFF
--- a/alloc.h
+++ b/alloc.h
@@ -5,7 +5,10 @@ struct alloc_state;
 struct tree;
 struct commit;
 struct tag;
-struct repository;
+struct repository
+{
+    int fetch_if_missing;
+}repo;
 
 void *alloc_blob_node(struct repository *r);
 void *alloc_tree_node(struct repository *r);
--- a/builtin/pack-objects.c
+++ b/builtin/pack-objects.c
@@ -34,6 +34,7 @@
 #include "dir.h"
 #include "midx.h"
 #include "trace2.h"
+#include "alloc.h"
 
 #define IN_PACK(obj) oe_in_pack(&to_pack, obj)
 #define SIZE(obj) oe_size(&to_pack, obj)
--- a/builtin/pack-objects.c
+++ b/builtin/pack-objects.c
@@ -2847,14 +2847,14 @@ static int option_parse_missing_action(const struct option *opt,
 
 	if (!strcmp(arg, "allow-any")) {
 		arg_missing_action = MA_ALLOW_ANY;
-		fetch_if_missing = 0;
+		repo.fetch_if_missing = 0;
 		fn_show_object = show_object__ma_allow_any;
 		return 0;
 	}
 
 	if (!strcmp(arg, "allow-promisor")) {
 		arg_missing_action = MA_ALLOW_PROMISOR;
-		fetch_if_missing = 0;
+		repo.fetch_if_missing = 0;
 		fn_show_object = show_object__ma_allow_promisor;
 		return 0;
 	}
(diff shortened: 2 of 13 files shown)


alloc.h
```diff
@@ -5,7 +5,10 @@ struct alloc_state;
 struct tree;
 struct commit;
 struct tag;
-struct repository;
+struct repository
+{
+    int fetch_if_missing;
+}repo;
```
dscho: In this header, this is just a forward declaration of `struct repository`. The actual struct is defined in `repository.h`. You probably want to add the field there.

builtin/pack-objects.c
```diff
@@ -2847,14 +2848,14 @@ static int option_parse_missing_action(const struct option *opt,
 
 	if (!strcmp(arg, "allow-any")) {
 		arg_missing_action = MA_ALLOW_ANY;
-		fetch_if_missing = 0;
+		repo.fetch_if_missing = 0;
```
dscho: At this point, there is no variable `repo`. The variable you are looking for is probably `the_repository`, and as it is a pointed, you need to use `->` rather than `.` to dereference it.

fetch-object.c
```diff
@@ -4,14 +4,15 @@
 #include "strbuf.h"
 #include "transport.h"
 #include "fetch-object.h"
+#include "alloc.h"
 
 static void fetch_refs(const char *remote_name, struct ref *ref)
 {
 	struct remote *remote;
 	struct transport *transport;
 	int original_fetch_if_missing = fetch_if_missing;
```
dscho: This line also needs to be adjusted.

revision.c
```diff
@@ -2334,7 +2335,7 @@ static int handle_revision_opt(struct rev_info *revs, int argc, const char **arg
 		revs->ignore_missing = 1;
 	} else if (opt && opt->allow_exclude_promisor_objects &&
 		   !strcmp(arg, "--exclude-promisor-objects")) {
-		if (fetch_if_missing)
+		if (repo.fetch_if_missing)
```
dscho: IIRC `struct rev_info` contains a pointer to the repository in question. `repo` does not exist at this point.

fetch-object.c
```diff
@@ -4,14 +4,15 @@
...
 
-	fetch_if_missing = 0;
+	repo.fetch_if_missing = 0;
```
dscho: You will need to adjust the signature of `fetch_refs()` to receive a pointer to the repository, the convention is to use `struct repository *r`.

sha1-file.c
```diff
@@ -1416,7 +1417,7 @@ static int loose_object_info(struct repository *r,
 	return (status < 0) ? status : 0;
 }
 
-int fetch_if_missing = 1;
+repo.fetch_if_missing = 1;
```
dscho: That is not the correct spot to initialize default values of structs. The correct spot is `repo_init()` in `repository.c`.

sha1-file.c
```diff
@@ -1475,7 +1476,7 @@ int oid_object_info_extended(struct repository *r, const struct object_id *oid,
 		}
 
 		/* Check if it is a missing object */
-		if (fetch_if_missing && repository_format_partial_clone &&
+		if (repo.fetch_if_missing && repository_format_partial_clone &&
```
dscho: This function already accepts a parameter `r` of type `struct repository *`, please use that instead of the non-existing `repo` variable.

dscho:
@dev-shubham1 gentle ping?

dev-shubham1:
> @dev-shubham1 gentle ping?

dev-shubham1:
> @dev-shubham1 gentle ping?

working on it.I will submit the updated patch by today.

PR #651, opened 2019-09-25
t0061: fix test for argv[0] with spaces (MINGW only)

SyntevoAlex:
The test was originally designed for the case where user reported
that setting GIT_SSH to a .bat file with spaces in path fails on
Windows: https://github.com/git-for-windows/git/issues/692

The test has two different problems:
1. It fails to run on clean MINGW, because 'test-fake-ssh.exe' is
copied away from its dependencies 'libiconv.dll' and 'zlib1.dll'.
Even when test succeeds, this is still not quite right, because it
unexpectedly uses MINGW dependencies instead of the built ones.
2. With (1) out of the way, the test succeeds with AND without fix
eb7c7863 that addressed user's problem. This happens because the core
problem was misunderstood, leading to conclusion that git is unable to
start any programs with spaces in path on Win7. But in fact
   a) it only affects cmd.exe scripts, such as .bat scripts
   b) it only happens when cmd.exe receives at least two quoted args
   c) it happens on any Windows (verified on Win10).
Therefore, correct test must involve .bat script and two quoted args.

Fix both problems by using .bat script.
NOTE: With this change, the test now correctly fails without eb7c7863.

Signed-off-by: Alexandr Miloslavskiy <alexandr.miloslavskiy@syntevo.com>

### DIFF
--- a/t/t0061-run-command.sh
+++ b/t/t0061-run-command.sh
@@ -210,10 +210,24 @@ test_expect_success MINGW 'verify curlies are quoted properly' '
 	test_cmp expect actual
 '
 
-test_expect_success MINGW 'can spawn with argv[0] containing spaces' '
-	cp "$GIT_BUILD_DIR/t/helper/test-fake-ssh$X" ./ &&
-	test_must_fail "$PWD/test-fake-ssh$X" 2>err &&
-	grep TRASH_DIRECTORY err
+test_expect_success MINGW 'can spawn .bat with argv[0] containing spaces' '
+	bat_file="$TRASH_DIRECTORY/bat with spaces in name.bat" &&
+	out_file="bat_out_file.txt" &&
+	
+	# Every .bat invocation will log its arguments to $out_file
+	rm -f $out_file &&
+	printf "echo %%* >> $out_file" >"$bat_file" &&
+	
+	# Ask git to invoke .bat; clone will fail due to fake SSH helper
+	test_must_fail env GIT_SSH="$bat_file" git clone myhost:src ssh-clone &&
+	
+	# Spawning .bat can fail if there are two quoted cmd.exe arguments.
+	# .bat itself is first (due to spaces in name), so just one more is
+	# needed to verify. GIT_SSH will invoke .bat multiple times:
+	# 1) -G myhost
+	# 2) myhost "git-upload-pack src"
+	# First invocation will always succeed. Test the second one.
+	grep "git-upload-pack" "$out_file"
 '
 
 test_done


dscho:
Looks good!

Please note that running the test within a `make` or `prove` run will still pick up the correct `.dll` files. Even running the test script on its own will work, just not when building with MSVC.

Maybe you want to adjust the commit message?

Oh, and I restarted the failed Travis job. It was a temporary problem with `apt` that prevented Ubuntu packages from being installed in the Linux32 job.

To submit the patch to the Git mailing list for review, you could use [GitGitGadget](https://gitgitgadget.github.io) as before, or [submitGit](https://submitgit.herokuapp.com), or [send it manually](https://github.com/git-for-windows/git/blob/master/CONTRIBUTING.md#submit-your-patch).

SyntevoAlex:
For me the test fails at least when running it with `prove` or with `./t0061-run-command.sh`.

I have now tried to put `which libiconv.dll` into the test body and received this output when running `./t0061-run-command.sh` from Git SDK:
`which: no libiconv.dll in (<snip>/git-for-windows/t/../bin-wrappers:/c/Users/alex/bin:/mingw64/bin:/usr/local/bin:/usr/bin:/bin:/mingw64/bin:/usr/bin:/c/Users/alex/bin:<snip>)`

So this is expected that executable fails to run, because indeed its dependencies can't be found on PATH and also can't be found next to .exe.

Am I misunderstanding something?

dscho:
> For me the test fails at least when running it with `prove` or with `./t0061-run-command.sh`.

Oh, I thought via `make`... (`DEFAULT_TEST_TARGET = prove`). That will add `t/helper/` to the `PATH`, which is why it works in the Azure Pipeline (where `make` is used).

dscho:
Also, if you build in Git for Windows' SDK using GCC, and run the test suite also in that SDK, there won't be any `.dll` files in `t/helper/`.

SyntevoAlex:
So it seems that commit message was correct, then? Because the test fails without additional external settings? If not, can you please suggest the desired wording?

dscho:
> So it seems that commit message was correct, then?

In a manner of speaking.

What it left out was that the common way to build Git is _not_ using MSVC, but GCC. So in most cases, even when Git is built on Windows, there is _no_ problem with `.dll` files.

That's the part I am missing from the commit message.

SyntevoAlex:
Oh! Only now I understood that I pushed the wrong button and created PR in a wrong repo.
I have now re-created as https://github.com/gitgitgadget/git/pull/356

PR #666, opened 2019-11-03
[Outreachy] check-ref-format: parse-options 

Georgecanfly:
This command currently handles its own argv so by teaching it to
use parse-options instead we can standardize the way commands
handle user input across the project.

As a consequence of using OPT_BOOL data structure on --normalize &
--refspec-pattern, --no-normalize & --no-refspec-pattern can now be 
used.

NO_PARSEOPT flag was also removed to update git.c with the
conversion of the structure in this command.

This is a rough draft and I need some advice if I'm doing this
correctly since its being built but it is failing tests.

Helped by: emily shaffer <emilyshaffer@google.com>
Helped by: johannes schindelin <johannes.schindelin@gmx.de>

Signed-off-by: george espinoza <gespinoz2019@gmail.com>

### DIFF
--- a/builtin/check-ref-format.c
+++ b/builtin/check-ref-format.c
@@ -59,7 +59,6 @@ int cmd_check_ref_format(int argc, const char **argv, const char *prefix)
 	enum {
 		CHECK_REF_FORMAT_BRANCH,
 	};
-
 	int i = 0;
 	int verbose;
 	int normalize;
--- a/builtin/check-ref-format.c
+++ b/builtin/check-ref-format.c
@@ -6,10 +6,13 @@
 #include "refs.h"
 #include "builtin.h"
 #include "strbuf.h"
+#include "parse-options.h"
 
-static const char builtin_check_ref_format_usage[] =
-"git check-ref-format [--normalize] [<options>] <refname>\n"
-"   or: git check-ref-format --branch <branchname-shorthand>";
+static const char * const builtin_check_ref_format_usage[] = {
+	N_("git check-ref-format [--normalize] [<options>] <refname>\n"),
+	N_("   or: git check-ref-format --branch <branchname-shorthand>"),
+	NULL,
+};
 
 /*
  * Return a copy of refname but with leading slashes removed and runs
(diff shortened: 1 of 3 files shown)


Georgecanfly:
These are the tests my function edits are failing. @nasamuffin @dscho 

https://pastebin.com/GDV7F2WP

Georgecanfly:
When I was pushing it took an old commit and brought it into this fork as well, I tried rebase -i to delete it and ended up just manually putting merge-ours.c back to how it was before my edits. 
I had seen it after I did this:

~/git$ git pull -r
First, rewinding head to replay your work on top of it...
Applying: [Outreachy] merge-ours: include parse-option.h

builtin/check-ref-format.c
```diff
@@ -53,31 +56,28 @@ static int check_ref_format_branch(const char *arg)
 
 int cmd_check_ref_format(int argc, const char **argv, const char *prefix)
 {
-	int i;
-	int normalize = 0;
+	enum {
+		CHECK_REF_FORMAT_BRANCH,
+	};
+	int i = 0;
```
dscho: I think it would make more sense to get rid of `i` now. Its role was to iterate through the command-line options, which is now no longer performed in this function.

builtin/check-ref-format.c
```diff
@@ -53,31 +56,28 @@ static int check_ref_format_branch(const char *arg)
 
 int cmd_check_ref_format(int argc, const char **argv, const char *prefix)
 {
-	int i;
-	int normalize = 0;
```
dscho: I think `normalize` still needs to be initialized, lest it is used in an uninitialized state.

builtin/check-ref-format.c
```diff
@@ -53,31 +56,28 @@ static int check_ref_format_branch(const char *arg)
...
+	enum {
+		CHECK_REF_FORMAT_BRANCH,
+	};
+	int i = 0;
+	int verbose;
+	int normalize;
+	int allow_onelevel;
+	int refspec_pattern;
```
dscho: All of these need to be initialized to default values.

dscho:
> These are the tests my function edits are failing. @nasamuffin @dscho
> 
> https://pastebin.com/GDV7F2WP

Better would have been to paste it here, in this bug tracker.

Even better would have been to run the test scripts with `-i -v -x`, as described e.g. here: https://github.com/git-for-windows/git/wiki/Running-Git's-regression-tests#running-individual-tests

This is essentially what the Azure Pipeline does, and when you follow the failed checks all the way to the tests tab, you will see this: https://dev.azure.com/git/git/_build/results?buildId=1300&view=ms.vss-test-web.build-test-results-tab&runId=1018498&resultId=104057&paneView=attachments (you will want to either scroll down in the right pane, all the way to the end, or click into that pane to set the focus, then search via <kbd>Ctrl</kbd>+<kbd>F</kbd>):

```
expecting success of 1402.6 'ref name 'foo/bar/baz' is valid': 
		git check-ref-format  'foo/bar/baz'
	
++ git check-ref-format foo/bar/baz
error: last command exited with $?=1
not ok 6 - ref name 'foo/bar/baz' is valid

#	
#			git check-ref-format  'foo/bar/baz'
#		
expecting success of 1402.7 'ref name 'foo/bar/baz' is valid with options --normalize': 
		git check-ref-format --normalize 'foo/bar/baz'
	
++ git check-ref-format --normalize foo/bar/baz
error: last command exited with $?=1
```

That is not really helpful. I suspect that _all_ of these failures come from the fact that most of the newly-added variables are uninitialized.

Georgecanfly:
Ok great @dscho I'll look into all your suggestions from your comments. Thank you! I also just realized I did a PR on git's normal repository and not through gitgitgadget. So I'll have to transfer over my work there.

dscho:
> So I'll have to transfer over my work there.

Okay! Please do close this here PR after that, and add a link to the other PR, so that lurkers are redirected to the correct place.

Georgecanfly:
Correct place for a PR has been moved to GitGitGadget: 

https://github.com/gitgitgadget/git/pull/449

Closing this PR.

PR #667, opened 2019-11-08
contrib: Add --recursive flag for bash completion

purpleidea:


chriscool:
Hi James, Thanks for working on this!

Only patches sent to the Git mailing list can be applied by Junio Hamano though. So I can send it to the mailing list on your behalf, but in case it is discussed (which can always happen) it might be better if you do it yourself.

If you don't want to configure git send-email or you MUA, you can try [GitGitGadget](https://gitgitgadget.github.io/) to send it to the mailing list.

purpleidea:
@chriscool Hey it's all yours if you want it. You can even take the author credit. It's a pretty trivial patch that was in my long-term pending queue, and I don't have my msmtp setup at the moment. Thanks!

dscho:
> Hey it's all yours if you want it.

Why not put a _little_ bit more effort into it, and contribute back to the project from which you benefit, by your own account, quite frequently?

What you said could be construed to "throw scraps to the dogs", and I am sure that you appreciate all the work that was put into Git more than treating the other developers in such a manner.

> I don't have my msmtp setup at the moment.

That's why Chris pointed you to GitGitGadget. You open a PR at https://github.com/gitgitgadget/git (which has git/git's branches mirrored), and it allows you to send the patches as mails by adding a PR comment `/submit`.

While at it, you probably want to put a little finishing touch on the patch by looking whether other equally useful options are equally missing from the completions.

purpleidea:
@dscho Okay, I'll have a look, wasn't meant to be a negative comment. I spend my life contributing to open source for free, so chill =D I used to be on the git ml, but it's too high volume. Cheers!

chriscool:
@purpleidea thanks for taking a look at sending the patch! It really helps as coming up with a patch is often less than 50% of the work. By the way I don't think your comment was negative.

purpleidea:
Replaced with: https://github.com/gitgitgadget/git/pull/461

purpleidea:
FWIW, since I usually send PR's with the CLI via `hub pull-request`, because gitgitgadget is in a different repo it was hard to figure out how to do, so I had to use the webui. Maybe that bit could work here directly? Cheers

dscho:
> I usually send PR's with the CLI via `hub pull-request`, because gitgitgadget is in a different repo it was hard to figure out how to do

From the output of `hub pull-request --help`:

```
Opens a pull request on GitHub for the project that the "origin" remote
points to. [...]
```

The trick, therefore, would be to set your `origin` correctly. Or patch `hub`, I am sure they will welcome Pull Requests: https://github.com/github/hub

(You could easily find out all of that yourself, I had to download and build `hub` to do so...)

PR #679, opened 2019-12-13
rebase: make the default backend configurable and change the default

newren:
This is v5 of en/rebase-backend based on v2.25.0.  This series does a lot of work around making the default rebase backend configurable, and switches the default from the *apply* backend to the *merge* one.

Changes since v4:
   * Added a note about an unfortunately-timed Ctrl-C to the docs, as suggested by SZEDER.
   * Addressed feedback from Phillip
   * Added a new patch at the end renaming the two backends.  *PLEASE* take a look at its commit message for rationale and let me know if you agree with it.

Note that in particular, the last patch at the end could have been partially squashed into other patches, but if people objected to the patch in general I wanted Junio to be able to take the rest of the series and just drop that patch.

As with v4, there are some things mentioned earlier that are not addressed:
   * I did not remove the --am (now --apply) option as suggested by Phillip, since Junio and Phillip were still discussing whether it is wanted/needed.  I kind of like having it, but have no strong opinion.
   * I noted the post-commit hook in the differences between backends.  Emily is investigating what changes need to happen there, so I merely documented the existing differences.
   * I did not address the last two items Jonathan brought up as I couldn't find enough information to reproduce or understand the problems.

Cc: Johannes.Schindelin@gmx.de, phillip.wood@dunelm.org.uk, liu.denton@gmail.com, gitster@pobox.com, plroskin@gmail.com, alban.gruin@gmail.com, szeder.dev@gmail.com, jrnieder@gmail.com, emilyshaffer@google.com


dscho:
Heh, seems that 9106.9 really wants to verify that there is a `rebase-apply` directory... talk about overzealous regression tests...

newren:
> Heh, seems that 9106.9 really wants to verify that there is a `rebase-apply` directory... talk about overzealous regression tests...

Yep, was just looking at that and shaking my head as well.  Still haven't figured out 9140, but I'm suspecting an empty patch or something?

newren:
Yeah, 9140 is "empty" patches.  One commit started empty, and was followed by another commit that is a cherry-pick of another commit upstream.  The interactive backend goes to rebase that, optimizes out the two empty patches, and then throws an error because it doesn't have any work to do.  The am-backend, in contrast notes the empty patches, drops them, and completes the operation successfully.

dscho:
> The interactive backend goes to rebase that, optimizes out the two empty patches, and then throws an error because it doesn't have any work to do.

Hrm. Usually such a scenario would result in a todo list containing only the command `noop`.

newren:
> > The interactive backend goes to rebase that, optimizes out the two empty patches, and then throws an error because it doesn't have any work to do.
> 
> Hrm. Usually such a scenario would result in a todo list containing only the command `noop`.

Nah, the buffer will be non-empty; in particular, it will have a single commented-out line corresponding to a commit which started empty.

dscho:
> Nah, the buffer will be non-empty; in particular, it will have a single commented-out line corresponding to a commit which started empty.

I think that this is the real bug here: If all of the commands are commented-out, there should be a noop appended.

It does not make sense for the interactive rebase to use a single `noop` (and to succeed) if all of the patches have been applied upstream already, but to fail if there is even one empty commit (that is commented out because it should be skipped).

newren:
I was hoping retriggering would make the infrastructure failure go away but "Error LNK1181: cannot open input file 'libeay32.lib'" isn't something in my control and clearly is unrelated to my change -- especially since the windows tests succeeded and the windows build succeeded it's just the "Visual Studio Build" that failed.  So I'll submit anyway...

dscho:
> I was hoping retriggering would make the infrastructure failure go away but "Error LNK1181: cannot open input file 'libeay32.lib'" isn't something in my control and clearly is unrelated to my change -- especially since the windows tests succeeded and the windows build succeeded it's just the "Visual Studio Build" that failed. So I'll submit anyway...

Yes, I provided a fix in https://github.com/gitgitgadget/git/pull/527. (Are you saying you didn't see this? 👀 😄 )

newren:
> > I was hoping retriggering would make the infrastructure failure go away but "Error LNK1181: cannot open input file 'libeay32.lib'" isn't something in my control and clearly is unrelated to my change -- especially since the windows tests succeeded and the windows build succeeded it's just the "Visual Studio Build" that failed. So I'll submit anyway...
> 
> Yes, I provided a fix in [gitgitgadget#527](https://github.com/gitgitgadget/git/pull/527). (Are you saying you didn't see this?   )

Yep, I missed it.  Thanks for the pointer; good to know that it's been tracked down and that I can expect it until Junio merges that change so I can just watch it.

PR #684, opened 2019-12-28
diff-highlight: highlight range-diff

jablko:
Piping "git range-diff" through diff-highlight currently has no effect,
for two reasons:

  1. There are ANSI escapes before and after the "@@" hunk headers (when
     color is enabled) which diff-highlight fails to match. One solution
     is to match both escapes (/^$COLOR*\@\@$COLOR* /). This patch drops
     the trailing space from the pattern instead.

  2. Unlike "git log", "git range-diff" diffs are indented, which again,
     diff-highlight fails to match. This patch allows hunk headers
     preceded by any amount of whitespace, and then skips past that
     indentation when parsing subsequent lines. It reuses the machinery
     that handles the --graph output.


dscho:
@jablko please note that the Git project *requires* your [Sign-off](https://git-scm.com/docs/SubmittingPatches#sign-off). In the past, there have been contributions that lacked that `Signed-off-by:` footer and that were completely ignored as a consequence.

jablko:
@dscho Thanks a lot for your help! I've now signed-off. :heavy_check_mark:

dscho:
With many Git contributors trickling back in from holidays, you might want to reply to your cover letter on the mailing list, with a "gentle ping" or something like that ;-)

dscho:
@jablko I still think that you may want to re-start the conversation on the Git mailing list by replying with a gentle ping.

dscho:
@jablko have you received the mail with Jeff King's review? If so, please do reply (the custom on the Git mailing list is to use reply-to-all and to avoid top-posting but to reply inline like Jeff did).

phil-blain:
@jablko today I wanted to use this feature, remembered seeing a PR about it, googled it and discovered it was not ready yet. Thus, I also encourage you to continue working on this :)

PR #686, opened 2020-01-03
unpack-trees: exit check_updates() early if updates are not wanted

newren:
In trying to understand check_updates(), I found I was able to simplify the function by making it exit early when updates are not wanted.

Changes since v1:
  * Added two bullet points to the end of the commit message to explain the questions Junio brought up about function calls that are skipped by the early return.

### DIFF
--- a/unpack-trees.c
+++ b/unpack-trees.c
@@ -372,15 +372,20 @@ static int check_updates(struct unpack_trees_options *o)
 	state.refresh_cache = 1;
 	state.istate = index;
 
+	if (!o->update || o->dry_run) {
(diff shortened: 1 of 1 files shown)


newren:
I literally only changed the commit message, so previous testing should be adequate.  Also, errors like:
```
E: Failed to fetch https://packages.microsoft.com/repos/azure-cli/dists/xenial/main/binary-amd64/Packages  Writing more data than expected (40278 > 39457)
E: Some index files failed to download. They have been ignored, or old ones used instead.
```

are clearly not errors caused by my changes, so I'm just going to resubmit as-is...

PR #693, opened 2020-01-14
unpack-trees: stop ignoring return code from cache_tree_update

newren:
Here's a simple patch to stop ignoring the return code from cache_tree_update() in unpack_trees().  I suspect y'all want a testcase, but I'm struggling to come up with one.  When I was between commits with a large pile of changes with some local merge rework, I had several bugs and some of them resulted in the creation of broken trees.  However, I didn't take good notes.  This patch was only applicable/helpful when other things were broken, but it seems like it's still a good cleanup.


newren:
Oh, hmm...seems like that error code was meant to be ignored?  That's an awful lot of failures...

PR #694, opened 2020-01-15
t6025: modernize style

periperidip:
The tests in the file `t6025-merge-symlinks.sh` were written 11 years
ago and hence contain various style violations such as absence of
indentation in the code, irregular style usage of the redirection
operator, etc. Update it to match the latest CodingGuidlines

Signed-off by: Shourya Shukla <shouryashukla.oo@gmail.com>

### DIFF
--- a/t/t6025-merge-symlinks.sh
+++ b/t/t6025-merge-symlinks.sh
@@ -4,7 +4,6 @@
 #
 
 test_description='merging symlinks on filesystem w/o symlink support.
-
 This tests that git merge-recursive writes merge results as plain files
 if core.symlinks is false.'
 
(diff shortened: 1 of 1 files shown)


dscho:
> There is an issue in commit [34a0e43](https://github.com/git/git/commit/34a0e4368350cbf5f28b9fbd421abc00d46b64a9):
> Commit not signed off

@periperidip I think this is because the second dash in "Signed-off-by:" is a space in your commit. Would you please fix?

t/t6025-merge-symlinks.sh
```diff
@@ -17,14 +16,14 @@ test_expect_success 'setup' '
     git commit -m initial &&
     git branch b-symlink &&
     git branch b-file &&
-    l="$(printf file | 
-    git hash-object -t blob -w --stdin)" &&
+    l="$(printf >file
```
dscho: This is incorrect in two ways: `printf file` outputs `file` without line feed, but `printf` on its own most likely errors out. You have to redirect _in addition_ to keeping the original argument. Also, since you want this statement to be run on its own, not in a pipe, it needs to be on its own line:

```sh
printf file >out &&
l="$(git hash-object -t blob -w --stdin <out)" &&
```

t/t6025-merge-symlinks.sh
```diff
@@ -40,7 +39,7 @@ test_expect_success 'merge master into b-symlink, which has a different symbolic
 '
 
 test_expect_success 'the merge result must be a file' '
-    test -f symlink
+    test_path_is_file symlink
```
dscho: It probably makes sense to [split this commit](https://git-scm.com/docs/git-rebase#_splitting_commits) into two: the first one avoids the pipes, the second one replaces the `test -f` invocations (those are two separate concerns, therefore they should be addressed by two separate commits, too).

Oh, and please do not forget to fix the `Signed-off-by` line: there should be a `-` instead of a space before the `by:`.

PR #573, opened 2019-02-08
Clone now respects submodules.recurse config option

Maddimax:
Adds reading of the .gitconfig to git clone such that it activates the --recursive option if submodules.recurse is enabled in the config.

PR #695, opened 2020-01-15
Add the usage  of the submodules.recurse parameter on clone

Masmiseim36:
I try to finish the pullrequest #573 from Maddimax.
This adds the usage  of the submodules.recurse parameter on clone


t/t7407-submodule-foreach.sh
```diff
@@ -349,11 +349,26 @@ test_expect_success 'use "git clone --recursive" to checkout all submodules' '
...
+		git rev-parse --resolve-git-dir nested1/nested2/.git &&
+		git rev-parse --resolve-git-dir nested1/nested2/nested3/.git &&
+		git rev-parse --resolve-git-dir nested1/nested2/nested3/submodule/.git
+	)
+'
+
 test_expect_success 'test "update --recursive" with a flag with spaces' '
 	git clone super "common objects" &&
-	git clone super clone5 &&
+	git clone super clone6 &&
```
dscho: Better add the new test case at the end of the script, to avoid renaming these.

t/t7407-submodule-foreach.sh
```diff
@@ -349,11 +349,26 @@ test_expect_success 'use "git clone --recursive" to checkout all submodules' '
...
+		cd clone5 &&
+		git rev-parse --resolve-git-dir .git &&
+		git rev-parse --resolve-git-dir sub1/.git &&
+		git rev-parse --resolve-git-dir sub2/.git &&
+		git rev-parse --resolve-git-dir sub3/.git &&
+		git rev-parse --resolve-git-dir nested1/.git &&
+		git rev-parse --resolve-git-dir nested1/nested2/.git &&
+		git rev-parse --resolve-git-dir nested1/nested2/nested3/.git &&
+		git rev-parse --resolve-git-dir nested1/nested2/nested3/submodule/.git
+	)
```
dscho: It would be better to do this in a single invocation, and then to compare it to the expected values. Besides, I don't think that it needs to be exhaustive. So I would do this instead:

```sh
git -C clone5 rev-parse \
    --resolve-git-dir .git \
    --resolve-git-dir nested1/nested2/nested3/submodule/.git >actual &&
cat >expect <<-EOF &&
.git
$(pwd)/nested1/nested2/nested3/submodule/.git
EOF
test_cmp expect actual
```

builtin/clone.c
```diff
@@ -929,6 +929,19 @@ static int path_exists(const char *path)
...
 }
 
+/**
+ * Read config variables.
+ */
+static int git_pull_config(const char *var, const char *value, void *cb)
+{
+	if (!strcmp(var, "submodule.recurse")) {
+		option_recurse_submodules.nr = git_config_bool(var, value) ?
+			1 : option_recurse_submodules.nr;
```
dscho: This is most likely incorrect, as `option_recurse_submodules` is a `string_list`, and increasing the `nr` without even allocating a corresponding item, let alone initializing it, is most likely causing a crash.

builtin/clone.c
```diff
@@ -929,19 +931,19 @@ static int path_exists(const char *path)
...
-}
-
+/**
+ * Read config variables.
+ */
+static int git_clone_config(const char *var, const char *value, void *cb)
+{
+	if (!strcmp(var, "submodule.recurse")) {
+		option_recurse_submodules.nr = git_config_bool(var, value) ?
+			RECURSE_SUBMODULES_ON : RECURSE_SUBMODULES_OFF;
```
dscho: This does way more than just rename the function. And it still modifies `nr` without allocating/initializing the item. The correct thing to do here is more likely to be this:

```c
string_list_append(&option_recurse_submodules, ".");
```

(compare to what `recurse_submodules_cb()` does when no `arg` is provided)

dscho:
> I try to finish the pullrequest #573 from Maddimax.

Thank you for taking on this task!

The changes already look pretty good, but they should be squashed into a single one (use the `squash` command in an [interactive rebase](https://git-scm.com/docs/git-rebase#_interactive_mode), and while at it, add your sign-off via the [`--signoff` option](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---signoff)).

Do make sure to describe the motivation for this change well in the commit message ;-)

t/t7407-submodule-foreach.sh
Masmiseim36: OK, no problem. Can move it.
I thought it would be better readable to have the two tests for cloning with submodules (‘git clone –recursive’ vs ‘git clone -c submodule.recurse=true’) side by side.

t/t7407-submodule-foreach.sh
Masmiseim36: Sorry, don’t get it. I just copied the test for ‘git clone –recursive’. This command should do the identical job like git clone -c submodule.recurse=true’. Why do different tests?

builtin/clone.c
Masmiseim36: True, the Size is used as an indication of submodules active or not.

t/t7407-submodule-foreach.sh
dscho: It is always a balance between readability of the code after applying the patches, and readability of the patch itself. In this case, I would prefer the latter, to have an easier time preventing unintentional mistakes (such as forgetting _one_ place to rename, or forgetting to introduce a rename when merging an unrelated branch that adds another test case, too).

t/t7407-submodule-foreach.sh
dscho: Because you can improve the style? You also do not need to be _so_ extensive in an essentially slightly repeated test. Testing whether _one_ nested submodule was cloned should be _plenty_ enough.

builtin/clone.c
dscho: But what we really want to do is to stay as close to what passing the `--recurse-submodules` option would have accomplished, which is to add `"."`, not `"true"`...

dscho:
Please ignore the failing Visual Studio build for now. Your next push should "fix" this by virtue of `master` being fixed.

It would be good to fix the commit message, though, as the GitGitGadget bot pointed out. Please wrap the paragraphs at <=76 columns/line and leave empty lines between them.

Also, instead of "clone: respect submodules.recurse config option for automatically clone submodules" you could say "clone: respect the `submodule.recurse` config setting".

Please note that I suggested the `submodule.recurse` name, [which already exists](https://git-scm.com/docs/git-config#Documentation/git-config.txt-submodulerecurse), as opposed to `submodules.recurse` (note the singular vs plural). As a fun exercise, you could use `git blame` and/or `git log -L ...` to find out why it specifically excludes `clone`. It may have been intentional rather than due to implementation details.

t/t7407-submodule-foreach.sh
Masmiseim36: OK, I will take a closer look. I like to understand the test regression code a bit better first

dscho:
> The first line must be separated from the rest by an empty line

I see that you broke the first line of the commit message (you really have to look at the commit locally, it is not obvious in GitHub's UI):

```
clone: respect submodule.recurse config option for automatically clone
submodules

Simplify cloning repositories by making it transparent to the user if
submodules are used. The user doesn’t have to know if he has to add an
extra parameter to get the full project including the linked submodules.

It is done analog to the pull command by using an own config function
instead of using just the default config
```

Maybe you can think of another way to phrase it so that it gets shorter?

Also, I really think that you are _asking_ to be questioned why the documentation explicitly excludes `clone` from supporting that config setting. There's _got_ to be some information in the commit log about the documentation or on the Git mailing list about that patch that added this, and it would be good not only to be prepared to field those questions, but to preemptively write about it in the commit message.

Finally, you will still need to sign-off on the patch (easy: call `git commit --amend --signoff`). Looking at that, I am a bit perplexed by the author/committer information in your commit: the author seems to be "Masmiseim <masmiseim@gmx.de>" but the committer seems to be "Markus Klein <Markus.Klein@sma.de>" instead. It would be good to settle on the latter, I think. You can re-set the authorship via `git commit --amend --reset-author`.

dscho:
@Masmiseim36 ping?

Masmiseim36:
> @Masmiseim36 ping?
Hello,
I’m not gone, just a bit busy, sorry.
I went through some commits regarding clone and submodules and updated the commit comment a bit.
In the moment I’m trying to figure out the internals of the test code. I’m by far not an expert on shell-scripts.
However, I plan to finish this.

dscho:
I restarted the failed test (it failed for unrelated reasons in a flaky test, the re-run should "fix" it).

You will still be required to add your [Sign-off](https://git-scm.com/docs/SubmittingPatches#sign-off), GitGitGadget won't let you send the patch otherwise.

dscho:
Hmm. I think the Signed-off-by: line is _still_ incomplete, as it only says "Markus", not "Markus Klein". The reviewers on the Git mailing list won't like that, https://git-scm.com/docs/SubmittingPatches#sign-off specifically states:

> Also notice that a real name is used in the Signed-off-by: line. Please don’t hide your real name.

Masmiseim36:
Hello dscho,

I’ve extended the “Signed-off-by:” with my full name.

Thanks for the guidance and the feedback. I really appreciate that.

dscho:
Looks good to me. I think it's `/submit` time!

Masmiseim36:
Hello Johannes,

 

 

Von: gitgitgadget-git[bot] <notifications@github.com> 
Gesendet: Samstag, 1. Februar 2020 22:24
An: git/git <git@noreply.github.com>
Cc: Markus Klein <masmiseim@gmx.de>; Mention <mention@noreply.github.com>
Betreff: Re: [git/git] Add the usage ot he submodules.recurse parameter on clone (#695)

 

On the Git mailing list <https://public-inbox.org/git/nycvar.QRO.7.76.6.2002012218410.46@tvgsbejvaqbjf.bet> , Johannes Schindelin wrote (reply to this <https://github.com/gitgitgadget/gitgitgadget/wiki/ReplyToThis> ):

  This message is in MIME format.  The first part should be readable text,
  while the remaining parts are likely unreadable without MIME-aware tools.
 
--8323328-864764858-1580591979=:46
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: quoted-printable
 
Hi Markus,
 
 
On Fri, 31 Jan 2020, Markus Klein via GitGitGadget wrote:
 
> From: Markus <masmiseim@gmx.de <mailto:masmiseim@gmx.de> >
 
This line should probably match…
 
> 
> Simplify cloning repositories with submodules when the option
> submodules.recurse ot he by the user. This makes it transparent ot he
> user if submodules are used. The user doesn=E2=80=99t have to know if he=
 ot he add
> an extra parameter to get the full project including the used submodules=
..
> This makes clone behave identical to other commands like fetch, pull,
> checkout, … which include the submodules automatically if this option =
is
> set.
> 
> It is implemented analog ot he pull command by using an own config
> function instead of using just the default config. In contrast ot he pu=
ll
> command, the submodule.recurse state is ot hes an array of strings as =
it
> can take an optionally pathspec argument which describes which submodule=
s
> should be recursively initialized and cloned. To recursively initialize =
and
> clone all submodules a pathspec of „.“ ot he be used.
> The regression test is simplified compared ot he test for „git clone
> --recursive“ as the general functionality is already checked there.
> 
> Signed-off-by: Markus Klein <masmiseim@gmx.de <mailto:masmiseim@gmx.de> >
 
… this line. I.e. you will want to run `git config –global user.name
„Markus Klein“` and then `git commit –amend –reset-author`.
 
Ciao,
Johannes
 
Thanks for the hint. I’ve fixed this
 
Best Regards
 
Markus
 
 
> ---
>     Add the usage of the submodules.recurse parameter on clone
> 
>     I try to finish the pullrequest #573 from Maddimax. This adds the us=
age
>     of the submodules.recurse parameter on clone
> 
> Published-As: https://github.com/gitgitgadget/git/releases/tag/pr-git-69=
5%2FMasmiseim36%2Fdev%2FCloneWithSubmodule-v1
> Fetch-It-Via: git fetch https://github.com/gitgitgadget/git pr-git-695/M=
asmiseim36/dev/CloneWithSubmodule-v1
> Pull-Request: https://github.com/git/git/pull/695
> 
>  builtin/clone.c              | 16 +++++++++++++++-
>  t/t7407-submodule-foreach.sh | 11 +++++++++++
>  2 files changed, 26 insertions(+), 1 deletion(-)
> 
> diff --git a/builtin/clone.c b/builtin/clone.c
> index 0fc89ae2b9..21b9d927a2 100644
> --- a/builtin/clone.c
> +++ b/builtin/clone.c
> @@ -26,6 +26,8 @@
>  #include "dir-iterator.h"
>  #include "iterator.h"
>  #include "sigchain.h"
> +#include "submodule-config.h"
> +#include "submodule.h"
>  #include "branch.h"
>  #include "remote.h"
>  #include "run-command.h"
> @@ -929,6 +931,18 @@ static int path_exists(const char *path)
>        return !stat(path, &sb);
>  }
> 
> +/**
> + * Read config variables.
> + */
> +static int git_clone_config(const char *var, const char *value, void *c=
b)
> +{
> +      if (!strcmp(var, "submodule.recurse") && git_config_bool(var, value)) =
{
> +              string_list_append(&option_recurse_submodules, "true");
> +              return 0;
> +      }
> +      return git_default_config(var, value, cb);
> +}
> +
>  int cmd_clone(int argc, const char **argv, const char *prefix)
>  {
>        int is_bundle =3D 0, is_local;
> @@ -1103,7 +1117,7 @@ int cmd_clone(int argc, const char **argv, const c=
har *prefix)
> 
>        write_config(&option_config);
> 
> -      git_config(git_default_config, NULL);
> +      git_config(git_clone_config, NULL);
> 
>        if (option_bare) {
>                if (option_mirror)
> diff --git a/t/t7407-submodule-foreach.sh b/t/t7407-submodule-foreach.sh
> index 6b2aa917e1..44b32f7b27 100755
> --- a/t/t7407-submodule-foreach.sh
> +++ b/t/t7407-submodule-foreach.sh
> @@ -383,6 +383,17 @@ test_expect_success 'use "update --recursive nested=
1" to checkout all submodules
>                git rev-parse --resolve-git-dir nested1/nested2/nested3/submodule/.gi=
t
>        )
>  '
> +test_expect_success 'use "git clone" with submodule.recurse=3Dtrue to c=
heckout all submodules' '
> +      git clone -c submodule.recurse=3Dtrue super clone7 &&
> +      (
> +              git -C clone7 rev-parse --resolve-git-dir .git --resolve-git-dir nest=
ed1/nested2/nested3/submodule/.git >actual &&
> +              cat >expect <<-EOF &&
> +              .git
> +        $(pwd)/clone7/.git/modules/nested1/modules/nested2/modules/nested3/mo=
dules/submodule
> +              EOF
> +              test_cmp expect actual
> +      )
> +'
> 
>  test_expect_success 'command passed to foreach retains notion of stdin'=
 '
>        (
> 
> base-commit: d0654dc308b0ba76dd8ed7bbb33c8d8f7aacd783
> --
> gitgitgadget
> 
 
--8323328-864764858-1580591979=:46--

dscho:
> Thanks for the hint. I’ve fixed this Best Regards Markus

This was the only information in that otherwise garbled reply, right?

dscho:
> > Thanks for the hint. I’ve fixed this Best Regards Markus

I was able to verify this by looking at this: https://github.com/git/git/commit/a59ca4e27b7879f07d99d56cb972761e59714731.patch

> This was the only information in that otherwise garbled reply, right?

If so, please note that nobody on the Git mailing list saw your reply nor your change.

To send another iteration of this patch, please use GitGitGadget again: add a comment consisting of `/submit` and nothing else.

Before that, you will want to edit the initial comment (which will be sent as a sort of cover letter, embedded within the single-patch mail) to say something along these lines:

> Changes since v1:
> - Fixed the commit author.

Masmiseim36:
> > Thanks for the hint. I’ve fixed this Best Regards Markus
> 
> This was the only information in that otherwise garbled reply, right?

Sorry for the confusion, Maybe I should try a different E-Mail client.
Yes, that was all the information

dscho:
For what it's worth: by "initial comment", I meant the first comment on the PR, i.e. the PR description, not the commit message...

dscho:
@Masmiseim36 I saw that Junio offered a demo patch; did you end up using that? In any case, what is the status of this patch?

PR #698, opened 2020-01-20
git-p4: add hooks for p4-changelist

seraphire:
**Request for help** - if there are any experts with python and Windows, I would appreciate a review of the changes that I have presented here.

----

Our company's workflow requires that our P4 check-in messages have a specific format. A helpful feature in the GIT-P4 program would be a hook that occurs after the P4 change list is created but before it is displayed in the editor that would allow an external program to possibly edit the changelist text.

**v1:**
My suggestion for the hook name is p4-pre-edit-changelist.

It would take a single parameter, the full path of the temporary file.
If the hook returns a non-zero exit code, it would cancel the current P4 submit.

The hook should be optional.

**v2:**
Instead of a single hook, p4-pre-edit-changelist, follow the git convention for hook names and add the trio of hooks that work together, similar to `git commit.`

The hook names are: 

- p4-prepare-changelist
- p4-changelist
- p4-post-changelist

The hooks should follow the same convention as git commit, so a new command line option for the git-p4 submit function `--no-verify` should also be added.

**v3:**
2-Feb-2020
This version reorganizes the commits to be easier to read. The function for running the git hook has been rewritten to include suggestions.  This version no longer restricts the executable supported by windows for the hook.  It will first look for a file with the hook name without an extension and if found, it will pass it to sh.exe (Git for Window's MINGW shell) Otherwise it will take the file with the lowest alphabetical extension and ask Windows to execute the program.

**v5:**
10-Feb-2020
Integrated the suggested changes from the mailing list into this release.  The changes are:
* Restructure the commits to pull the p4-pre-submit hook changes into the commit for adding git_run_hook.
* Update the git_run_hook to include additional tests for GIT_DIR being set if getting the environment returns no value.
* Updated the windows file resolution to ensure that the script to be run is NOT the `.SAMPLE` version of the hooks.
* Split the RCS keyword cleanup message into its own commit.

**v6:**  
14-Feb-2020
Update commit for git-p4: restructure code in submit to be python 3 compatible:

> This change comes from 'git-p4: restructure code in submit' in
> 'bk/p4-pre-edit-changelist' which introduced the use of the `<>`
> operator. In Python 2, this is valid but in Python 3, it was removed.
> 
> We can simply replace the `<>` with `!=` which is the new way of
> writing "not equals".

### DIFF
--- a/Documentation/git-p4.txt
+++ b/Documentation/git-p4.txt
@@ -387,6 +387,41 @@ It can be bypassed with the `--no-verify` command line option.
 
 One usage scenario is to run unit tests in the hook.
 
+p4-prepare-changelist
+~~~~~~~~~~~~~~~~~~~~~
+
+The `p4-prepare-changelist` hook is executed right after preparing
+the default changelist message and before the editor is started.
+It takes one parameter, the name of the file that contains the
+changelist text. Exiting with a non-zero status from the script
+will abort the process.
+
+The purpose of the hook is to edit the message file in place,
+and it is not supressed by the `--no-verify` option. This hook
+is called even if `--prepare-p4-only` is set.
+
+p4-changelist
+~~~~~~~~~~~~~
+
+The `p4-changelist` hook is executed after the changelist
+message has been edited by the user. It can be bypassed with the
+`--no-verify` option. It takes a single parameter, the name
+of the file that holds the proposed changelist text. Exiting
+with a non-zero status causes the command to abort.
+
+The hook is allowed to edit the changelist file and can be used
+to normalize the text into some project standard format. It can
+also be used to refuse the Submit after inspect the message file.
+
+p4-post-changelist
+~~~~~~~~~~~~~~~~~~
+
+The `p4-post-changelist` hook is invoked after the submit has
+successfully occured in P4. It takes no parameters and is meant
+primarily for notification and cannot affect the outcome of the
+git p4 submit action.
+
+
 
 Rebase options
 ~~~~~~~~~~~~~~
--- a/Documentation/git-p4.txt
+++ b/Documentation/git-p4.txt
@@ -374,13 +374,19 @@ These options can be used to modify 'git p4 submit' behavior.
     been submitted. Implies --disable-rebase. Can also be set with
     git-p4.disableP4Sync. Sync with origin/master still goes ahead if possible.
 
-Hook for submit
-~~~~~~~~~~~~~~~
+Hooks for submit
+----------------
+
+p4-pre-submit
+~~~~~~~~~~~~~
+
 The `p4-pre-submit` hook is executed if it exists and is executable.
 The hook takes no parameters and nothing from standard input. Exiting with
 non-zero status from this script prevents `git-p4 submit` from launching.
+It can be bypassed with the `--no-verify` command line option.
 
 One usage scenario is to run unit tests in the hook.
+
 
 Rebase options
 ~~~~~~~~~~~~~~
--- a/Documentation/githooks.txt
+++ b/Documentation/githooks.txt
@@ -515,6 +515,52 @@ The exit status determines whether git will use the data from the
 hook to limit its search.  On error, it will fall back to verifying
 all files and folders.
 
+p4-changelist
+~~~~~~~~~~~~~
+
+This hook is invoked by `git-p4 submit`.
+
+The `p4-changelist` hook is executed after the changelist
+message has been edited by the user. It can be bypassed with the
+`--no-verify` option. It takes a single parameter, the name
+of the file that holds the proposed changelist text. Exiting
+with a non-zero status causes the command to abort.
+
+The hook is allowed to edit the changelist file and can be used
+to normalize the text into some project standard format. It can
+also be used to refuse the Submit after inspect the message file.
+
+Run `git-p4 submit --help` for details.
+
+p4-prepare-changelist
+~~~~~~~~~~~~~~~~~~~~~
+
+This hook is invoked by `git-p4 submit`.
+
+The `p4-prepare-changelist` hook is executed right after preparing
+the default changelist message and before the editor is started.
+It takes one parameter, the name of the file that contains the
+changelist text. Exiting with a non-zero status from the script
+will abort the process.
+
+The purpose of the hook is to edit the message file in place,
+and it is not supressed by the `--no-verify` option. This hook
+is called even if `--prepare-p4-only` is set.
+
+Run `git-p4 submit --help` for details.
+
+p4-post-changelist
+~~~~~~~~~~~~~~~~~~
+
+This hook is invoked by `git-p4 submit`.
+
+The `p4-post-changelist` hook is invoked after the submit has
+successfully occured in P4. It takes no parameters and is meant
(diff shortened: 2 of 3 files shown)


kisserlb:
In syntax git pull should be reversed to git push and git push to git pull 
that's all folks 
Mr IRCv3/Charmmer  CEO OF MASTEL

seraphire:
I re-read everything, and I had a typo in the title of the
gitgitgadget submission, "p4-pre-pedit-changelist" This should
have been "p4-pre-edit-changelist"

I have updated the commit message to follow the correct
formatting and fixed the title of the PR on github.

Since cee00 approved this PR, I'm assuming there is
no other work request for this.

Thank you to including this additional feature.

dscho:
>  Since cee00 approved this PR, I'm assuming there is no other work request for this.

That means nothing: that account has done nothing to gain credibility after being registered two days ago. The only public activity was to "approve" this PR. It is a pity that _anybody_ can add their approval to _any_ PR, no matter their standing in the project.

seraphire:
On 1/30/2020 1:57 PM, gitgitgadget-git[bot] wrote:
>
> On the Git mailing list 
> <https://public-inbox.org/git/xmqqmua4dc0b.fsf@gitster-ct.c.googlers.com>, 
> Junio C Hamano wrote (reply to this 
> <https://github.com/gitgitgadget/gitgitgadget/wiki/ReplyToThis>):
>
> |Ben Keene <seraphire@gmail.com> writes: > ... the name of the hook 
> from "p4-pre-edit-changelist" to follow the > git standard hooks: > > 
> * "p4-prepare-changelist" - This will replace the proposed hook but 
> still > Â  take only the filename. This hook will be called, even if 
> the > Â  prepare-p4-only option is selected. With "to follow", I 
> presume that this corresponds to the "prepare-commit-msg" hook? Does 
> "changelist" in P4 lingo roughly correspond to "commit", or to "commit 
> message"? |
|Yes that is correct, a changelist is P4's commit unit of work.|
> |> * "p4-changelist" - this is a new hook that will be added after the 
> > Â  user edit of the changelist text, but prior to the actual 
> submission. > Â  This hook will also take the temporary file as it's 
> only parameter > Â  and a failed response will fail the submission. 
> And I presume that this corresponds to "commit-msg" hook? |
|Yes, that is the idea.|||
> |>> * Given that "git commit" has a pair of hooks for log message, is 
> >> adding one new hook a reasonable thing? If so, the log mesasge >> 
> should explain why (e.g. perhaps the other one already is there, >> or 
> perhaps the other one is not applicable in the context of >> 
> interacting with P4 with such and such reasons).) > > I agree with 
> your suggestion. OK. Let's find them in the updated patch ;-) |
|It took a bit, but it's done!|||
> |>> * Is it reasonable not to have a mechanism to disable/skip the >> 
> hook, like "git commit" does? If not, the log message should >> 
> explain why such an escape hatch, which is needed for "git >> commit", 
> is not needed. > The existing hook, p4-pre-submit, does not have an 
> escape hatch, > so I did not add one to this method, but I can 
> certainly add one. It is probably a good idea in the longer term, but 
> we can certainly punt and then revisit to cover them with --no-verify 
> later (as long as we do document our intention to do so in the log 
> message). |
|It was a minor addition to add the no-verify feature.|||

dscho:
@seraphire please note that once again, your replies in this PR won't be seen on the Git mailing list. You will have to reply via email.

seraphire:
Hi Johannes,

The odd part is that I had actually replied to the email directly,
and it appears to have pushed my response back into github.

The force-push after the comment was fixing a simple typo that I
didn't want to push back to the mailing list until I received additional
feed back. It didn't seem right to generate noise for removing a
duplicate word.  The methon to the PR was to Yang's work on the Py2
to Py3 migration.

I don't know why it looks like I didn't reply to the mailing list.

- Ben

dscho:
> The odd part is that I had actually replied to the email directly, and it appears to have pushed my response back into github.

I cannot find it here, though: https://lore.kernel.org/git/pull.698.v2.git.git.1580507895.gitgitgadget@gmail.com/#r

Not a big deal, I just wanted you to understand that GitGitGadget won't mirror PR comments back to the Git mailing list, and won't do that in the foreseeable future, either.

seraphire:
On 2/4/2020 9:10 PM, gitgitgadget-git[bot] wrote:
>
> This branch is now known as |bk/p4-pre-edit-changelist| 
> <https://github.com/gitster/git/commits/bk/p4-pre-edit-changelist>.
>
I was in the process of introducing the changes that
Junio had suggested.  As this has now been picked up,
should I still commit my changes?

dscho:
>  This branch is now known as |bk/p4-pre-edit-changelist| <https://github.com/gitster/git/commits/bk/p4-pre-edit-changelist>.
> I was in the process of introducing the changes that Junio had suggested.  As this has now been picked up, should I still commit my changes?

Yes, you should. Maybe you want to reply on the mailing list to announce your intentions, lest Junio merge this to `next`?

(The rule is: as long as patches are in `pu`, they are fair game for complete rewriting, but as soon as they hit `next`, there should only be incremental changes on top, except in very rare cases when the patches are kicked out of `next` again.)

seraphire:
FYI - I am working on implementing the suggestions from Junio and 
Johannes into this patch set. I hope to have something shortly.

dscho:
> FYI - I am working on implementing the suggestions from Junio and Johannes into this patch set. I hope to have something shortly.

@seraphire Junio will not get a notification on this, nor anybody else on the Git mailing list. This comment of yours is only visible in this here PR, and only I am following it (and maybe half a dozen lurkers).

You probably _really_ want to reply via email to the Git mailing list, not to the PR notifications.

seraphire:
Here is the list of addresses I'm replying to:

git/git <reply@reply.github.com>

Johannes Schindelin <notifications@github.com>

git/git <git@noreply.github.com>

Mention <mention@noreply.github.com>

Do I need to add: git@vger.kernel.org as well? ( I added it to this 
email to ensure that it makes it.)

Any other addresses I need to ensure exist when I reply?

dscho:
> Here is the list of addresses I'm replying to: git/git <reply@reply.github.com> Johannes Schindelin <notifications@github.com> git/git <git@noreply.github.com> Mention <mention@noreply.github.com>

Those are _all_ unmonitored `@github.com` email addresses.

dscho:
>  Do I need to add: git@vger.kernel.org as well? ( I added it to this email to ensure that it makes it.)

If you want to reply to a specific mail that is mirrored into this PR as a PR comment, you will have to follow the instructions at https://github.com/gitgitgadget/gitgitgadget/wiki/ReplyToThis and use the "On the Git mailing list" link at the top of said PR comment.

dscho:
Please ignore the failing "Documentation" job; I proposed a workaround for that in #707.

seraphire:
Fixed for Python 3 integration:

> This change comes from 'git-p4: restructure code in submit' in
> 'bk/p4-pre-edit-changelist' which introduced the use of the `<>`
> operator. In Python 2, this is valid but in Python 3, it was removed.
> 
> We can simply replace the `<>` with `!=` which is the new way of
> writing "not equals".

PR #711, opened 2020-02-16
check-ignore: fix handling with negated patterns

newren:
Changes since v2:
- Add missing comma, as pointed out by Junio.


newren:
The travis CI was an infrastructure failure/flicker, so I'm going to ignore it...

newren:
Only difference from v2 was an added comma in the documentation, so I'm going to skip waiting for the test results and just submit...

PR #715, opened 2020-02-26
merge-recursive: apply collision handling unification to recursive case

newren:
This commit ties up a loose end that I was unaware of with the en/merge-path-collision topic from very early 2019.


cloudcalad:
Abrir
Grasias

Facebook

PR #718, opened 2020-02-27
Codify that for_each_rawref / for_each_ref should iterate over the

hanwen:
"refs/" name space.

This happens implicitly in the files/packed ref backend; making it
explicit will simplify adding alternate ref storage backends.

### DIFF
--- a/refs.c
+++ b/refs.c
@@ -1569,7 +1569,7 @@ static int do_for_each_ref(struct ref_store *refs, const char *prefix,
 
 int refs_for_each_ref(struct ref_store *refs, each_ref_fn fn, void *cb_data)
 {
-	return do_for_each_ref(refs, "", fn, 0, 0, cb_data);
+	return do_for_each_ref(refs, "refs/", fn, 0, 0, cb_data);
 }
 
 int for_each_ref(each_ref_fn fn, void *cb_data)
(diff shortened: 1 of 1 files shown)


refs.c
```diff
@@ -1629,8 +1629,8 @@ int for_each_namespaced_ref(each_ref_fn fn, void *cb_data)
 
 int refs_for_each_rawref(struct ref_store *refs, each_ref_fn fn, void *cb_data)
 {
-	return do_for_each_ref(refs, "", fn, 0,
-			       DO_FOR_EACH_INCLUDE_BROKEN, cb_data);
+	return do_for_each_ref(refs, "refs/", fn, 0, DO_FOR_EACH_INCLUDE_BROKEN,
```
dscho: If I were you, I would avoid pulling the `DO_FOR_EACH_INCLUDE_BROKEN` flag into the previous line: this is a change you do not need to make, and it is distracting from the _actual_ change (on the Git mailing list, you don't have color, so it will be even harder to review what you did).

dscho:
@hanwen gentle reminder that this PR is still open ;-)

hanwen:
thanks Dscho.

sorry for the messy commit message. I just wanted to pass this by GGG's CI to see if there were gotchas I didn't think of.

hanwen:
I'll put this in my reftable series now.

refs.c
hanwen: this is clang-format's doing.  I prefer to let the tools make mechanical decisions; they're much better at it than I am.

PR #719, opened 2020-03-02
doc: use 'ref' instead of 'commit' for merge-base arguments

tnir:
The notation `<commit>` can be misunderstandable only for commit SHA1,
but `merge-base accepts` any commit references. Like `reflog`, the name of
arguments should be `<ref>` rather than `<commit>`.

Signed-off-by: Takuya Noguchi <takninnovationresearch@gmail.com>


dscho:
GitGitGadget needs to know your public email address, as the Git mailing list requires it.

To that end, it looks at your public profile. Can you make sure that at least one email address is visible there?

tnir:
Thanks @dscho

dscho:
@tnir what's the status of this PR? Does it need to be updated and a new iteration `/submit`ted?

PR #720, opened 2020-03-07
Teach ref-filter API to correctly handle CRLF in messages

phil-blain:
The ref-filter API does not correctly handle commit or tag messages that
use CRLF as the line terminator. Such messages can be created with the
`--verbatim` option of `git commit` and `git tag`, or by using `git
commit-tree` directly.

This impacts the output of `git branch -v`, and `git branch`, `git
tag` and `git for-each-ref` when used with a `--format` argument
containing the atoms `%(contents:subject)` or `%(contents:body)`.

The first patch in this series adds a new test library, `t/lib-crlf-messages.sh`, 
that contains functions to test this behaviour for commands using either 
the ref-filter or the pretty API to display messages. 

The second patch fixes the bug in ref-filter.c and adds corresponding tests.

Finally, the third patch adds a test that checks the behaviour of `git log` in the presence of 
CRLF in messages, to prevent futur regressions.

Cc: Michael J Gruber <git@grubix.eu>, Matthieu Moy <git@matthieu-moy.fr>, John Keeping <john@keeping.me.uk>, Karthik Nayak <karthik.188@gmail.com>, Jeff King <peff@peff.net>, Alex Henrie <alexhenrie24@gmail.com>

### DIFF
--- a/ref-filter.c
+++ b/ref-filter.c
@@ -1050,10 +1050,18 @@ static char *copy_subject(const char *buf, unsigned long len)
 {
 	char *r = xmemdupz(buf, len);
 	int i;
+	struct strbuf sb = STRBUF_INIT;
 
-	for (i = 0; i < len; i++)
+	strbuf_attach(&sb, r, len, len + 1);
+	for (i = 0; i < sb.len; i++) {
 		if (r[i] == '\n')
 			r[i] = ' ';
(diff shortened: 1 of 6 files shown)


phil-blain:
I'll use GitGitGadget on Gitgitgadget's Git fork : see gitgitgadget/git#576.

PR #722, opened 2020-03-11
Fix 2.26.0 rebase regression and documentation shortcoming

newren:
This two commit series addresses two points raised by Peff about rebase backend issues.  The first is a two-line fix to a regression in 2.26.0 (when "the eighth batch for 2.26.0" added the dropping of commits which become empty, if the last commit in the series was the one that became empty the rebase would complete without cleaning out state files), and the other is just a documentation update about a backend difference that we were previously unaware of.

Changes since v1:
  * Clean out any MERGE_MSG file in addition to CHERRY_PICK_HEAD, and add a test

Cc: Phillip Wood <phillip.wood123@gmail.com>, Jeff King <peff@peff.net>, Junio C Hamano <gitster@pobox.com>

### DIFF
--- a/Documentation/git-rebase.txt
+++ b/Documentation/git-rebase.txt
@@ -699,6 +699,16 @@ suffer from the same shortcoming.  (See
 https://lore.kernel.org/git/20200207132152.GC2868@szeder.dev/ for
 details.)
 
+Commit Rewording
+~~~~~~~~~~~~~~~~
+
+When a conflict occurs while rebasing, rebase stops and asks the user
+to resolve.  Since the user may need to make notable changes while
+resolving conflicts, after conflicts are resolved and the user has run
+`git rebase --continue`, the rebase should open an editor and ask the
+user to update the commit message.  The merge backend does this, while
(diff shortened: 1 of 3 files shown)


dscho:
@emaste any idea why freebsd_12 is failing? It looks like it received signal 15 (`SIGTERM`?)?

emaste:
There's an issue with FreeBSD VMs used by Cirrus at the moment it seems, see more details at https://github.com/cirruslabs/cirrus-ci-docs/issues/592?email_token=AAH4SVQE234ICQ6EITUYULTRGEP53A5CNFSM4LCSJLD2YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOEOB6TGY#issuecomment-595847579

We're trying to figure out what's happening.

You can try the rerun button on the Cirrus panel.

newren:
I'll ignore the FreeBSD one then and just submit...

dscho:
> You can try the rerun button on the Cirrus panel.

Right, I should have done that right away. That turned the build green.

PR #726, opened 2020-03-14
Sparse checkout improvements -- improved sparsity updating

newren:
This series provides a replacement for the sparsity updating in sparse-checkout that is based on the logic from `git read-tree -mu HEAD`.  The most important bit is patch 9 and its lengthy commit message explaining the current state and rationale for most the series, though patches 16 and 17 have additional related directions and rationale for the series.  Those three patches are the most important to review.

Changes since v2:
  - addressed Stolee's cleanups and added his Reviewed-by to the series (hopefully the next two changes don't invalidate that)
  - added a test for the new 'reapply' subcommand (noticed it was missing with Stolee's test coverage report)
  - fix a minor issue with two of the other tests I modified -- when I'm grepping through stderr for a command, I need to make sure to actually record stderr from that command (otherwise my grep is looking through the stderr of a previous command that was recorded).  Also, since I messed this up, I added one or two more sanity checks while I was at it.

Cc: Derrick Stolee <dstolee@microsoft.com>


newren:
The error in the Travis build was only for gccXcode and was due to 'An error occurred while generating the build script.', i.e. it was an infrastructure failure, not a failure of any code I changed.  So I'm going to ignore that one.

PR #731, opened 2020-03-18
RelNotes/2.26.0: fix various typos

newren:
gitgitgadget requires a description even though for single-commit series it doesn't use it to generate a cover letter.  So now there's some useless text between the commit message and diff.  Sorry.


newren:
> This branch is now known as [`master`](https://github.com/gitster/git/commits/master).

Well then, let's go ahead and close this out.  :-)

PR #739, opened 2020-03-27
pull: avoid running both merge and rebase

newren:
Cc: Norbert Kiesel <nkiesel@gmail.com>, Jeff King <peff@peff.net>


newren:
FreeBSD build timed out download stuff, so I'll ignore it.  The TravisCI build decided to wait an hour before even starting, so I'll ignore it too.

PR #740, opened 2020-03-27
git-rebase.txt: fix typo

phil-blain:
CC: Junio C Hamano <gitster@pobox.com>, Elijah Newren <newren@gmail.com>

### DIFF
--- a/Documentation/git-rebase.txt
+++ b/Documentation/git-rebase.txt
@@ -605,7 +605,7 @@ BEHAVIORAL DIFFERENCES
 -----------------------
 
 git rebase has two primary backends: apply and merge.  (The apply
-backend used to known as the 'am' backend, but the name led to
+backend used to be known as the 'am' backend, but the name led to
 confusion as it looks like a verb instead of a noun.  Also, the merge
 backend used to be known as the interactive backend, but it is now
 used for non-interactive cases as well.  Both were renamed based on


phil-blain:
@dscho Is gitgitgadget the same version of gitgitgadget/git and here ? I thought one-patch contribution did not need a cover letter anymore ?

phil-blain:
Also, for such contributions (ex. only `Documentation/` is changed) would it be possible for the CI scripts to only run the Documentation build ? or would that be too complicated and we have enough CI ressources to not bother ?

dscho:
> Is gitgitgadget the same version on gitgitgadget/git and here ?

Should be. Let me verify (re-running the job).

dscho:
Yep, claims to be [up to date](https://dev.azure.com/gitgitgadget/git/_build/results?buildId=35233&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=a46d8783-b972-547c-42b9-2f94289d3941&l=14): https://github.com/gitgitgadget/gitgitgadget/commit/91e1028 is [the latest commit on `master`](https://github.com/gitgitgadget/gitgitgadget/commits/master).

Meaning that the Error should not be an error, at least not for single-patch contributions.

dscho:
> Also, for such contributions (ex. only `Documentation/` is changed) would it be possible for the CI scripts to only run the Documentation build?

I am not aware of any way to do this, unless we jump through several hoops: we would have to start the Pipeline by one job that determines what needs to be tested, and then we would have to make all other jobs dependent, and skip them if they should not be included.

At some stage, Azure Pipelines might support some sort of Test Impact Analysis (and then we would _still_ need quite a bit of support from core Git's side to use that), but for the time being, they don't.

phil-blain:
@dscho I tried to just add CC's but they didn't get parsed properly

webstech:
> I tried to just add CC's but they didn't get parsed properly

There may be an expectation of an empty line preceding the `cc` (ie it is a footer).

PR #742, opened 2020-03-29
checking if new travis code work with azure

sgn:
I will send it via git-sendemail


dscho:
> I will send it via git-sendemail

As you wish. Should you make up your mind and decide to use the more convenient way, there's nothing preventing you from doing that now. :-)

PR #746, opened 2020-04-01
sequencer: honor GIT_REFLOG_ACTION

newren:
Changes since v1:
   - Just use preprocessor constant (instead of mix of it and string constant)
   - Add a comment next to code that surprised both Phillip and Dscho

Cc: jrnieder@gmail.com, elbrus@debian.org, ijackson@chiark.greenend.org.uk, phillip.wood@dunelm.org.uk, alban.gruin@gmail.com, Johannes.Schindelin@gmx.de


Marzog00:
> /submit

PR #760, opened 2020-04-15
range-diff: fix a crash in parsing git-log output

vasild:
* override a possibly user-customized `format.pretty` that would render `git log` output unparsable by `git range-diff`

* don't use negative string precision, e.g. `"%.*s", -5, "foo"`

Changes since v1:
- Fixed a typo in the commit message (found by Taylor Blau)

### DIFF
--- a/range-diff.c
+++ b/range-diff.c
@@ -123,16 +123,19 @@ static int read_patches(const char *range, struct string_list *list,
 			struct patch patch = { 0 };
 			struct strbuf root = STRBUF_INIT;
 			int linenr = 0;
+			int orig_len;
 
 			in_header = 0;
 			strbuf_addch(&buf, '\n');
 			if (!util->diff_offset)
 				util->diff_offset = buf.len;
 			line[len - 1] = '\n';
+			orig_len = len;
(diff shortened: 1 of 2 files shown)


range-diff.c
```diff
@@ -106,6 +108,17 @@ static int read_patches(const char *range, struct string_list *list,
 			continue;
 		}
 
+		if (util == NULL) {
```
dscho, line 111: In Git's source code, we usually write `if (!util)` in such instances. Maybe you want to do the same here?

range-diff.c
```diff
@@ -123,16 +123,18 @@ static int read_patches(const char *range, struct string_list *list,
 			struct patch patch = { 0 };
 			struct strbuf root = STRBUF_INIT;
 			int linenr = 0;
+			int oldlen;
```
dscho, line 126: Maybe `orig_len`?

range-diff.c
vasild: Done in b3384880c7. Btw, `git grep '== NULL'` returns 298 matches.

range-diff.c
vasild: Done in b3384880c7.

PR #762, opened 2020-04-18
fix minor glitch in an illustration in the documentation

Nudin:
In the example by Jon Loeliger the selector 'A^2' was duplicated. Since it is the only duplicate and doesn't fix in the structure I assume it is there by incident and  since it can confuse readers this patch removes it.


dscho:
> * the first line should be like a header and typically start with a prefix like "tests:" or "commit:", and
> 
> 

@Nudin would you mind amending the commit message that way?

dscho:
> > * the first line should be like a header and typically start with a prefix like "tests:" or "commit:", and
> > 
> > * the commit messages' body should be describing the "why?" of the change.
> > 
> > * Finally, the commit messages should end in a [Signed-off-by:](https://git-scm.com/docs/SubmittingPatches#dco) line matching the commits' author.
> 
> @Nudin would you mind amending the commit message that way?

I see that the commit has a sign-off. But the first two bullet points still await to be addressed.

PR #764, opened 2020-04-20
problems with git in linux ubuntu / impossible install or update git 

Anna-Myzukina:
impossible to run
 
```
sudo apt install git
```

I get the next output:

Package git is not available but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'git' has no installation candidate


dscho:
Please understand that there is a good reason why issues are disabled in this repository (and that there are better ideas than to abuse Pull Requests for it). You probably want to use any of the avenues described in https://git-scm.com/community to reach out about this.

PR #765, opened 2020-04-22
Xdiff as submodule

j143:
This PR is here for CI testing.


j143:
@dscho - would you allow this to run CI.

dscho:
Well, it's now running.

I have to admit that I'm not super enthusiastic about having this as a submodule.

j143:
What is the best way, @dscho .

dscho:
I don't know... you could imitate sha1collisiondetection, which is both a submodule *and* copied code.

It might be best to discuss this on the Git mailing list. Other core Git contributors might not be so submodule-averse as I am. Of course, that might have changed, what with submodules playing a critical role in almost *all* the security fixes of the past five years. Only a tiny fraction of the vulnerabilities would have been exploitable even without submodules. And the wounds of v2.26.1 & v2.26.2 are *very* fresh, still.

dscho:
> you could imitate sha1collisiondetection, which is both a submodule _and_ copied code.

Please note that the fact that there _is no copied code_ but _only_ a submodule is responsible for [all those build failures](https://dev.azure.com/git/git/_build/results?buildId=1976&view=logs&j=1f6ce3b7-a951-539a-a522-fbaaeab307c4&t=ec8367e0-c809-579c-af5f-fa27c0b49cdc&l=597):

```
make: *** No rule to make target 'xdiff/xdiffi.c', needed by 'xdiff/xdiffi.o'.  Stop.
```

dscho:
As I mentioned elsewhere, it might make a _lot_ more sense to include xdiff via a subtree merge than as a submodule.

j143:
Thank you for the input. I will try to come back in awhile.

PR #768, opened 2020-04-24
commit:fix use of uninitialized value [...] in server log

rapsys:
This change fix the message about uninitialized value when trying to
access undefined hash indexes.

The error message fixed:
Use of uninitialized value $params{"action"} in string eq at gitweb.cgi
line 1377

Add myself as warning fix author.

Signed-off-by: Raphaël Gertz <git@rapsys.eu>

### DIFF
--- a/gitweb/README
+++ b/gitweb/README
@@ -66,5 +66,8 @@ AUTHORS
 Originally written by:
   Kay Sievers <kay.sievers@vrfy.org>
 
+Perl warning fix:
+  Raphaël Gertz <git@rapsys.eu>
+
 Any comment/question/concern to:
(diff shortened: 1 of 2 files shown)


rapsys:
commit: Fix 'use of uninitialized value' warnings in server error log

This is a patch I tried to submit long ago.

With the current perl script, there are a lot of errors in web server log
because of undefined hash indexes.

This patch change that by testing that a hash index is defined before doing
tests on it.

Plz merge :)

Signed-off-by: Raphaël Gertz git@rapsys.eu

rapsys:
Canceled

PR #769, opened 2020-04-27
Change the explanation so that it cannot be misinterpreted.

mickerules:
The ", as if you are giving orders to the codebase to change its behavior" describes a correct and two incorrect examples. It is unclear which of the examples it refers to and can be easily misinterpreted.

### DIFF
--- a/Documentation/SubmittingPatches
+++ b/Documentation/SubmittingPatches
@@ -133,10 +133,10 @@ The body should provide a meaningful commit message, which:
 . alternate solutions considered but discarded, if any.
 
 [[imperative-mood]]
-Describe your changes in imperative mood, e.g. "make xyzzy do frotz"
-instead of "[This patch] makes xyzzy do frotz" or "[I] changed xyzzy
-to do frotz", as if you are giving orders to the codebase to change
-its behavior.  Try to make sure your explanation can be understood
+Describe your changes in imperative mood, e.g. "make xyzzy do frotz", 
+as if you are giving orders to the codebase to change its behavior.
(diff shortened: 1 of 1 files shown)


dscho:
> Invalid author email in c524897e1b7cb2662f589b6a1403cac9249f5e5d: "64397351+mickerules@users.noreply.github.com"

@mickerules the Git project requires the real name and the correct email address for contributions...

Documentation/SubmittingPatches
```diff
@@ -135,8 +135,10 @@ The body should provide a meaningful commit message, which:
 [[imperative-mood]]
 Describe your changes in imperative mood, e.g. "make xyzzy do frotz", 
 as if you are giving orders to the codebase to change its behavior.
-Do not describe you changes in present or past tense, e.g. "[This patch] makes xyzzy do frotz" or "[I] changed xyzzy
-to do frotz". Try to make sure your explanation can be understood
+Do not describe you changes in present or past tense,
```
dscho, line 138: you*r* changes, right?

Please do wrap the commit message at 72 or 76 columns per line. Also, the second commit is essentially a fixup for the first one. Please use [`git rebase -i` to squash the second into the first](https://git-scm.com/docs/git-rebase#_interactive_mode): no need to send a known-incorrect patch to the Git mailing list, together with a patch that immediately fixes it.

dscho: Oh, and I also forgot to mention that the commit message should start with a prefix, in this case I think `SubmittingPatches: ` would be excellent.

PR #771, opened 2020-04-27
rebase: display an error if --root and --fork-point are both provided

newren:
CC: alexander.berg@atos.net


newren:
Looks like windows builds failed because we exceeded our collective time allotment of some sort, and the osx builds failed due to libintl finding rules.  So, I'll just ignore those and submit...

PR #775, opened 2020-05-01
git-p4.py: fix --prepare-p4-only error with multiple commits

seraphire:
When using git p4 submit with the --prepare-p4-only option, the program
should prepare a single p4 changelist and notify the user that more
commits are pending and then stop processing.

A bug has been introduced by the p4-changelist hook feature that
causes the program to continue to try and process all pending
changelists at the same time.

The function applyCommit should return True when applying the commit
was successful and the program should continue. In the case of the
--prepare-p4-only flag, the function should return False, alerting the
caller that the program should not proceed with additional commits.

Change the return value from True to False in the applyCommit function
when git-p4 is executed with --prepare-p4-only flag.

### DIFF
--- a/git-p4.py
+++ b/git-p4.py
@@ -2537,11 +2537,12 @@ class P4Submit(Command, P4UserMap):
                 ok = self.applyCommit(commit)
             if ok:
                 applied.append(commit)
-            else:
-                if self.prepare_p4_only and i < last:
-                    print("Processing only the first commit due to option" \
-                          " --prepare-p4-only")
+                if self.prepare_p4_only:
+                    if i < last:
+                        print("Processing only the first commit due to option" \
+                                " --prepare-p4-only")
                     break
+            else:
                 if i < last:
                     # prompt for what to do, or use the option/variable
                     if self.conflict_behavior == "ask":


seraphire:
On 5/6/2020 12:11 AM, gitgitgadget-git[bot] wrote:
>
> On the Git mailing list 
> <https://lore.kernel.org/git/xmqqwo5p918d.fsf@gitster.c.googlers.com>, 
> Junio C Hamano wrote (reply to this 
> <https://github.com/gitgitgadget/gitgitgadget/wiki/ReplyToThis>):
>
> |"Ben Keene via GitGitGadget" <gitgitgadget@gmail.com> writes: > From: 
> Ben Keene <seraphire@gmail.com> > > When using git p4 submit with the 
> --prepare-p4-only option, the program > should prepare a single p4 
> changelist and notify the user that more > commits are pending and 
> then stop processing. > ... When merged to 'pu', this seems to break 
> builds/tests at Travis CI.|

|I'll look into it. Thanks for the heads up.
|

||

dscho:
> |I'll look into it. Thanks for the heads up. | ||

As with previous PR comments: the Git mailing list does not see those. Please use the instructions in the ["reply to this" link](https://github.com/gitgitgadget/gitgitgadget/wiki/ReplyToThis) together with the [link to the email you are responding to](https://lore.kernel.org/git/xmqqwo5p918d.fsf@gitster.c.googlers.com) in order to reply to the original mail that only has been [mirrored here](https://github.com/git/git/pull/775#issuecomment-624431798).

seraphire:
I'm not sure how there are two branches, but the code I posted Saturday with commit https://github.com/git/git/commit/a554a65503e498ccecc43b379ed5363d6ccf0072 is the code that passes the tests.  The previous submission is the one that does not pass the tests.

I see that I didn't /submit so I will do that now.

APACGAMONDE:
Thank you!

APACGAMONDE:
Great thanks for all contributions. Follow please APACGAMONDE-Bank

PR #776, opened 2020-05-02
ci: allow skipping CI/PR builds on GitHub

dscho:
It was mentioned to me that it might not be totally helpful to run _all_ the builds whenever an in-progress branch is pushed to GitHub. For example, if a contributor has dozens of topic branches in flight, they might not want to have all of them built whenever a new end-of-day push happens.

This patch tries to address that, by following the convention that I believe Travis CI invented: if a commit message contains the needle `[skip ci]`, any CI build (i.e. a build triggered by a push) is skipped. Likewise, PR builds are skipped when the PR title and/or description contains `[skip pr]`.

Ideally, GitHub workflows will support this feature at some stage, but until then, we could have this (admittedly not very elegant) workaround.

Cc: Jeff King <peff@peff.net>, Jeff Hostetler <jeffhost@microsoft.com>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,6 +7,7 @@ env:
 
 jobs:
   windows-build:
+    if: "!contains(toJSON(github.event.commits.*.message), '[skip ci]') && !contains(toJSON(github.event.pull_request), '[skip pr]')"
     runs-on: windows-latest
     steps:
     - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


dscho:
I'll close this in favor of [`jk/ci-only-on-selected-branches`](https://github.com/gitgitgadget/git/commits/jk/ci-only-on-selected-branches).

PR #778, opened 2020-05-04
written in C,perl,Python

madilraza:


dscho:
@madilraza is this intended as a real code contribution? I cannot tell because it does not follow the style recommended in https://git-scm.com/docs/SubmittingPatches... If so, please adjust the style. You may want to add "Unix shell script" to the list in that case, too.

madilraza:
thank you for guideline i will update my mistakes

PR #784, opened 2020-05-13
merge-recursive: fix rename/rename(1to2) for working tree with a binary

newren:
rename/rename(1to2) is unique among our conflict resolutions in that it tries to put the as-merged-as-possible contents into two different paths, so I think this should be the only code path that needs an update for this issue.


newren:
Failed tests look like infrastructure failures (inability to download some sdk); so I'm going to ignore and submit.

PR #785, opened 2020-05-14
unpack-trees: also allow get_progress() to work on a different index

newren:
This is a fix to a minor 2.27 regression, from the en/sparse-checkout series.

Cc: jeffhost@microsoft.com

### DIFF
--- a/unpack-trees.c
+++ b/unpack-trees.c
@@ -333,10 +333,10 @@ static void load_gitmodules_file(struct index_state *index,
 	}
 }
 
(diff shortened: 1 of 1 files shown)


newren:
Not sure what's up with the cancelled linux-gcc job or the weird t5516 failure on OSX, so I just amended the commit (to get a new timestamp) and force pushed to trigger a rebuild/retest.

newren:
Okay, the only issues this time were clearly infrastructure related (failure to download some SDKs) so I'm going to ignore and submit.

PR #796, opened 2020-05-28
doc: ls-tree paths do not support wildcards

onlynone:


onlynone:
@dscho could you `/allow` me here?

PR #797, opened 2020-05-30
Anonymize URLs in the reflog

dscho:
This came up in an internal audit, but we do not consider this to be a big deal: the reflog is local and not really shared with anybody.

Changes since v1:

- Changed the if...else if...else cadence to move the `die()` to the last arm
- Stopped the memory leak of `display_repo` (allocated by `transport_anonymize_url()`)

### DIFF
--- a/builtin/clone.c
+++ b/builtin/clone.c
@@ -939,7 +939,7 @@ int cmd_clone(int argc, const char **argv, const char *prefix)
 {
 	int is_bundle = 0, is_local;
 	const char *repo_name, *repo, *work_tree, *git_dir;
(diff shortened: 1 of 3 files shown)


SHwareSystemsDev:
To me this should be a per-new-repo option, or if that's not enabled a per-commit-enabled-user option for those new repos, not a new behavior forced on all users after a client upgrade, to elect their commits get anonymized or not. For some the current behavior IS a feature, I'd think, not a bug.

dscho:
You misunderstand. This is about the reflog. And about avoiding to write secrets to reflogs.

PR #800, opened 2020-06-03
gitk: add option to perform 'git fetch' command

chiga17:
Hello everyone,

This is my first time trying to contribute.
Just want to add option to perform fetching to gitk tool.

BR,
Vladimir

### DIFF
--- a/gitk-git/gitk
+++ b/gitk-git/gitk
@@ -2,7 +2,7 @@
 # Tcl ignores the next line -*- tcl -*- \
 exec wish "$0" -- "$@"
 
-# Copyright Â© 2005-2016 Paul Mackerras.  All rights reserved.
+# Copyright © 2005-2016 Paul Mackerras.  All rights reserved.
 # This program is free software; it may be used, copied, modified
 # and distributed under the terms of the GNU General Public Licence,
 # either version 2, or (at your option) any later version.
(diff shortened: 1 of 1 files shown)


dscho:

Please review this part of the welcome message:

> Also, it is a good idea to review the commit messages one last time, as the Git project expects them in a quite specific form:
> 
>     * the lines should not exceed 76 columns,
> 
>     * the first line should be like a header and typically start with a prefix like "tests:" or "commit:", and
> 
>     * the commit messages' body should be describing the "why?" of the change.
> 
>     * Finally, the commit messages should end in a [Signed-off-by:](https://git-scm.com/docs/SubmittingPatches#dco) line matching the commits' author.

chiga17:
I've added new commit with updated commit message according to rules from "welcome message", but for some reason "GitGitGadget PR Handler" still checks first commit with 'too short commit message'.

webstech:
> I've added new commit with updated commit message according to rules from "welcome message", but for some reason "GitGitGadget PR Handler" still checks first commit with 'too short commit message'.

You should probably squash this into one commit.  From the reviewer's point of view, they only want to see the final result (although some might argue adding the function key could be a separate commit).  The first commit needs to be amended to include more than a commit message (the commit message body).

chiga17:
I've created new pull request with squashed commits:
https://github.com/git/git/pull/807

dscho:
For future reference: there is no need to open a new PR. Just use `git push --force` to update the branch.

PR #801, opened 2020-06-04
sparse-checkout: avoid staging deletions of all files

newren:
Changes since v1:
  * Fixed the ls command for portability, using the construct Junio suggested

CC: warmsocks@gmail.com, stolee@gmail.com


newren:
Cancelled?!?  Well, osx-clang passed, and one passing on that platform is good enough for me when the other dies for some unknown infrastructure reason.

dscho:
> Cancelled?!? Well, osx-clang passed, and one passing on that platform is good enough for me when the other dies for some unknown infrastructure reason.

That's probably an internal problem with the agents.

PR #804, opened 2020-06-08
improve git-diff documentation and A...B handling

chris3torek:
`git diff -h` help is succinct, but perhaps too much so.

The symmetric-diff syntax, `git diff A...B`, is defined by the documentation to compare the merge base of A and B to commit B.  It does so just fine when there is a merge base.  It compares A and B directly if there is no merge base, and it is overly forgiving of bad arguments after which it can produce nonsensical diffs.  It also behaves badly with other odd/incorrect usages, such as `git diff A...B C..D`.

The first patch simply adjusts a test that will fail if the second patch is accepted.  The second patch adds special handling for the symmetric and range diff syntax so that the option parsing works, plus a small test suite.  The third patch updates the documentation, including adding a section for combined commits, and makes the help output more verbose (to match the SYNOPSIS and provide common diff options like `git-diff-files`, for instance).

Changes since v3:
  - correct `>` / `>=` goof
  - fix test nit per Philip Oakley

### DIFF
--- a/Documentation/git-diff.txt
+++ b/Documentation/git-diff.txt
@@ -11,15 +11,17 @@ SYNOPSIS
 [verse]
 'git diff' [<options>] [<commit>] [--] [<path>...]
 'git diff' [<options>] --cached [<commit>] [--] [<path>...]
-'git diff' [<options>] <commit> <commit> [--] [<path>...]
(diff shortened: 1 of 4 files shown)


Hello87Kitty:
This must have been sent to me by mistake.

PR #808, opened 2020-06-16
Sparse checkout status

newren:
Some of the feedback of folks trying out sparse-checkouts at $dayjob is that sparse checkouts can sometimes be disorienting; users can forget that they had a sparse-checkout and then wonder where files went.  This series adds some output to 'git status' and modifies git-prompt slightly as an attempt to help.

(As per discussion on v1, a "git sparse-checkout [status|info|status]" might supplement the changes from this series later; this is just aimed at giving users a simple reminder that they are in a sparse checkout.)

Changes since v2:
  * Added an extra commit to document how git-prompt works with in-progress state of operations (rebase/merge/cherry-pick/revert/bisect), since I want to refer to it for comparison later in the docs.
  * As requested by SZEDER, added the ability to shorten or remove the information relating to sparsity state in the git-prompt.
  * Since there are now three options for whether and how to show information related to sparsity state, I put a fair amount of effort into reasoning among the options about what is a good default and why; I included this in the final commit message.  (While I suspect that we can easily change the default for the prompt in the future based on the big warning at the top of git-sparse-checkout.txt, having the rationale up-front for the original decision may help us out in the future.)

CC: dstolee@microsoft.com, szeder.dev@gmail.com

### DIFF
--- a/contrib/completion/git-prompt.sh
+++ b/contrib/completion/git-prompt.sh
@@ -74,6 +74,11 @@
 # revert, cherry-pick, or bisect, the prompt will include information
 # related to the operation, often in the form "|<OPERATION-NAME>".
 #
+# When the repository has a sparse-checkout, a notification of the form
+# "|SPARSE" will be included in the prompt.  This can be shortened to a
+# single '?' character by setting GIT_PS1_COMPRESSSPARSESTATE, or omitted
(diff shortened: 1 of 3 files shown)


snapback98:
Someone is gathering info on my phone behind the screen I find new stuff everyday

Get Outlook for iOS<https://aka.ms/o0ukef>

PR #815, opened 2020-07-03
Correctly escape @ of stashes

mbektchiev:
Autocomplete suggestions for stashes are broken due to `stash@`
being suggested without escaping.

Reproducible on `GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin19)`
and `macOS Catalina 10.15.5`.

### DIFF
--- a/contrib/completion/git-completion.bash
+++ b/contrib/completion/git-completion.bash
@@ -2998,13 +2998,15 @@ _git_stash ()
 			if [ $cword -eq 3 ]; then
 				__git_complete_refs
 			else
-				__gitcomp_nl "$(__git stash list \
-						| sed -n -e 's/:.*//p')"
+				__gitcomp_nl "$(__git stash list |
+						sed -n -e 's/:.*//p' |
+						sed 's/@/\\@/')"
 			fi
 			;;
 		show,*|apply,*|drop,*|pop,*)
-			__gitcomp_nl "$(__git stash list \
-					| sed -n -e 's/:.*//p')"
+			__gitcomp_nl "$(__git stash list |
+					sed -n -e 's/:.*//p' |
+					sed 's/@/\\@/')"
 			;;
 		*)
 			;;


atabaytebrizl:
3 Tem 2020 Cum 13:43 tarihinde Martin Bektchiev <notifications@github.com>
şunu yazdı:

mbektchiev:
Hi Junio,

Thank you for reviewing this patch! I've amended and pushed the branch and
you can find my comments inline:


> "Martin Bektchiev via GitGitGadget" <gitgitgadget@gmail.com> writes:
>
> > From: Martin Bektchiev <martin.bektchiev@progress.com>
> >
> > Autocomplete suggestions for stashes are broken due to `stash@`
> > being suggested without escaping.
>
> What is unclear to readers of this sentence is why '@' needs to be
> quoted in the first place.  '@' is not like '$' that has a syntactic
> significance to the shell.
>
> In fact, I just tried
>
>     $ git stash show stas<TAB>
>
> in a test repository where there is only one stash entry and got it
> completed to
>
>     $ git stash show stash@{0}
>
> and pressing <Enter> from this state did exactly what I expected to
> see.
>
> > Reproducible on `GNU bash, version 3.2.57(1)-release` and
> > `macOS Catalina 10.15.5`.
>
> What did you reproduce?  The completion gave me "stash@{0}" (without
> surrounding double quotes) in my above experiment?  If so, that does
> seem reproducible, but I do not see "suggestions ... are broken" here.

The problem is visible when you have more than one stash. If
you press TABm autocompletion stops at the `{` and if you then
press TAB once more for a list of all stashes instead of that
the completion changes to `stashstash{` and becomes broken at
this point.

>
> > @@ -2999,12 +2999,14 @@ _git_stash ()
> >  				__git_complete_refs
> >  			else
> >  				__gitcomp_nl "$(__git stash list \
> > -						| sed -n -e 's/:.*//p')"
> > +						| sed -n -e 's/:.*//p' \
> > +						| sed 's/@/\\@/')"
>
> This is not a new problem introduced by this patch, but we should
> get rid of these unnecessary backslashes and pipe at the beginning
> of the second line, both of which make the resulting code harder
> than necessary to read.  Ending the line with '|' pipe would give
> enough clue to the shell that you haven't finished talking to it,
> and you can continue to the next line without any backslashes just
> fine.

Fixed.

>
> >  			fi
> >  			;;
> >  		show,*|apply,*|drop,*|pop,*)
> >  			__gitcomp_nl "$(__git stash list \
> > -					| sed -n -e 's/:.*//p')"
> > +					| sed -n -e 's/:.*//p' \
> > +					| sed 's/@/\\@/')"
>
> Ditto.

Fixed.
>
> >  			;;
> >  		*)
> >  			;;
> >
> > base-commit: a08a83db2bf27f015bec9a435f6d73e223c21c5e
>


P.S. I responded with `git send-email` but since I still cannot see my email received it might have not been sent successfully. Hopefully you'll see my comment here ...

dscho:
No, the response will have to come by way of email. Nobody tackled the project yet where GitGitGadget mirrors PR comments back to the mailing list.

Did you get any mail from "MAILER-DAEMON" or some such, indicating why the list did not accept your mail? Or maybe your mail needs to be sent again or something?

mbektchiev:
I'm not sure what I'm doing wrong... Now I replied again -- this time from GMail as per [these instructions](https://github.com/gitgitgadget/gitgitgadget/wiki/ReplyToThis#if-you-use-webmail-gmail-and-friends) and still, nothing shows in the mailing list.

mbektchiev:
My responses still seem to not reach the mailing list... 😕 
I'm posting it here:

```
Ok, then I think it's best if I just close this PR. I didn't know that it worked alright in newer versions of BASH. And I haven't found a way to make it work without the escaping of @. It seems that the patch doesn't bring any value to the community after all...

Cheers and thanks once more for the review and insights!

Best wishes,
Martin
```

PR #822, opened 2020-07-26
Improvements to ref-filter

harry-hov:
This patch series intends to introduce some improvements and features to file ref-filter.[c/h] which mainly contains formatting logic for `git for-each-ref`.


kamolsrisan:
good

kamolsrisan:
good

PR #829, opened 2020-08-12
Fix handling of matching nonbare repo vs. matching something under nonbare repo

newren:
This fixes the issue reported at https://lore.kernel.org/git/CADjceoQxoL932W4mkfhG6VOgrQBhs9k6tXkWSkraKVPmUP+uCw@mail.gmail.com/, a regression in git-2.27.0 (that wasn't reported until after 2.28.0).  It involves running ls-files --others with globs when nonbare repositories are present.

Thanks to christian w for reporting, and Kyle Meyer for finding a way to simplify the testcase and bisecting.  Very helpful.

### DIFF
--- a/dir.c
+++ b/dir.c
@@ -1792,9 +1792,12 @@ static enum path_treatment treat_directory(struct dir_struct *dir,
 		nested_repo = is_nonbare_repository_dir(&sb);
 		strbuf_release(&sb);
 	}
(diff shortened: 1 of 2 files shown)


newren:
I don't care that "Gem::ConfigMap is deprecated", and my patch certainly didn't cause that, so I'll ignore that failing test.

PR #830, opened 2020-08-14
Extend and add a little more generalization to the mem_pool API

newren:
In my new merge algorithm, I made use of the mem_pool API in a few places...but I also needed to add a few more functions and also needed to make the API a bit more general.

Changes since v2:
  * Remove 'x' from mem_pool_xstr[n]?dup() names and remove check for NULL since mem_pool_alloc() already handles that (via xmalloc), as suggested by René
  * Rewrite mem_pool_strndup() to not rely on strnlen(), since that may be too new from some systems (comes from POSIX 2008 ).  Also pointed out by René

CC: Matheus Tavares <matheus.bernardino@usp.br>, Eric Sunshine <sunshine@sunshineco.com>, René Scharfe <l.s.r@web.de>

### DIFF
--- a/fast-import.c
+++ b/fast-import.c
@@ -526,14 +526,6 @@ static unsigned int hc_str(const char *s, size_t len)
 	return r;
 }
 
-static char *pool_strdup(const char *s)
-{
-	size_t len = strlen(s) + 1;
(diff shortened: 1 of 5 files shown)


lukaspupkalipinski:

PR #831, opened 2020-08-16
Clean up some memory leaks in and around dir.c

newren:
Some memory leaks in dir.c were making it hard for me to find my own leaks in merge-ort.  I followed a few threads and tried to both clean up the leaks and make the API less prone to incorrect use.

Changes since v1:
  * dropped the first patch, since Peff submitted an identical patch a day before me
  * implemented Peff's suggestions to rename dir_free() to dir_clear() and have it call dir_init() at the end so that folks can re-use the struct after dir_clear() if they so choose.


lukaspupkalipinski:

PR #832, opened 2020-08-17
ignore binary file diffs in git-p4 commit

berislavlopac:
Currently, when submitting commits, `git p4 submit` helpfully builds a diff of changes, and then displays them in a text editor before continuing (mimicking standard git commit behaviour). For changed files it asks git for the diff, but if a file is added it dumps all of its lines; to do so it opens a file, but as a text -- which obviously fails if a file is binary, raising `UnicodeDecodeError`.

This simple patch catches that exception and stops building the diff, so only the name of the added file is included.


berislavlopac:
Uh a mistake, apologies. :(

PR #833, opened 2020-08-17
ignore binary file diffs in git-p4 commit

berislavlopac:
Currently, when submitting commits, `git p4 submit` helpfully builds a diff of changes, and then displays them in a text editor before continuing (mimicking standard git commit behaviour). For changed files it asks git for the diff, but if a file is added it dumps all of its lines; to do so it opens a file, but as a text -- which obviously fails if a file is binary, raising `UnicodeDecodeError`.

This simple patch catches that exception and stops building the diff, so only the name of the added file is included.

### DIFF
--- a/git-p4.py
+++ b/git-p4.py
@@ -1847,9 +1847,13 @@ class P4Submit(Command, P4UserMap):
                 newdiff += "+%s\n" % os.readlink(newFile)
             else:
                 f = open(newFile, "r")
-                for line in f.readlines():
-                    newdiff += "+" + line
-                f.close()
+                try:
+                    for line in f.readlines():
+                        newdiff += "+" + line
+                except UnicodeDecodeError:
+                    pass
+                finally:
+                    f.close()
 
         return (diff + newdiff).replace('\r\n', '\n')
 


dscho:
>  Commit not signed off

@berislavlopac This is the issue GitGitGadget talked about when it said:

I also see that the commit message still needs to be prefixed by `git-p4:` (or `p4:` if you prefer), and that it needs to be wrapped at <= 74 columns/line.

berislavlopac:
Yes I can see that -- my apologies, I was a bit hasty when submitting the PR. I suppose the recommended approach is to close this PR and start another? Thank you! :blush:

dscho:
@berislavlopac oh, no worries! I invented GitGitGadget for the specific purpose of catching issues like this, also to help me prevent those things in my own contributions.

The recommended way to fix this is to use an [interactive rebase](https://hackernoon.com/beginners-guide-to-interactive-rebasing-346a3f9c3a6d), and then force-push (which will automatically update the PR).

berislavlopac:
Does anyone have an idea how to pass this? I'm pretty sure this failing `vs-build` check failure has nothing to do with this update...

dscho:
> Does anyone have an idea how to pass this? I'm pretty sure this failing `vs-build` check failure has nothing to do with this update...

Yep, the `vs-build` problem has been fixed in `master` (but not in `maint`). Just ignore it and `/submit` even if the build does not pass the `vs-build` job.

PR #834, opened 2020-08-18
[PATCH]v3 GitSVN: Multi line support of ignore-path, include-paths and skiping of empty commits

lukaspupkalipinski:
I used the ignore-paths option to ignore a lot of stuff I don’t need. The ignore pattern works well, but it could and up in empty commits. So just the message without any modifications / changes. The patch below skip a commit if all changes are ignored by the ignore-paths option. In order to use this feature I includes the option to read configuration for ignore-path, include-paths in several lines. So that the user is not limited by the max. char. per line definition.
In Addition this patch includes the optimizations which are mansion from your side.

Regarding the subrouties comments i oriented by this example: https://www.perlmonks.org/?node_id=29905


lukaspupkalipinski:

perl/Git/SVN/Fetcher.pm
```diff
@@ -142,32 +144,27 @@ sub is_path_ignored {
...
-			}
+    foreach $path ( keys %$paths ) {
+		$ignored = $self->is_path_ignored($path);
+		if ( !$ignored ) {
+			return 0;
 		}
-	}
-	return 1;
+    }
+    return 1;
```
dscho, line 167: There are no performance changes I see in this commit, but a lot of bogus white-space changes... And no sign-off, and the commit message does not really say what this is about. Please address this.

lukaspupkalipinski: Done see below. for Soem reason the whole file was had this issue.

dscho: Could you use an [interactive rebase](https://hackernoon.com/beginners-guide-to-interactive-rebasing-346a3f9c3a6d) to turn this into a clean patch? In the Git project, no patch series is accepted that introduces a bug that is fixed in the same patch series.

dscho:
I would highly recommend using the prefix `svn:` instead of the prefix `added:`, as the former conveys a much better piece of information.

Also, please continue in lower-case after the colon, e.g. `svn: add support for multi-line ignore-paths`.

Further, please to take care to wrap the commit messages at <=76 columns per line. The Git project is very particular about commit messages, and your  contribution will find a lot more interest when the commit messages are crafted with care, imitating the style of the existing commits (for example, I would leave out "as you know", that's more likely to irk reviewers than to amuse them).

perl/Git/SVN.pm
```diff
@@ -1188,6 +1188,22 @@ sub find_parent_branch {
...
+Input:  $path - array of strings (Paths) in a commit
+		$rev - Revision number
+
+Output: $log_entry if successfull
+		null if skipped
+		(die) on fetch error
+
+=cut
+
+############################################################
```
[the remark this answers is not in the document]
lukaspupkalipinski: Done. Please take a look.
thanks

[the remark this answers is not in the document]
dscho: Have you pushed the changed branch yet?

webstech:
Is there still interest in this change?  The user repo is gone (not that it is needed at this point).

dscho:
I guess @lukaspupkalipinski could reply to the contribution with a ping, potentially Cc:ing Eric Wong, the `git svn` maintainer.

If there is feedback that requires changes, the absence of the original fork will be a problem.

dscho:
I guess we should close this :-(

pddshk:
> Is there still interest in this change? The user repo is gone (not that it is needed at this point).

There is interest, I guess small, but I am interested in this. Maybe I can pick this up. I'll take a look in a while

pddshk:
I wonder if it is a good idea to craft 100+ lines regex if there is exact list of files to ignore and it is too large. In this case it may prove useful to read ignored files from file. And the last but not least: we can craft list of files to fetch ahead of time, before fetching and skip empty commits (should also be configurable) if the list is empty

bobibehi:
در تاریخ جمعه ۱۸ ژوئیه ۲۰۲۵، ۱۱:۴۶ Pavel Dydyshko ***@***.***>
نوشت:

PR #841, opened 2020-09-12
config: improve error message for boolean config

KlotzAndrew:
Currently invalid boolean config values return messages about 'bad
numeric', which I found misleading when the error was due to a
boolean string value. This change makes the error message reflect
the boolean value.

The current approach relies on `GIT_TEST_GETTEXT_POISON` 
being a boolean value, moving its special case out from 
`die_bad_number()` and into `git_config_bool_or_int()`. An
alternative could be for `die_bad_number()` to handle boolean
values when erroring, although the function name might need to
change if it is handling non-numeric values.

changes since v1
 - moved boolean error message change out of `git_config_bool_or_int`
to just in `git_config_bool` and added `die_bad_boolean` instead of
modifying `die_bad_number`. 

changes since v2
 - added a test for boolean config values
 - changed the condition to hit `die_bad_bool` from `if (0 <= v)` to `if (v < 0)`
 - removed change in get-text-poison.sh test

Signed-off-by: Andrew Klotz <agc.klotz@gmail.com>

cc: Jeff King <peff@peff.net>

### DIFF
--- a/config.c
+++ b/config.c
@@ -1180,6 +1180,20 @@ static void die_bad_number(const char *name, const char *value)
 	}
 }
 
+NORETURN
+static void die_bad_bool(const char *name, const char *value)
+{
+	if (!strcmp(name, "GIT_TEST_GETTEXT_POISON"))
+		/*
+		 * We explicitly *don't* use _() here since it would
+		 * cause an infinite loop with _() needing to call
+		 * use_gettext_poison().
+		 */
+		die("bad boolean config value '%s' for '%s'", value, name);
+	else
+		die(_("bad boolean config value '%s' for '%s'"), value, name);
+}
+
 int git_config_int(const char *name, const char *value)
 {
 	int ret;
@@ -1252,8 +1266,10 @@ int git_config_bool_or_int(const char *name, const char *value, int *is_bool)
 
 int git_config_bool(const char *name, const char *value)
 {
-	int discard;
-	return !!git_config_bool_or_int(name, value, &discard);
+	int v = git_parse_maybe_bool(value);
+	if (v < 0)
+		die_bad_bool(name, value);
+	return v;
 }
 
 int git_config_string(const char **dest, const char *var, const char *value)
--- a/t/t1300-config.sh
+++ b/t/t1300-config.sh
@@ -675,6 +675,13 @@ test_expect_success 'invalid unit' '
 	test_i18ngrep "bad numeric config value .1auto. for .aninvalid.unit. in file .git/config: invalid unit" actual
 '
 
+test_expect_success 'invalid unit boolean' '
+	git config commit.gpgsign "1true" &&
+	test_cmp_config 1true commit.gpgsign &&
+	test_must_fail git config --bool --get commit.gpgsign 2>actual &&
+	test_i18ngrep "bad boolean config value .1true. for .commit.gpgsign." actual
+'
+
 test_expect_success 'line number is reported correctly' '
 	printf "[bool]\n\tvar\n" >invalid &&
 	test_must_fail git config -f invalid --path bool.var 2>actual &&


config.c
```diff
@@ -990,21 +990,12 @@ NORETURN
 static void die_bad_number(const char *name, const char *value)
 {
 	const char *error_type = (errno == ERANGE) ?
-		N_("out of range") : N_("invalid unit");
-	const char *bad_numeric = N_("bad numeric config value '%s' for '%s': %s");
+		_("out of range") : _("invalid unit");
+	const char *bad_numeric = _("bad numeric config value '%s' for '%s': %s");
```
dscho, line 994: Maybe split this `N_()` -> `_()` change out into its own commit? It is a bit distracting from the concern(s) covered by the rest of the patch.

config.c
```diff
@@ -1097,7 +1088,17 @@ int git_config_bool_or_int(const char *name, const char *value, int *is_bool)
...
+		return v;
+
+	if (!strcmp(name, "GIT_TEST_GETTEXT_POISON"))
+		/*
+		 * We explicitly *don't* use _() here since it would
+		 * cause an infinite loop with _() needing to call
+		 * use_gettext_poison().
+		 */
+		die("bad boolean config value '%s' for '%s'", value, name);
+	die(_("bad boolean config value '%s' for '%s'"), value, name);
```
dscho, line 1101: It is really hard to spot the difference between these two `die()` calls without further guidance. The difference, of course, is that the `GIT_TEST_GETTEXT_POISON` variant does _not_ use translation, while the regular code path _does_.

Maybe there is a way to make this more obvious, e.g. by adding another comment, or using `else` here?

config.c
```diff
@@ -990,21 +990,12 @@ NORETURN
...
 		value = "";
 
-	if (!strcmp(name, "GIT_TEST_GETTEXT_POISON"))
-		/*
-		 * We explicitly *don't* use _() here since it would
-		 * cause an infinite loop with _() needing to call
-		 * use_gettext_poison(). This is why marked up
-		 * translations with N_() above.
-		 */
-		die(bad_numeric, value, name, error_type);
```
dscho, line 1006: It might make sense to move this code block _first_, in its own commit, and to make it more obvious why it needs to be moved, an explanation in the commit message would be nice.

config.c
KlotzAndrew: makes sense, moved to a separate commit with message

config.c
KlotzAndrew: updated the commit message and left in the first commit, what do you think?

dscho: I had another look, and only now realized that this block was not just moved within a function, but it was moved between functions. That's _kind_ of okay because we know that we only handle that environment variable via `git_config_bool_or_int()`.

But the move also implies that the [entire `switch` block](https://github.com/git/git/blob/v2.28.0/config.c#L1011-L1030) is skipped, which means that other calls to `git_config_bool_or_int()` won't report the origin of the config setting any longer, which is probably not so desirable.

I don't know that it makes sense to continue the review here, as I am not a gate keeper and the _real_ review should happen on the Git mailing list, anyway (i.e. _after_ `/submit`). But if I were to review this, I would suggest to look into extending the signature of `die_bad_number()` (and the entire call chain all the way up to `git_config_bool_or_int()`) to include a parameter indicating whether we asked for a numeric value, for a Boolean, or for either. It would blow up the `switch` block substantially, though.

As I said, it's probably a good idea to send this off to the Git mailing list. It would be good to change https://github.com/git/git/pull/841#issue-486022166 into a proper cover letter that contains the motivation for working on these changes, and the considerations that let you settle for the current solution, maybe soliciting suggestions what other approaches should be considered.

KlotzAndrew: thanks for the suggestion, adding the alternative to the PR message

sosbmw12345 (approved):

sosbmw12345 (approved):

PR #848, opened 2020-09-17
remote: in case of error don't show success message.

cschlack:
**Expected behaviour**
```
$ git remote set-head origin -a
error: Not a valid ref: refs/remotes/origin/master
```

**Actual behaviour**
```
$ git remote set-head origin -a
error: Not a valid ref: refs/remotes/origin/master
origin/HEAD set to master
```

I believe `... set to master` should not be shown in case of an error.
That's why I added an **else** to the if-condition.

Signed-off-by: Christian Schlack <christian@backhub.co>

### DIFF
--- a/builtin/remote.c
+++ b/builtin/remote.c
@@ -1356,7 +1356,7 @@ static int set_head(int argc, const char **argv)
 			result |= error(_("Not a valid ref: %s"), buf2.buf);
 		else if (create_symref(buf.buf, buf2.buf, "remote set-head"))
 			result |= error(_("Could not setup %s"), buf.buf);
-		if (opt_a)
+		else if (opt_a)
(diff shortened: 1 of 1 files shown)


cschlack:
Thanks @dscho

dscho:
If you look at [the history of that file](https://github.com/git/git/commits/c1ec8a0b3e5adc5fc732708c7851b2d91f27daab/builtin/remote.c) you will see that `bugfix` is not really used as an area. In this instance, I would say that `remote:` is a better prefix.

Please have a look at the commit history and see whether you want to imitate the style of those commit messages a bit more (the current commit message reads more like an email than the existing commit messages).

dscho:
Oh, and of course I _had_ to forget the most important thing: your patch makes a ton of sense to me! Thank you for contributing it.

scarpell (approved):

PR #864, opened 2020-10-06
userdiff: php: Add "final" and "abstract" to the list of function modifiers

phansys:
PHP allows some function modifiers that are not recognized in our current hunk header pattern

       final public function foo() { }
       abstract protected function bar() { }

Some context about the PHP semantics:

In PHP a class can be declared as abstract and final, while methods can be declared as final,
static and abstract, along with its visibility (public, protected and private).
Regarding classes, an abstract class can not be declared as final.
For the methods, from a strict syntax perspective, the order of the modifiers is irrelevant.

For more information, see:

* https://www.php.net/manual/en/language.oop5.abstract.php#example-213;
* https://www.php.net/manual/en/language.oop5.final.php#language.oop5.traits.static.ex1.

This commit adds "final" and "abstract" to the list of function modifiers.

### DIFF
--- a/t/t4018/php-abstract-method
+++ b/t/t4018/php-abstract-method
@@ -1,0 +1,7 @@
+abstract class Klass
+{
+    abstract public function RIGHT(): ?string
+    {
+        return 'ChangeMe';
+    }
+}
--- a/t/t4018/php-final-method
+++ b/t/t4018/php-final-method
@@ -1,0 +1,7 @@
+class Klass
+{
+    final public function RIGHT(): string
+    {
+        return 'ChangeMe';
+    }
+}
--- a/userdiff.c
+++ b/userdiff.c
@@ -143,7 +143,7 @@ PATTERNS("perl",
 	 "|=~|!~"
 	 "|<<|<>|<=>|>>"),
 PATTERNS("php",
-	 "^[\t ]*(((public|protected|private|static)[\t ]+)*function.*)$\n"
+	 "^[\t ]*(((public|protected|private|static|abstract|final)[\t ]+)*function.*)$\n"
 	 "^[\t ]*((((final|abstract)[\t ]+)?class|interface|trait).*)$",
 	 /* -- */
 	 "[a-zA-Z_][a-zA-Z0-9_]*"


dscho:
Thank you for your contribution! Before you send it off, could you amend the commit message to convince the reader that this patch improves Git (instead of repeating the diff)? I bet the reviewers on the Git mailing list are not all that familiar with PHP...

phansys:
Thank you!
I've updated the message. Please, let me know if it seems fine for you.

dscho:
>  I've updated the message. Please, let me know if it seems fine for you.

Looks good to me. `/submit` time, I guess?

phansys:
Forcing new build.

phansys:
In response to your comments, the first pattern is for methods, and the second is for classes.
In PHP a class can be declared as abstract and final, while methods can be declared as final, static and abstract, along with its visibility (public, protected and private).
An abstract class can not be declared as final.
For the methods, from a strict syntax perspective, the order of the modifiers is irrelevant.
For more information, see:
* https://www.php.net/manual/en/language.oop5.abstract.php#example-213;
* https://www.php.net/manual/en/language.oop5.final.php#language.oop5.traits.static.ex1.

I've added two new test cases, for the final and the abstract modifiers.
Please, let me know if they seem fine for you.

dscho:
> In response to your comments, the first pattern is for methods, and the second is for classes.
> In PHP a class can be declared as abstract and final, while methods can be declared as final, static and abstract, along with its visibility (public, protected and private).
> An abstract class can not be declared as final.
> For the methods, from a strict syntax perspective, the order of the modifiers is irrelevant.
> For more information, see:
> 
>     * https://www.php.net/manual/en/language.oop5.abstract.php#example-213;
> 
>     * https://www.php.net/manual/en/language.oop5.final.php#language.oop5.traits.static.ex1.
> 
> 
> I've added two new test cases, for the final and the abstract modifiers.
> Please, let me know if they seem fine for you.

Next time you send a new iteration, please update the [_first comment_](https://github.com/git/git/pull/864#issue-498215738) with this information; That comment is converted into the cover letter (see for example [the cover letter for your v2](https://lore.kernel.org/git/pull.864.v2.git.git.1602041898670.gitgitgadget@gmail.com)).

phansys:
> This patch series was integrated into seen via [5d397fa](https://github.com/git/git/commit/5d397fa9b5fb2d67811ff9e01c5231171791159d).

Thank you everyone for the support!
Should I close this PR now?

dscho:
It's only in `seen` so far. Next step is `next`, and after that it will enter `master`, at which point GitGitGadget will close the PR. If I were you, I'd do nothing :-)

PR #868, opened 2020-10-08
userdiff: PHP: catch "abstract" and "final" functions

JN89:
PHP permits functions to be defined like

       final public function foo() { }
       abstract protected function bar() { }

but our hunk header pattern does not recognize these decorations.
Add "final" and "abstract" to the list of function modifiers.

Helped-by: Johannes Sixt <j6t@kdbg.org>
Signed-off-by: Junio C Hamano <gitster@pobox.com>


webstech:
Resubmitting PR #864?

PR #873, opened 2020-10-10
Create cmake.yml

andysalvadory:

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -1,0 +1,46 @@
+name: CMake
+
+on: [push]
+
+env:
+  # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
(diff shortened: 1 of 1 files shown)


klo1991qa (approved):

dscho:
Thank you for opening this PR!

However, there are many issues with it:

- Git already has a CMake-based axis in its GitHub workflow.
- The CMake configuration targets only Windows so far, more work would be needed to support Linux and macOS.
- The commit message is way too short and not in the form requested in `SubmittingPatches`.
- The cover letter is missing. In fact, it is identical to the template text describing what the Git project asks for, suggesting that it has not even be read.

I *suspect* that this PR was opened to get a Hacktoberfest T-shirt. If that is the case, you can still get there, by diligently improving the issues I mentioned above.

PR #878, opened 2020-10-12
Make test selection easier by specifying description substrings instead of just numeric counters

newren:
This patch series enables me to change

   ./t9999-my-test.sh --run=1-3,5,17,19

into

   ./t9999-my-test.sh --run=setup,rename,symlink

and have it pick out tests with "setup", "rename", or "symlink" in their description and run those.  Saves me a lot of time, especially since numbers for tests aren't readily accessible.  The easiest way for me to get the numbers corresponding to the tests I want to run, is to run all the tests and look at the output to match up the descriptions with their numbers -- thus defeating the point of selecting just a subset of the tests to run in the first place.

Changes since v5:
  - Update comment to match the code (selectors are now only split on commas, not on whitespace)

cc: Jeff King <peff@peff.net>
cc: Andrei Rybak <rybak.a.v@gmail.com>

### DIFF
--- a/t/README
+++ b/t/README
@@ -258,25 +258,27 @@ For an individual test suite --run could be used to specify that
 only some tests should be run or that some tests should be
 excluded from a run.
 
-The argument for --run is a list of individual test numbers or
-ranges with an optional negation prefix that define what tests in
-a test suite to include in the run.  A range is two numbers
-separated with a dash and matches a range of tests with both ends
-been included.  You may omit the first or the second number to
-mean "from the first test" or "up to the very last test"
-respectively.
+The argument for --run, <test-selector>, is a list of description
+substrings or globs or individual test numbers or ranges with an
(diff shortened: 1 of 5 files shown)


newren:
Looks like CI is just automatically cancelling all runs and marking them failed, so I'm going to ignore it.  The basic testsuite passes locally and the change from the previous round is pretty small.  So...

newren:
Not much I can do about cancelled tests.  The test failure is due to a flaky test; I'll send a separate email about it.  (Besides, there were no code or test changes between v2 and v3.)  So...

dscho:
> The test failure is due to a flaky test; I'll send a separate email about it.

Excellent. I agree that just checking for `null` is flaky when the host name can contain that substring. Checking for `(null)` is most likely what that test needs to do instead.

PR #885, opened 2020-10-17
[Outreachy][Patch v3] t2200,t9832: avoid using 'git' upstream in a pipe

lkmandy:
Changes since v2:

- Optimized commit message

Signed-off-by: Amanda Shafack <shafack.likhene@gmail.com>

cc: Emily Shaffer <emilyshaffer@google.com>, Jonathan Nieder <jrnieder@gmail.com>
cc: Amanda Shafack <shafack.likhene@gmail.com>

### DIFF
--- a/t/t2200-add-update.sh
+++ b/t/t2200-add-update.sh
@@ -179,7 +179,8 @@ test_expect_success 'add -u resolves unmerged paths' '
 
 test_expect_success '"add -u non-existent" should fail' '
 	test_must_fail git add -u non-existent &&
-	! (git ls-files | grep "non-existent")
+	git ls-files >actual &&
+	! grep "non-existent" actual
 '
 
 test_done
--- a/t/t9832-unshelve.sh
+++ b/t/t9832-unshelve.sh
@@ -68,7 +68,8 @@ EOF
 		cd "$git" &&
 		change=$(last_shelved_change) &&
 		git p4 unshelve $change &&
-		git show refs/remotes/p4-unshelved/$change | grep -q "Further description" &&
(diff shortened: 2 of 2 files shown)


t/t2200-add-update.sh
```diff
@@ -179,7 +179,10 @@ test_expect_success 'add -u resolves unmerged paths' '
 
 test_expect_success '"add -u non-existent" should fail' '
 	test_must_fail git add -u non-existent &&
-	! (git ls-files | grep "non-existent")
+	! (
+		git ls-files >actual &&
+		grep "non-existent" actual
+	)
```
dscho, line 185: Welcome to the Git project! May I suggest one change that I think will improve your patch even further?

```suggestion
	git ls-files >actual &&
	! grep "non-existent" actual
```

lkmandy: @dscho  The code looks cleaner this way and I think it cuts out the redundancy in using the logical not operator. 
Thanks for this and I'll appreciate any more explanations in case I missed out in my understanding.

dscho: I would suggest this way to go forward:

- accept the change I proposed (there should be a button to commit it)
- pull the changes back into your local branch
- use `git rebase -i` to squash the change into the correct commit (change the third `pick` into `fixup`)
- force-push the branch to your fork
- `/submit` the second iteration of the patch series

I look forward to it!

lkmandy: I have made the requested changes. Thanks for your guidance

ZENNITTITTHZENTERPRISE:
Than for the mail.

PR #898, opened 2020-10-21
Support both merge backends in the testsuite, via environment variable

newren:
This patch series builds on top of en/dir-rename-tests.  It can be applied independent of my recent "new merge-ort API" series, but it is meant to use the same environment variable as patch 4 of that series (so if one of these two series updates that variable name, then they both should).

As promised, here is a series that makes the testsuite changes needed to simultaneously support both merge backends, keyed off a GIT_TEST_MERGE_ALGORITHM environment variable setting.  This series divides the changes up into commits based on the type of behavioral change (D/F conflicts, conflict marker labels, dirty file handling, etc.).  Because I haven't submitted the implementation of the merge-ort functions (it will start coming in my next series), none of these tests pass when GIT_TEST_MERGE_ALGORITHM=ort -- yet.  I am providing the tests (and merge-ort API) first because they provide a nice overview of the high-level differences in behavior of the two merge backends, and when that environment variable isn't set the old recursive algorithm is used and all the tests pass fine.  (If someone wants to see these tests pass with the new merge backend, they can check out the 'ort' branch of my repo, or wait for me to break up the changes into reviewable chunks.)

### DIFF
--- a/Documentation/technical/directory-rename-detection.txt
+++ b/Documentation/technical/directory-rename-detection.txt
@@ -18,7 +18,8 @@ It is perhaps easiest to start with an example:
 More interesting possibilities exist, though, such as:
 
   * one side of history renames x -> z, and the other renames some file to
-    x/e, causing the need for the merge to do a transitive rename.
+    x/e, causing the need for the merge to do a transitive rename so that
+    the rename ends up at z/e.
 
   * one side of history renames x -> z, but also renames all files within x.
     For example, x/a -> z/alpha, x/b -> z/bravo, etc.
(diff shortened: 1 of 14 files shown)


newren:
Ugh, I won't be able to use gitgitgadget to submit this, because it'll re-email all the en/dir-rename-tests patches.  I'll just use it for testing, I guess.

newren:
Actually, just go over to https://github.com/newren/git/pull/1 and /submit that one, and see if it works.  Then I don't have to worry about the en/dir-rename-tests stuff.

PR #907, opened 2020-10-30
fix gcc 4.0 big endian sha1

alex-free:
Currently, sha1.c is compiled in little endian mode when using gcc 4.0 on Mac OS X 10.4 PowerPC (which is big endian). This results in a build that always states there is a sha1 mismatch. The below patch allows big endian to be detected correctly using GCC 4.0 and probably older versions as well.

### DIFF
--- a/sha1dc/sha1.c
+++ b/sha1dc/sha1.c
@@ -43,6 +43,14 @@
    you will have to add whatever macros your tool chain defines to indicate Big-Endianness.
  */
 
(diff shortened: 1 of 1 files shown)


alex-free:
Thank you @dscho !

/submit

dscho:
Please don't forget [your Signed-off](https://git-scm.com/docs/SubmittingPatches#dco). It is required by the Git project.

PR #915, opened 2020-11-14
userdiff: PHP: improved test cases used to catch "abstract" and "final" functions

phansys:
Since the original tests cases were using other detectable keywords (like "public" or "protected"), they are able to pass even if the keywords intended to be covered are not present.
In the case of the test for "abstract" keyword, the method's body was removed in the fixture, since PHP doesn't allow an abstract method to have a body.


phansys:
Thank you Johannes.
The message was updated according your suggestion.

phansys:
Thanks for your feedback Junio.
Given your input, I think we should improve the way or the cases we are testing.
I try to share a brief about PHP methods:
1. functions inside classes are considered methods (although they are declared using the `function` keyword);
2. methods can have a visibility modifier: `public`, `protected` or `private`;
3. methods can be declared as `static` if they don't need the class to be instantiated;
4. methods can be declared as `final` in order to avoid them to be overridden by inherited classes;
5. methods declared in an abstract class can also be declared as `abstract` (they provide a contract like the methods declared in an interface, and they can't have a body);

Most of the facts between points 2 and 5 can be combined, except for some non logic cases (IE, 4 and 5, since there is no sense on declaring a contract on a method that can not be inherited).

IMO, we should provide separated cases for each individual scenario:
* `function`
* `public function`
* `protected function`
* `private function`
* `static function`
* `final function`
* `abstract function`

WDYT?

PR #917, opened 2020-11-19
fast-import: correctly handle multiple mark sets

mnauw:
Commit ddddf8d7e2 ("fast-import: permit reading multiple marks files",
2020-02-22) modified the insert_mark function to accept a specified
mark_set rather than the one-and-only global marks variable.

However, it did not take into account that the marks pointer value
may have to change due to list/tree expansion, which was not reflected
back to the input pointer parameter.  As a result, the list would
accumulate incorrectly.  As a side-effect, frequent allocation will
also occur, which would lead to a OOM kill or a thrashing system.

Signed-off-by: Mark Nauwelaerts <mark.nauwelaerts@gmail.com>

### DIFF
--- a/builtin/fast-import.c
+++ b/builtin/fast-import.c
@@ -150,7 +150,7 @@ struct recent_command {
 	char *buf;
 };
 
-typedef void (*mark_set_inserter_t)(struct mark_set *s, struct object_id *oid, uintmax_t mark);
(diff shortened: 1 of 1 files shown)


mnauw:
Oops ... looks like this was already fixed in 3f018ec716292b4d757385686f42f57af3bca685

dscho:
I was about to `/allow` you ;-)

mnauw:
Thanks for about to allow ;-)

PR #925, opened 2020-11-30
doc: make HTML manual reproducible

raboof:
This makes sure the generated id's inside the html version of the
documentation use the same id's when the same version of the
manual is generated twice.

Signed-off-by: Arnout Engelen <arnout@bzzt.net>


raboof:
Could somebody `/allow` me to submit this PR to the mailinglist?

raboof:
(submitted with `git send-email --to git@vger.kernel.org HEAD^` instead, https://public-inbox.org/git/20201201095037.20715-1-arnout@bzzt.net/T/#u)

PR #928, opened 2020-12-04
Support having non-utf-8 characters returned by p4

bassdr:
When perforce server is not configured for Unicode, and commands like `p4 users` returns a string with non-ascii characters (eg. one of the user's FullName has a french `é` in it), git-p4 was giving `Exception: failure accessing depot: could not connect`.
With this patch, if such character is encountered, it will honor the new `git-p4.textEncoding` config option, or silently replace the erronous character and continue.


dscho:
@bassdr would you mind wrapping the commit message at 76 columns/line, and adding your [sign-off](https://git-scm.com/docs/SubmittingPatches#sign-off)?

dscho:
Also, please edit [the first comment](https://github.com/git/git/pull/928#issue-532795393), as it will be sent as cover letter (inlined for a single patch), and you don't want the PR template to be sent to the Git mailing list.

Finally, you will want to Cc: Luke Diamand [just like was done here](https://github.com/gitgitgadget/git/pull/782#issue-517202609).

bassdr:
@dscho thank you for your feedback, I think I addressed your comments. Let me know if I need to do something else. Thanks

dscho:
Looks good to me!

bassdr:
By the way, this patch is good enough for my use case, but if we want to go one step further regarding character handling, we could:
 - check the server configuration: I think `p4 counter Unicode` will return `0` or `1`, otherwise `p4 info` will contain such information
    - If not Unicode, then we force `P4CHARSET=none` and use my approach
    - if Unicode, then set `P4CHARSET=utf8` and let p4 do the conversion as it knows best what to do.

dscho:
> 
> 
> By the way, this patch is good enough for my use case, but if we want to go one step further regarding character handling, we could:
> 
>     * check the server configuration: I think `p4 counter Unicode` will return `0` or `1`, otherwise `p4 info` will contain such information
>       
>       * If not Unicode, then we force `P4CHARSET=none` and use my approach
>       * if Unicode, then set `P4CHARSET=utf8` and let p4 do the conversion as it knows best what to do.

Maybe you want to add that reply to the mailing list thread?

dscho:
> @dscho Yes, how do we do this? reply to the email directly?

Yes.

QueenMax:
in trouble

ryujinxx (approved):

PR #930, opened 2020-12-08
Document 'T' status code for git status

jairbubbles:
Git status can return 'T' status code which stands for "typechange". I can't document more the behavior but it would have helped me a lot to see that line in the documentation so I guess it can help others too.

cc: Jeff King <peff@peff.net>
cc: Julien Richard <julien.richard@ubisoft.com>


jairbubbles:
@dscho I fixed my public email, should it be enough?

dscho:
> @dscho I fixed my public email, should it be enough?

I think so, yes.

PR #944, opened 2021-01-09
gitk: check main window visibility before waiting for it to show

tpietzsch:
If the main window is already visible when gitk waits for it to
become visible, gitk hangs forever.

This commit adds a check whether the window is already visible.
See https://wiki.tcl-lang.org/page/tkwait+visibility

The symptom of the for-ever waiting is that gitk shows just an
empty window, like this: https://tinyurl.com/yxr5qlf6
It happens (to me) on macOS 11.1 with git-gui installed via
homebrew. It doesn't happen on macOS 10.14.


leewillis77:
Confirming that this patch fixes the issues I've had with git-gui installed via homebrew.

mnpoonia:
I have the patch applied and still face the issue.

cheater:
experiencing this issue in 2.43.1, or at least the error message is the same.

on top of this we need to figure out who broke this and how, because breaking critical dev tools is bad and should be precluded by a corrected contribution process.

dscho:
> we need to figure out who broke this and how, because breaking critical dev tools is bad and should be precluded by a corrected contribution process.

@cheater no, we do not want this, at least not with this attitude. Especially after reading [this article](https://www.404media.co/xz-backdoor-bullying-in-open-source-software-is-a-massive-security-vulnerability/), I will not tolerate bullying of volunteers who donate their free time to contribute to open source, for the most part. You may want to edit your comment accordingly.

cheater:
@dscho this is not about bullying anyone. Please don't read bullying into this. This is about fixing the root issue, which is a situation where changes can be done without finding out that something important has broken.

No one wants to bully anyone here. As I said by the end of the text you quoted, "should be precluded by a corrected contribution process". for a simple example, "if you work on something that affects gitk, check if it still works on all major platforms". etc. this is a medium-simple check, but it seems to not have been done: running gitk on affected systems shows a blank window (so it's immediately obvious that something's broken).

so, again, not about bullying anyone, it's about finding out who did the work, what they tried, what they didn't try, and what should be tried next time around that was missed the last time.

cheater:
also reproducible on 2.44.0 by the way

cheater:
so the few questions you might ask are:
- can this testing be automated?
- does git need a manual testing process for releases that include changes to the GUI tools or the text interface?
- do we need a manual testing playbook?
- what platforms should and shouldn't be supported?
- how do we find out that a change *should* trigger such a testing process?

cheater:
but in the end, _someone_ did break the thing. it's only fair to ask them what happened, which will let us learn for the future. contributing to important code bases brings some responsibility with it and one of those seems to be "hey X, this commit you made broke something, can you tell us what happened?" this isn't bullying, it's just professional communication about an issue being encountered by end users.

cheater:
specifically the article you brought up happened _because_ someone asked that question, and because that someone found out _who_ did this. for all you know, given how broken it is right now, gitk might now be stealing ssh keys. so we need to look into this issue, again: find out who committed this, possibly look at other commits from that author, etc. this is not bullying any more than the guy who found the xz backdoor was bullying a state actor.

cheater:
(but just to be 100% clear i do not _ascribe_ any malice here at all, i just think we need to figure out what happened, that's all)

dscho:
Defending that off-putting an entitled behavior makes it actually worse, not better.

PR #950, opened 2021-01-20
mergetool: use mergetool_find_win32_cmd from git-mergetool--lib.sh to locate exe

michaelcompressconsult:
kdiff3 works fine on windows - with the exception that it cannot be found because of git subsystem not searching windows environment.

Adding these 3 lines makes it work on windows using windows kdiff3 too.
The 'which' should capture the unix/linux world.

Explicit set paths will cause this code not to be executed, see git/git-mergetool--lib.sh
(Also noticed there but different issue: the code in is_available() does not consider git settings like get_merge_tool_path() does, so maybe the later should be called in is_available to get an accurate list of available tools?)

kdiff3 can be found here:  http://kdiff3.sourceforge.net/

### DIFF
--- a/mergetools/kdiff3
+++ b/mergetools/kdiff3
@@ -27,5 +27,5 @@ exit_code_trustable () {
 }
 
 translate_merge_tool_path() {
-                mergetool_find_win32_cmd "kdiff3.exe" "kdiff3"
+    which kdiff3 || mergetool_find_win32_cmd "kdiff3.exe" "Kdiff3"
 }
--- a/mergetools/kdiff3
+++ b/mergetools/kdiff3
@@ -27,5 +27,5 @@ exit_code_trustable () {
 }
 
 translate_merge_tool_path() {
-                mergetool_find_win32_cmd "kdiff3.exe" "Kdiff3"
+                mergetool_find_win32_cmd "kdiff3.exe" "kdiff3"
 }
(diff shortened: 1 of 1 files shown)


dscho:
@michaelcompressconsult please review the existing commit messages in https://github.com/git/git/commits/HEAD/mergetools/kdiff3 and follow their example re: real email address, ["Signed-off-by:" footer](https://git-scm.com/docs/SubmittingPatches#sign-off) and verbosity of the commit message.

Also, please squash these 3 commits into a single one: the Git project frowns on introducing incorrect changes and immediately resolving the issues within the same patch series.

michaelcompressconsult:
anyone an idea how to fix the problem? i made my email public alredy...

michaelcompressconsult:
thx, just saw the /allow

dscho:
> i made my email public alredy...

I don't see it in https://github.com/michaelcompressconsult:

image

michaelcompressconsult:
made a new pull request #952 to squash commit to one commit.

dscho:
You can just force-push to update the PR. No need to open several PRs.

michaelcompressconsult:
As there were 3 commits and the original was made on github web interface edits (where there is no way to sign commits and squash) i redid the 3 lines on a real checkout on a new branch.  Which of course means a new pull request.
Had to do it at home - in office just web interface is possible.

dscho:
FWIW you can easily fetch and check out branches locally, even branches that have been created in the web UI.

PR #952, opened 2021-01-26
mergetools/kdiff3: make kdiff3 work on windows too

michaelcompressconsult:
**mergetools/kdiff3: make kdiff3 work on windows too**

Currently the native kdiff3 mergetool is not found by git mergetool on windows.
The message "The merge tool kdiff3 is not available as 'kdiff3'" is displayed.

But it is important especially for GUI to use this native version on windows.
Kdiff3 for various systems can be downloaded from https://download.kde.org/stable/kdiff3/

Bug cause:
On Windows the executable name has to be translated (kdiff3.exe instead of kdiff) and the windows path has to be searched - similar to winmerge.

Fix:
This change is using mergetool_find_win32_cmd from the library in the translate_merge_tool_path().
This is done the same way as in winmerge.

However this translation must not be made on linux/unix, so a "type kdiff3" test is made on kdiff3 and only if not found the windows search is tried. 

Signed-off-by: Michael Schindler <michael@compressconsult.com>

### DIFF
--- a/mergetools/kdiff3
+++ b/mergetools/kdiff3
@@ -25,3 +25,12 @@ merge_cmd () {
 exit_code_trustable () {
 	true
 }
+
+translate_merge_tool_path() {
+	if type kdiff3 >/dev/null 2>/dev/null
+	then
+		echo kdiff3
+	else
+		mergetool_find_win32_cmd "kdiff3.exe" "Kdiff3"
+	fi
+}


dscho:
Please edit the PR description; It will be sent as cover letter.

michaelcompressconsult:
I give up getting this into git distribution.
"Please edit the PR description; It will be sent as cover letter."
WHAT should i put WHERE? where is a "pull record description?" 

I keep adding these lines to the git installations on my local windows machine so it will start.

dscho:
> "Please edit the PR description; It will be sent as cover letter."

This means that you are asked to hit the <kbd>...</kbd> button on the top of https://github.com/git/git/pull/952#issue-561983954, and to select <kdb>Edit</kdb>: 



Right now, this pull request description simply repeats the commit message, which is not really helpful to understand the context of your submission. You could explain e.g. that you are using kdiff3 on Windows and how it fails before this patch.


> Please rebase the branch and force-push.

This means that GitGitGadget asks you to run essentially these commands:

```console
$ git pull --rebase https://github.com/git/git master
$ git push --force https://github.com/michaelcompressconsult/git kdiff3_win
```

(Make sure that your current branch _is_ called `kdiff3_win`.

mergetools/kdiff3
```diff
@@ -28,4 +28,4 @@ exit_code_trustable () {
 
 translate_merge_tool_path() {
     which kdiff3 || mergetool_find_win32_cmd "kdiff3.exe" "Kdiff3"
-}
\ No newline at end of file
+}
```
dscho, line 31: You will want to squash this into the previous patch that needed to be fixed up. Learn how to do that e.g. [here](https://blog.axosoft.com/learning-git-interactive-rebase/), using an _interactive rebase_.

mergetools/kdiff3
```diff
@@ -25,3 +25,7 @@ merge_cmd () {
 exit_code_trustable () {
 	true
 }
+
```
dscho, line 28: Commenting on this line, as I cannot comment directly on the commit message:

- The message claims that kdiff3 is available for Windows, but it does not provide any download link. Maybe you want to add that?
- When explaining that the executable name has to be translated, maybe mention which code you imitated to fix this.
- There is a typo in "_thse_ operating systems"
- I seem to remember that `type -p kdiff3 >/dev/null 2>&1` was preferred over `which kdiff3`, can you compare to other merge tools?

mergetools/kdiff3
```diff
@@ -25,3 +25,12 @@ merge_cmd () {
...
+translate_merge_tool_path() {
+	if type -p kdiff3 >/dev/null 2>/dev/null
```
gitster, line 30: Existing backend interfaces like bc and tortoisemerge uses a more portable "type" (without -p) for this purpose.  Is there a reason why this needs to use "-p"? If not, drop it.

dscho: My fault, I suggested it to replace `which`...

gitster: I have a mild prejudice against "whcih", which I found has been less reliable at least in the past (e.g. on Solaris I saw it exited with 0 status while giving "no such command" message, back when I still had access to several of them).

ryujinxx (approved):

mergetools/kdiff3
michaelcompressconsult: changed to type (without -p).
Originally i had which because type has a totally different mening in a windows environment. If mergetool is called in a windows context, for example in backticks in a perl started from windows, the windows type command (= cat in linux/unix) will be called. However windows "type kdiff3" will also fail with nonzero exit (file not found), so everything is fine.

dscho: `type` is a built-in of the Bash we're using to interpret the `mergetool` script, so there is no danger whatsoever that your command will run `type.exe` by mistake.

suriyo-cycron:
#1035

suriyo-cycron:
#952

PR #956, opened 2021-02-05
fix invalid unsigned arithmetic.

ihsinme:
I believe your checks are not correct.
in my opinion they are equivalent to checks !=.
I suggest a simple fix.


dscho:
Please have a look at https://github.com/git/git/blob/v2.30.0/Documentation/CodingGuidelines, or at least at other commits in Git's history to see in what format we are expecting commit messages to be.

It would probably be good to wave around some C standard in that commit message which either says that this is undefined behavior, or that it _is_ cast to `unsigned`.

ihsinme:
I looked at the code again.
in your case `ita_nr` is guaranteed to be less than or equal to `active_nr`.
this saves you the worry of such a transformation.

Excuse for troubling.

PR #967, opened 2021-02-19
git-p4 python3 latin1 decode support

Raffaello:

### DIFF
--- a/git-p4.py
+++ b/git-p4.py
@@ -181,7 +181,13 @@ def prompt(prompt_text):
 if bytes is not str:
     # For python3, always encode and decode as appropriate
     def decode_text_stream(s):
-        return s.decode() if isinstance(s, bytes) else s
+        if not isinstance(s, bytes):
+            return s
+        try:
+            return s.decode()
+        except UnicodeDecodeError as e:
+            print(e)
+            return s.decode('iso-8859-1')
     def encode_text_stream(s):
         return s.encode() if isinstance(s, str) else s
 else:


dscho:
>  it is a good idea to review the commit messages one last time, as the Git project expects them in a quite specific form:
> 
> * the lines should not exceed 76 columns,
> 
> * the first line should be like a header and typically start with a prefix like "tests:" or "commit:", and
> 
> 

@Raffaello please follow this guidance.

git-p4.py
```diff
@@ -181,7 +181,13 @@ def prompt(prompt_text):
...
     # For python3, always encode and decode as appropriate
     def decode_text_stream(s):
-        return s.decode() if isinstance(s, bytes) else s
+        if not isinstance(s, bytes):
+            return s
+        try:
+            return s.decode()
+        except UnicodeDecodeError as e:
+            print(e)
+            return s.decode('iso-8859-1')
```
gitster, line 190: It is not clear why latin-1 has to be so special. Perhaps there should be a way from the command line or per-repository configuration to specify "if the text does not .decode() correctly, try _this_ encoding as a fallback"?

Also, shouldn't an exception while decoding as latin-1 be caught and handled?

Raffaello: > It is not clear why latin-1 has to be so special. Perhaps there should be a way from the command line or per-repository configuration to specify "if the text does not .decode() correctly, try _this_ encoding as a fallback"?

UTF-8 is required for git and is a superset of ASCII, but not of latin-1 (which is a different superset of ASCII).
Latin-1 defines characters in the range 128-255 and these cause problems when interpreted as UTF-8.

Yes that as a fallback, as if it is a latin1 encoded file it will just raise an exception. 
This patch a specific problem, it is not a general solution, but this it is a starting point to address a problem in git-p4 with python3. 

A proper could be detecting the encoding of the file, but it was far larger complex that my personal scope problem.

> Also, shouldn't an exception while decoding as latin-1 be caught and handled?

In this case no as there is no recovery for that, so it will just let propagate the exception.

gitster: By "why is latin-1 so special", I didn't mean "latin-1 as opposed to utf-8". Among latin-2, iso-2022-jp, etc., there is nothing that makes latin-1 particularly special and needs to be treated specially, leaving users who have similar "My source repository is not using UTF-8" problem still unhappy.

Raffaello: oh well, it is just because I needed that specific decode charset and the "default" one in python is throwing an exception.
As I am not fully aware of all the other possibilities, I have just started this PR highlighting an edge case scenario where git-p4 will crash.

I am not experienced enough to cover all the uses cases:

If wouldn't be possible to have an auto-detect charset encoding, 
probably a better approach would be to catch that exception and ask user input for the proper encoding type rather then guessing it?

PR #976, opened 2021-03-06
Fix all leaks in t0001

ahunt:
This series fixes all the memory leaks that can cause t0001 to fail when run with LeakSanitizer (t0000 already passes without failures).

I suspect that none of these leaks had any user impact, and I'm aware that every change does cause some noise - I would have no complaints abandoning this series if it's not deemed valuable enough.

Note: this series does not guarantee that there are no leaks in t0000-t0001. It only fixes those leaks that are causing test failures. There is at least one test case in t0000 where git is invoked in a subshell, and the return value is ignored - meaning that a memory leak during that occurs during that invocation does not cause tests to fail:
https://git.kernel.org/pub/scm/git/git.git/tree/t/t0000-basic.sh#n1285

In case anyone is interested: I have been using the following workflow to find leaks and verify fixes - when using LSAN standalone I run into crashes in the leak-checking phases, therefore I'm using full ASAN instead (LSAN standalone mode is known to be less-well tested than leak-checking within ASAN [1], so this isn't a huge surprise or concern):

make CC="ccache clang" GIT_TEST_OPTS="-i -v" DEFAULT_TEST_TARGET="t0000-basic.sh" ASAN_OPTIONS="detect_leaks=1:abort_on_error=1" SANITIZE=address DEVELOPER=1 CFLAGS="-DSUPPRESS_ANNOTATED_LEAKS -g -fno-optimize-sibling-calls -O1 -fno-omit-frame-pointer" test

(I'm also rerunning the entire test suite with ASAN but with leak-checking disabled in order to gain some confidence that my fixes aren't inadvertently introducing memory safety issues.)

[1] https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer#stand-alone-mode

### DIFF
--- a/builtin/clone.c
+++ b/builtin/clone.c
@@ -967,7 +967,7 @@ int cmd_clone(int argc, const char **argv, const char *prefix)
 	char *path, *dir, *display_repo = NULL;
 	int dest_exists, real_dest_exists = 0;
 	const struct ref *refs, *remote_head;
-	const struct ref *remote_head_points_at;
+	const struct ref *remote_head_points_at = NULL;
 	const struct ref *our_head_points_at;
 	struct ref *mapped_refs;
 	const struct ref *ref;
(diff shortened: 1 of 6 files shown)


ahunt:
(Closing this PR because I accidentally created it against git/git instead of gitgitgadget/git.)

dscho:
> 
> 
> (Closing this PR because I accidentally created it against git/git instead of gitgitgadget/git.)

Both should work, though...

PR #981, opened 2021-03-17
replace test -f with test_path_is_file

krush11:
changes: replaced `test -f`  with `test_path_is_file`

reason: `test -f` and `test_path_is_file` have same functionality but `test_path_is_file` gives better debugging information on error

### DIFF
--- a/git-rebase--preserve-merges.sh
+++ b/git-rebase--preserve-merges.sh
@@ -101,7 +101,7 @@ write_basic_state () {
 }
 
 apply_autostash () {
-	if test -f "$state_dir/autostash"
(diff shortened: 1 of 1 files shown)


dscho:
> 
> 
> There is an issue in commit [ca2e8db](https://github.com/git/git/commit/ca2e8dbf6eb8d57803adad712f2ddddfd476a6c8):

Please have a look at the commit history via `git log -Gtest_path_is_file t/` and imitate the commit messages' style.

dscho:
@krush11 there is no need to close this PR, you can fix the commit and force-push.

PR #982, opened 2021-03-17
replace test -f with test_path_is_file

krush11:
Although  has the same functionality as test_path_is_file(), in
the case where test_path_is_file() fails, we get much better debugging
information.

Replace  with test_path_is_file so that future developers
will have a better experience debugging these test cases.

### DIFF
--- a/t/t7300-clean.sh
+++ b/t/t7300-clean.sh
@@ -28,15 +28,15 @@ test_expect_success 'git clean with skip-worktree .gitignore' '
 	mkdir -p build docs &&
 	touch a.out src/part3.c docs/manual.txt obj.o build/lib.so &&
 	git clean &&
-	test -f Makefile &&
-	test -f README &&
-	test -f src/part1.c &&
-	test -f src/part2.c &&
+	test_path_is_file Makefile &&
+	test_path_is_file README &&
+	test_path_is_file src/part1.c &&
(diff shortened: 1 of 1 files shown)


dscho:
@krush11 you need to sign off using your real name (see https://git-scm.com/docs/SubmittingPatches#sign-off):

> Also notice that a real name is used in the `Signed-off-by` trailer. Please don’t hide your real name.

krush11:
@dscho are there any further changes required before this is ready for merge?

dscho:
> 
> 
> @dscho are there any further changes required before this is ready for merge?

Looks good to me! Time to `/submit`.

dscho:
If you keep deleting your fork, you won't be able to submit new revisions of the patch series, frustrating everyone. Maybe keep your fork open until the patch series is merged into `master`, addressing feedback and providing new iterations every couple of days as needed?

krush11:
sry my bad, I am new to the contributing styles of git. Will try to avoid deletions from now on.

dscho:
> 
> 
> sry my bad, I am new to the contributing styles of git. Will try to avoid deletions from now on.

To be clear, this has nothing to do with contributing to Git per se. If you delete your fork, _any_ PR associated with a branch in that fork will be closed. That goes for every fork you have, no matter which project.

PR #988, opened 2021-03-24
sequencer: fix edit handling for cherry-pick and revert messages

newren:
Changes since v2:
  * Changed to use <0 for unspecified (instead of == -1), >0 for true (instead of == 1).
  * Removed assert() statement.
  * Removed unnecessary "want_edit" local variable

Reported-by: Renato Botelho <garga@freebsd.org>
Signed-off-by: Elijah Newren <newren@gmail.com>

### DIFF
--- a/builtin/revert.c
+++ b/builtin/revert.c
@@ -182,7 +182,7 @@ static int run_sequencer(int argc, const char **argv, struct replay_opts *opts)
 				"--signoff", opts->signoff,
 				"--no-commit", opts->no_commit,
 				"-x", opts->record_origin,
-				"--edit", opts->edit,
+				"--edit", opts->edit > 0,
 				NULL);
 
 	if (cmd) {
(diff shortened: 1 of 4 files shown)


t0p0lino:
Hi

jammyto10:
What is this email about because I cant get on Gitgit help me please thanks 

t Daniel
 
 
  On Wed, 31 Mar 2021 at 19:06, ***@***.***> wrote:   


On the Git mailing list, Junio C Hamano wrote (reply to this):
Elijah Newren ***@***.***> writes:

> ... would you two prefer
> that I add a new affirm() function that is ALSO compiled out in
> production?

The reason I do not think affirm would fly is because it shares the
most problematic trait with assert(): it can be comipled out.

There was at least one bug we found and fixed in a piece of code
that came later in a codepath that had an assert() in it.  The buggy
code (IIRC, it was added way after the assert() was written in an
unrelated patch) was unknowingly depending on a side-effect of an
innocuous function call (i.e. assert(func_tion(args) == expected))
made in the assert condition, and the bug did not reproduce in the
developer's environment.

PR #996, opened 2021-04-09
doc: clarify the filename encoding in git diff --name-only and --name-status

hexagonrecursion:
AFAICT parsing the output of `git diff --name-only master...feature`
is the intended way of programmatically getting the list of files modified
by a feature branch. It is impossible to parse text unless you know what
encoding it is in. The output encoding of diff --name-only and
diff --name-status was not documented.

I asked on the mailing list and got this:
https://public-inbox.org/git/YGx2EMHnwXWbp4ET@coredump.intra.peff.net/
> There's some discussion in Documentation/i18n.txt, which is included in
various manpages (e.g., https://git-scm.com/docs/git-log#_discussion)
but it doesn't seem to be mentioned in git-diff.
>
>The short answer is: mostly utf8, but historically on platforms that
don't care (like Linux) you could get away with other encodings.
>
> -Peff

My takeaway was to always parse it as utf8 regardless of platform or
environment.

Changes since v1:
- Replace "always" with "usually"
- Add a link to https://git-scm.com/docs/git-log
- Replace "usually" with "often"

### DIFF
--- a/Documentation/diff-options.txt
+++ b/Documentation/diff-options.txt
@@ -293,11 +293,14 @@ explained for the configuration variable `core.quotePath` (see
 linkgit:git-config[1]).
 
 --name-only::
-	Show only names of changed files.
+	Show only names of changed files. The file names are often encoded in UTF-8.
+	For more information see the discussion about encoding in the linkgit:git-log[1]
+	manual page.
 
 --name-status::
 	Show only names and status of changed files. See the description
 	of the `--diff-filter` option on what the status letters mean.
+	Just like `--name-only` the file names are often encoded in UTF-8.
 
 --submodule[=<format>]::
 	Specify how differences in submodules are shown.  When specifying


Documentation/diff-options.txt
```diff
@@ -293,11 +293,13 @@ explained for the configuration variable `core.quotePath` (see
 linkgit:git-config[1]).
 
 --name-only::
-	Show only names of changed files.
+	Show only names of changed files. The output is always encoded in
+	UTF-8 if this switch is present.
```
gitster, line 298: There is nothing that guarantees that "always".

As the page referred to in Peff's message says, "Note that Git at the core level treats path names simply as sequences of non-NUL bytes, there are no path name encoding conversions (except on Mac and Windows).",
it depends on platform's encoding, and hence depends on the project that recorded the history. They may have used legacy UNIX and recorded their trees in ISO-8859-1 (Latin-1), and the project's repository would work perfectly find among the participants of the project. Or they originated the project on more modern Linux and happened to have chosen UTF-8.

The same comment applies to the change in the other hunk.

gitster (changes requested):
Thanks for trying to improve our doc, but anything that gives a false sense of guarantee in these descriptions are unwelcome. Perhaps replacing "always" with "often" would give the most practical help to many users who are on Windows and Macs.

PR #997, opened 2021-04-09
Solving 2 Error documentations

NishantJoshi00:
This is to solve the problem with the errors where the documentation was not properly followed.

### DIFF
--- a/Documentation/RelNotes/2.13.0.txt
+++ b/Documentation/RelNotes/2.13.0.txt
@@ -20,7 +20,7 @@ Backward compatibility notes.
 
  * Git now avoids blindly falling back to ".git" when the setup
    sequence said we are _not_ in Git repository.  A corner case that
-   happens to work right now may be broken by a call to die("BUG").
+   happens to work right now may be broken by a call to die("bug").
    We've tried hard to locate such cases and fixed them, but there
    might still be cases that need to be addressed--bug reports are
    greatly appreciated.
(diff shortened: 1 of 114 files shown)


dscho:
This is probably too large a patch to be sent to the mailing list, and it breaks the tests, too...

NishantJoshi00:
Yes it did, I was actually working on a script that replaces all the error messages and make them documentation friendly, I used regex, there might have been some faults along the way, does the tests provide me with where the errors might have occoured

NishantJoshi00:
> Yes it did, I was actually working on a script that replaces all the error messages and make them documentation friendly, I used regex, there might have been some faults along the way, does the tests provide me with where the errors might have occoured

I just found what caused the error, how can I run these tests locally on my own, git fork

dscho:
Here is a helpful document: https://github.com/git-for-windows/git/wiki/Running-Git's-regression-tests

gitster (commented):
Thanks for trying to help improve our documentation.

I think the author of this commit misunderstands what the "Do not *BEGIN* with capital" rule about die/error messages is about. It refers *ONLY* to the fact that unlike ordinary English sentence, where we begin a sentence with a capital letter, these messages do not capitalize the first letter only because it is at the beginning of the sentence.

So, if a word, even the one that happens to be at the beginning of die/warning message, has other reason to be spelled in capital letters, that is not something we want to downcase. As an example, changes to die("BUG") are unwelcome. Also, I see references to HEAD that are not even at the beginning of the message are butchered, but these changes are unwarranted.

But many other changes are exactly the kind of fixes we would appreciate. "cannot do X" should not begin with "C", for example.

It is entirely possible that the documentation guideline document *is* misleading and needs to be improved to avoid the misunderstanding that led to the undesirable changes in this commit, and perhaps that is worth fixing separately.

Thanks.

NishantJoshi00:
Ok, cool I got it, i will do those changes to my script, and test it locally, thanks so much for your insight

gitster:
Note that a "patch generated by script" is generally frowned upon, as the time spent running a script and sending the results out without proof-reading is far less expensive than the reviewer time on the list, which is the most scarce resource that is shared among community members. In addition, touching files that have otherwise some real activity on would put extra burden on the maintainer.

PR #1002, opened 2021-04-12
refs: print errno for read_raw_ref if GIT_TRACE_REFS is set

hanwen:
The ref backend API uses errno as a sideband error channel.

cc: Han-Wen Nienhuys <hanwen@google.com>


hanwen:
in master through merge bf0d4c84917025652e4ecf401ec55d130ad6f5a6

PR #1007, opened 2021-04-16
git-prompt: allow color in bash command subst

Artoria2e5:
Replace \[ and \] with readline's \01 and \02 to make color work in ps1 command
substitution. Bash provides these two as a shorthand in the prompt variables,
but does not replace them when they are coming from a command.

See: https://superuser.com/a/301355


Artoria2e5:
It looks like I need to be allowed to use GitGetGadget or something..?

Artoria2e5:
Right, signing off. Gotta amend.

gitster:
Naïve and stupid question. How would this interact with bash when readline is not being used? Do versions of recent bash so married to readline that we do not have to worry about it?

Artoria2e5:
@gitster Good question! Bash has, well, never really allowed being built *without* readline: you are supposed to use its bundled version if you don't have one.

Nevertheless, some BSD systems and macOS use `libedit` to pretend to be readline. A quick test on macOS's built-in bash 3.2.57 shows it is taking the advice of `\01` too (as in pressing "up" and trying to edit the previous line does not cause anything distorted, while not using it does).

gitster:
> @gitster Good question! Bash has, well, never really allowed being built _without_ readline: you are supposed to use its bundled version if you don't have one.
> 
> Nevertheless, some BSD systems and macOS use `libedit` to pretend to be readline. A quick test on macOS's built-in bash 3.2.57 shows it is taking the advice of `\01` too (as in pressing "up" and trying to edit the previous line does not cause anything distorted, while not using it does).

Thanks.

It is something worth mentioning in the proposed commit log message, I would think.

PR #1010, opened 2021-04-23
refs/debug: trace into reflog expiry too

hanwen:
this is split out from my larger reftable series. It improves debug output, so it's both not crucial to the reftable series, and should be uncontroversial to merge into next.


hanwen:
This is in master as of May 7, 2021.

merged in a850356d1b29ad915acb9f9e144867e90db16da3

PR #1014, opened 2021-04-29
mergetool: Add win32 path translation for DiffMerge

rhuijben:
Re-use the path translation added for winmerge to automatically find
SouceGear DiffMerge on Windows.


michaelcompressconsult:
be warned: the mergetool_find_win32_cmd() must not be called on linux as the current patch proposes since it searches for an *.exe.

however the description says it is also available on linux and osX
> DiffMerge is a non-free (but gratis) tool that supports OS X, Windows and Linux.

i propose to wrap it like for kdiff3 (pull 952, https://github.com/git/git/pull/952)

dscho:
@michaelcompressconsult please reply on the Git mailing list (see the "reply to this" link in the comment you want to address).

PR #1018, opened 2021-05-05
builtin/init-db: preemptively plug repo_fmt leak

ahunt:
This patch preemptively plugs a leak which is currently only occuring in seen in hn/reftable.

This patch is orthogonal to that series (that series starts allocating new memory into an already existing struct - and my patch only adds a clear call for that struct - in other words my patch is safe to use independent of hn/reftable). But there's also no good reason to use this patch independently of that series since we're not leaking prior to that series. I'm not sure what the optimal logistics with my patch are, feel free to integrate it into that series and/or to fold my change into the relevant commit if that's easiest!

CC: "Han-Wen Nienhuys" <hanwen@google.com>,


ahunt:
Closing because it's obsolete (I think).

PR #1019, opened 2021-05-06
Attempt to clarify a blurry sentence.

lboclboc:
Signed-off-by: Lars Berntzon <lars.berntzon@cecilia-data.se>

### DIFF
--- a/Documentation/git-push.txt
+++ b/Documentation/git-push.txt
@@ -66,10 +66,10 @@ it can be any arbitrary "SHA-1 expression", such as `master~4` or
 The <dst> tells which ref on the remote side is updated with this
 push. Arbitrary expressions cannot be used here, an actual ref must
 be named.
-If `git push [<repository>]` without any `<refspec>` argument is set to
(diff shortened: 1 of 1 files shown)


dscho:
Isn't it "clarify"? I.e. without the "e"?

lboclboc:
Yes, your text is clear.  I realize now I miss-read the line as if it was:
If `git push [<repository>]` without any `<refspec>` argument is set, to -update some ref at the destination with `<src>` with....
I mentally placed the comma there and then I could not get pass that. But the main problem I have I guess is that the line refers to <src> and <dst>, but those comes from the ref-spec, but the whole line is about when ref-spec is not specified so the placeholders <src> and <dst> are not valid.

ryujinxx (approved):

dscho:
Is this still active?

lboclboc:
I still think the current documentation is wrong and impossible to understand correctly but I don't know how to formulate it better so let's close it now and I'll try to reformulated it.

PR #1020, opened 2021-05-07
Directory traversal fixes

newren:
This patchset fixes a few directory traversal issues, where fill_directory() would traverse into directories that it shouldn't and not traverse into directories that it should (one of which was originally reported on this list at [1]).  And it includes a few cleanups

Changes since v4:
  - Tweak the trace2 statistics emitting a bit, as per suggestions from Jeff.
  - Introduce a new readdir_skip_dot_and_dotdot() helper at the end of the series, and use it everywhere we repeat the same code to skip '.' and '..' entries from readdir.  Also use it in dir.c's read_cached_dir() so we can be consistent about skipping it, even for statistics, across platforms.

If anyone has any ideas about a better place to put the "Some sidenotes" from the sixth commit message rather than keeping them in a random commit message, that might be helpful.

[1] See https://lore.kernel.org/git/DM6PR00MB06829EC5B85E0C5AC595004E894E9@DM6PR00MB0682.namprd00.prod.outlook.com/ or alternatively https://github.com/git-for-windows/git/issues/2732.

cc: Derrick Stolee <stolee@gmail.com>
cc: Jeff King <peff@peff.net>
cc: Jeff Hostetler <jeffhost@microsoft.com>
cc: Josh Steadmon <steadmon@google.com>
cc: Jeff Hostetler <git@jeffhostetler.com>

### DIFF
--- a/builtin/clean.c
+++ b/builtin/clean.c
@@ -189,10 +189,8 @@ static int remove_dirs(struct strbuf *path, const char *prefix, int force_flag,
 	strbuf_complete(path, '/');
 
 	len = path->len;
-	while ((e = readdir(dir)) != NULL) {
+	while ((e = readdir_skip_dot_and_dotdot(dir)) != NULL) {
 		struct stat st;
-		if (is_dot_or_dotdot(e->d_name))
-			continue;
 
 		strbuf_setlen(path, len);
 		strbuf_addstr(path, e->d_name);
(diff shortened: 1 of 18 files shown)


newren:
Weird, I had a green checkmark with all builds passing, then I entered "/submit", then it told me there was a build failure.  That appears to just be an infrastructure failure, but still kind of annoying that it behaved that way.

PR #1029, opened 2021-05-29
Update README.md

ananya26-vishnoi:
Grammatical Errors

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,21 +1,21 @@
-[![Build status](https://github.com/git/git/workflows/CI/PR/badge.svg)](https://github.com/git/git/actions?query=branch%3Amaster+event%3Apush)
+[![Build status](https://github.com/git/git/workflows/CI/PR/badge.svg)](https://github.com/git/git/actions? query=branch%3Amaster+event%3Apush)
 
 Git - fast, scalable, distributed revision control system
 =========================================================
 
 Git is a fast, scalable, distributed revision control system with an
-unusually rich command set that provides both high-level operations
-and full access to internals.
+unusually rich command set that provides both high-level operations and low-level operations
+and full access to the internals.
 
-Git is an Open Source project covered by the GNU General Public
+Git is an Open Source project covered by the GNU General Public.
 License version 2 (some parts of it are under different licenses,
-compatible with the GPLv2). It was originally written by Linus
-Torvalds with help of a group of hackers around the net.
+compatible with the GPLv2). It was originally written by Linus.
+Linux Torvalds with the help of a group of hackers around the net.
 
 Please read the file [INSTALL][] for installation instructions.
 
 Many Git online resources are accessible from <https://git-scm.com/>
-including full documentation and Git related tools.
+including full documentation and Git-related tools.
 
 See [Documentation/gittutorial.txt][] to get started, then see
 [Documentation/giteveryday.txt][] for a useful minimum set of commands, and
@@ -31,32 +31,29 @@ installed).
 
 The user discussion and development of Git take place on the Git
 mailing list -- everyone is welcome to post bug reports, feature
-requests, comments and patches to git@vger.kernel.org (read
+requests, comments, and patches to git@vger.kernel.org (read
 [Documentation/SubmittingPatches][] for instructions on patch submission).
 To subscribe to the list, send an email with just "subscribe git" in
 the body to majordomo@vger.kernel.org. The mailing list archives are
 available at <https://lore.kernel.org/git/>,
-<http://marc.info/?l=git> and other archival sites.
+<http://marc.info/?l=git>, and other archival sites.
 
-Issues which are security relevant should be disclosed privately to
+Issues that are security relevant should be disclosed privately to
 the Git Security mailing list <git-security@googlegroups.com>.
 
 The maintainer frequently sends the "What's cooking" reports that
 list the current status of various development topics to the mailing
-list.  The discussion following them give a good reference for
-project status, development direction and remaining tasks.
+list.  The discussion following them gives a good reference for
+project status, development direction, and remaining tasks.
 
 The name "git" was given by Linus Torvalds when he wrote the very
 first version. He described the tool as "the stupid content tracker"
 and the name as (depending on your mood):
 
- - random three-letter combination that is pronounceable, and not
-   actually used by any common UNIX command.  The fact that it is a
+ - random three-letter combination that is pronounceable, and not actually used by any common UNIX command.  The fact that it is a
    mispronunciation of "get" may or may not be relevant.
- - stupid. contemptible and despicable. simple. Take your pick from the
-   dictionary of slang.
- - "global information tracker": you're in a good mood, and it actually
-   works for you. Angels sing, and a light suddenly fills the room.
+ - stupid. contemptible and despicable. simple. Take your pick from the dictionary of slang.
+ - "global information tracker": you're in a good mood, and it actually works for you. Angels sing, and a light suddenly fills the room.
  - "goddamn idiotic truckload of sh*t": when it breaks
 
 [INSTALL]: INSTALL


README.md
```diff
@@ -1,21 +1,21 @@
-[![Build status](https://github.com/git/git/workflows/CI/PR/badge.svg)](https://github.com/git/git/actions?query=branch%3Amaster+event%3Apush)
+[![Build status](https://github.com/git/git/workflows/CI/PR/badge.svg)](https://github.com/git/git/actions? query=branch%3Amaster+event%3Apush)
```
dscho, line 1: That's incorrect.

README.md
```diff
@@ -1,21 +1,21 @@
...
+unusually rich command set that provides both high-level operations and low-level operations
+and full access to the internals.
 
-Git is an Open Source project covered by the GNU General Public
+Git is an Open Source project covered by the GNU General Public.
 License version 2 (some parts of it are under different licenses,
-compatible with the GPLv2). It was originally written by Linus
-Torvalds with help of a group of hackers around the net.
+compatible with the GPLv2). It was originally written by Linus.
+Linux Torvalds with the help of a group of hackers around the net.
```
dscho, line 13: Seriously?

dscho (commented):
This commit is so riddled with inaccuracies, I am starting to think that it was meant as a joke.

gitster:
> This commit is so riddled with inaccuracies, I am starting to think that it was meant as a joke.

That's harsh ;-)  I can only see the two parts that you commented on and have no idea on other changes the patch may had, but it looked like a honest attempt with carelessness (both made me suspect there is a sloppy bulk edit script involved, without enough proofreading).

README.md
dscho: To clarify: this extra space _breaks_ the link.

README.md
```diff
@@ -1,21 +1,21 @@
-[![Build status](https://github.com/git/git/workflows/CI/PR/badge.svg)](https://github.com/git/git/actions?query=branch%3Amaster+event%3Apush)
+[![Build status](https://github.com/git/git/workflows/CI/PR/badge.svg)](https://github.com/git/git/actions? query=branch%3Amaster+event%3Apush)
 
 Git - fast, scalable, distributed revision control system
 =========================================================
 
 Git is a fast, scalable, distributed revision control system with an
-unusually rich command set that provides both high-level operations
-and full access to internals.
+unusually rich command set that provides both high-level operations and low-level operations
```
dscho, line 7: "low-level operations" and "full access to internals" strikes me as one and the same thing, therefore this hunk adds something redundant.

README.md
```diff
@@ -1,21 +1,21 @@
...
+and full access to the internals.
```
dscho, line 8: I am not a native speaker, either, but from my understanding, "access to internals" is just as valid as "access to the internals", but it means something subtly different.

As I said, I am not a native speaker, either, therefore I would not try to change it.

README.md
```diff
@@ -1,21 +1,21 @@
...
 
-Git is an Open Source project covered by the GNU General Public
+Git is an Open Source project covered by the GNU General Public.
```
dscho, line 10: This full stop is incorrect, as it would end the sentence in the middle of the term "GNU General Public License".

README.md
dscho: His name is not Linus Linux Torvalds, therefore this hunk would change the correct name to an incorrect name.

I guess that the addition of "the" in front of the word "help" is actually correct, though.

README.md
```diff
@@ -1,21 +1,21 @@
...
-compatible with the GPLv2). It was originally written by Linus
-Torvalds with help of a group of hackers around the net.
+compatible with the GPLv2). It was originally written by Linus.
+Linux Torvalds with the help of a group of hackers around the net.
 
 Please read the file [INSTALL][] for installation instructions.
 
 Many Git online resources are accessible from <https://git-scm.com/>
-including full documentation and Git related tools.
+including full documentation and Git-related tools.
```
dscho, line 18: From what I understand, this is more a matter of taste, both are correct. Therefore I would go with "if it ain't broke, don't fix it" approach here.

README.md
```diff
@@ -31,32 +31,29 @@ installed).
 
 The user discussion and development of Git take place on the Git
 mailing list -- everyone is welcome to post bug reports, feature
-requests, comments and patches to git@vger.kernel.org (read
+requests, comments, and patches to git@vger.kernel.org (read
```
dscho, line 34: This Oxford comma is a matter of hot debate, and prone to lead to lengthy discussions of dubious value. It is correct "enough" without the comma, hence I would prefer for it to be left alone.

README.md
```diff
@@ -31,32 +31,29 @@ installed).
...
 [Documentation/SubmittingPatches][] for instructions on patch submission).
 To subscribe to the list, send an email with just "subscribe git" in
 the body to majordomo@vger.kernel.org. The mailing list archives are
 available at <https://lore.kernel.org/git/>,
-<http://marc.info/?l=git> and other archival sites.
+<http://marc.info/?l=git>, and other archival sites.
```
dscho, line 39: Same thing about the Oxford comma here.

README.md
```diff
@@ -31,32 +31,29 @@ installed).
...
 
-Issues which are security relevant should be disclosed privately to
+Issues that are security relevant should be disclosed privately to
```
dscho, line 41: Both are correct, therefore I consider it to be pointless to change from one to the other.

README.md
```diff
@@ -31,32 +31,29 @@ installed).
...
 the Git Security mailing list <git-security@googlegroups.com>.
 
 The maintainer frequently sends the "What's cooking" reports that
 list the current status of various development topics to the mailing
-list.  The discussion following them give a good reference for
-project status, development direction and remaining tasks.
+list.  The discussion following them gives a good reference for
```
dscho, line 46: This is a correct grammar fix.

README.md
```diff
@@ -31,32 +31,29 @@ installed).
...
+project status, development direction, and remaining tasks.
```
dscho, line 47: Another Oxford comma. I would be more in favor of this change if it added value, but from my point of view, it does not.

README.md
```diff
@@ -31,32 +31,29 @@ installed).
...
 
 The name "git" was given by Linus Torvalds when he wrote the very
 first version. He described the tool as "the stupid content tracker"
 and the name as (depending on your mood):
 
- - random three-letter combination that is pronounceable, and not
-   actually used by any common UNIX command.  The fact that it is a
+ - random three-letter combination that is pronounceable, and not actually used by any common UNIX command.  The fact that it is a
```
dscho, line 53: The only change here is that the bullet point is unwrapped into a too-long line.

README.md
```diff
@@ -31,32 +31,29 @@ installed).
...
    mispronunciation of "get" may or may not be relevant.
- - stupid. contemptible and despicable. simple. Take your pick from the
-   dictionary of slang.
- - "global information tracker": you're in a good mood, and it actually
-   works for you. Angels sing, and a light suddenly fills the room.
+ - stupid. contemptible and despicable. simple. Take your pick from the dictionary of slang.
```
dscho, line 55: Again, this only turns a correctly-wrapped bullet point into a single line that is too long.

Mind you, there is quite the room for improvement in this particular text. It is quite offensive and off-putting to casual readers. It would be a worthy goal to improve that, but that is not what this patch does.

README.md
```diff
@@ -31,32 +31,29 @@ installed).
...
+ - "global information tracker": you're in a good mood, and it actually works for you. Angels sing, and a light suddenly fills the room.
```
dscho, line 56: This, too, turns a correctly-wrapped bullet point into a line that spans more than 80 columns, which is the opposite of what the Git project wants.

dscho (commented):
Since @gitster asked so nicely for a thorough review, here it is. There are two correct grammar fixes in between these changes: adding "the" in front of the word "help", and using the singular "gives" for the noun "discussion".

The patch also introduces a couple of Oxford commata, which I would consider unnecessary churn.

And the rest are changes that are undesirable, such as breaking a URL, ending a sentence where it should not be ended, and turning correctly formatted text into overly long lines.

In essence, two grammar fixes could be extracted from this patch, but the rest is probably not something the Git project wants to take.

And then there is the commit message: I would suggest to take a look at the commit messages in the commit history of the `README.md` file: `git log README.md`. It can be easily seen that the expected form of commit messages is much more verbose, with a more informative first line, and a Signed-off-by: trailer.

PR #1033, opened 2021-06-10
Fix uninitialised reads found with MSAN

ahunt:
V2 replaces an #if'd memset with some brace initialisation (patch 3/3) as per review comments.

I've also removed an irrelevant "technically" from commit message 2/3, and fixed a typo in commit message 3/3.

cc: Chris Torek <chris.torek@gmail.com>
cc: Jeff King <peff@peff.net>


Neustradamus:
@ahunt: Please look upstream.

PR #1035, opened 2021-06-12
Fixes typo. Adds a missing space then the text reads 'is not' instead of 'isnot'

jordimas:
Fixes typo. Adds a missing space then the text reads 'is not' instead of 'isnot'


jordimas:
This has been fixed already by another commit

PR #1042, opened 2021-07-06
fix 2 more test files that were accessing file system directly. 

hanwen:
This reduced the number of test failures in the reftable series by a bit.


dscho:
> 
> 
> Error: Ignoring PR with empty title and/or body

Huh. That's funny. I thought we allowed that.

Ah. It is not a single patch. In that case, there _will_ be a cover letter, and an empty cover letter won't explain things well.

PR #1045, opened 2021-07-12
Allow to translate diff message

jordimas:


dscho (commented):
Looks good, except of course for the commit message ;-)

jordimas:
> Looks good, except of course for the commit message ;-)

Fixed! Thanks

dscho:
`/submit` time?

jordimas:
> `/submit` time?

My understanding is that the next step is to write a comment with /submit'

dscho:
> 
> 
> > `/submit` time?
> 
> My understanding is that the next step is to write a comment with /submit', correct? Thanks First time I do this

Yes. See https://github.com/git/git/pull/1035#issuecomment-860095512 for more details.

PR #1012, opened 2021-04-26
refs: cleanup errno sideband ref related functions

hanwen:
v5 
- address Ævar's comment; punt on clearing errno.

cc: Han-Wen Nienhuys <hanwen@google.com>
cc: Jonathan Tan <jonathantanmy@google.com>
cc: Ævar Arnfjörð Bjarmason <avarab@gmail.com>

PR #1068, opened 2021-08-17
fixup! propagate errno from failing read

hanwen:
This fixes a crash triggered by the BUG() statement.

This can occur with symlinked .git/refs. To check availability,
refs_verify_refname_available will run refs_read_raw_ref() on each prefix,
leading to a read() from .git/refs (which is a directory).

When handling the symlink case, it is probably more robust to re-issue the
lstat() as a normal stat(), in which case, we would fall back to the directory
case.

For now, propagating errno from strbuf_read() is simpler and avoids the crash.

cc: Han-Wen Nienhuys hanwen@google.com
cc: Jonathan Tan jonathantanmy@google.com
cc: Ævar Arnfjörð Bjarmason avarab@gmail.com
cc: Jonathan Nieder <jrnieder@gmail.com>
cc: Jeff King <peff@peff.net>
cc: Carlo Marcelo Arenas Belón         <carenas@gmail.com>


hanwen:
this was folded into #1012

PR #1062, opened 2021-08-08
Give support for hooks based on platform

rafael-santiago:
The idea behind this commit can be useful for teams
that share git-hooks into a custom directory and
dealing with projects that must be developed,
built, maintained on several different platforms.

This commit allows the execution of git hooks
based on the current operating system.
A "native hook" is defined in the form:
    hooks/hook-name_platform
Where platform must be equivalent to the
content returned in sysname field in utsname
struct when calling uname() [but all normalized
in lowercase].

On Windows, independent of version, flavor, SP,
whatever it is simply "windows".

When a native hook is not found the standard
hook (.git/hook/hook-name), if found is executed
of course. In other words, the hook without a
platform postfix (_yyz) is the standard hook.
When native hook is not set as executable but
standard is set, the standard will be executed.

The main motivation of this extension is to
reduce dependency of scripting languages,
logical trinkets etc just to execute minor
tasks during scm events that could be done
natively but differently from a platform
to another. Less dependencies, cleaner
repos: a small step for a better world
for any software developer.

Thanks for taking the time to contribute to Git! Please be advised that the
Git community does not use github.com for their contributions. Instead, we use
a mailing list (git@vger.kernel.org) for code submissions, code reviews, and
bug reports. Nevertheless, you can use GitGitGadget (https://gitgitgadget.github.io/)
to conveniently send your Pull Requests commits to our mailing list.

Please read the "guidelines for contributing" linked above!

PR #1069, opened 2021-08-19
Give support for hooks based on platform

rafael-santiago:
The idea behind this commit can be useful for teams that share git-hooks into a custom directory and dealing with projects that must be developed, built, maintained on several different platforms.

This commit allows the execution of git hooks based on the current operating system. A "native hook" is defined in the form: hooks/hook-name_platform

Where platform must be equivalent to the content returned in sysname field in utsname struct when calling uname() [but all normalized in lowercase].

On Windows, independent of version, flavor, SP, whatever it is simply "windows".

When a native hook is not found the standard hook (.git/hook/hook-name), if found is executed, of course. In other words, the hook without a platform postfix (_yyz) is the standard hook. When native hook is not set as executable but standard is set, the
standard will be executed.

The main motivation of this extension is to reduce dependency of scripting languages, logical trinkets etc just to execute minor tasks during scm events that could be done natively but differently from a platform to another. Less dependencies, cleaner
repos: a small step for a better world for any software developer.

### DIFF
--- a/run-command.c
+++ b/run-command.c
@@ -1319,9 +1319,50 @@ int async_with_fork(void)
 #endif
 }
 
+static inline const char *platform_name(void)
+{
+	static const char *platform = NULL;
+#ifndef GIT_WINDOWS_NATIVE
+	static struct utsname un = { 0 };
+#endif
+	if (platform != NULL)
+		return platform;
+
+#ifndef GIT_WINDOWS_NATIVE
+	if (uname(&un) == 0) {
+		for (size_t s = 0; un.sysname[s] != 0; s++)
+			un.sysname[s] = tolower(un.sysname[s]);
+		platform = un.sysname;
+	}
+#else
+    platform = "windows";
+#endif
+
+    return platform;
+}
+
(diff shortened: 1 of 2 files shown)


dscho:
Since this is a continuation of https://github.com/git/git/pull/1062, I'd like to point out that force-pushing the PR branch will update the PR, there is not actually a need for opening a new PR 😃

rafael-santiago:
> 
> 
> Since this is a continuation of #1062, I'd like to point out that force-pushing the PR branch will update the PR, there is not actually a need for opening a new PR 😃

The problem was that all git incantation to put mine commits at the top of fixed merged stuff from upstream was becoming tricky and I was not secure about not screw up other people work. So I chosed the "lumberjack's solution" of squashing mine changes from my fork, reserve a patch from it, remove my fork and forking again from original git's repo. When I deleted my prior fork the related PR here was closed I was unable to open it again, so another "lumberjack's solution": open a new one... 😃

sancarn:
This would be pretty big for me, look forward to cross-platform hooks being easier in the near future 😊

PR #1072, opened 2021-08-20
depth=1

mablue:
Significant reduction in downloads from GitHub and download time when a noob just wants to download the latest version of a repository
by setting default depth to 1.
its really usefull cuz if we just think to that how much people in the world downloading withouth --depth command and just use this
```
 git clone {repo url}
```
tbnh, maybe also It can even help cool down the earth!
```
depth =50 to depth=1
```

### DIFF
--- a/builtin/pack-objects.c
+++ b/builtin/pack-objects.c
@@ -44,9 +44,8 @@
  */
 static struct packing_data to_pack;
 
-static inline struct object_entry *oe_delta(
-		const struct packing_data *pack,
-		const struct object_entry *e)
+static inline struct object_entry *oe_delta(const struct packing_data *pack,
+					    const struct object_entry *e)
 {
 	if (!e->delta_idx)
 		return NULL;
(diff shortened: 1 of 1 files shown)


gitster:
Not a good idea to make it the default, even though having an option to clone with a small depth may be useful at times.

What's the primary use case envisioned for repositories cloned in this way? Build, install and remove?
If one truly only cares about the download bandwidth, triggering "git archive" on the other end may be a better alternative.

mablue:
> Not a good idea to make it the default, even though having an option to clone with a small depth may be useful at times.
> 
> What's the primary use case envisioned for repositories cloned in this way? Build, install and remove?
> If one truly only cares about the download bandwidth, triggering "git archive" on the other end may be a better alternative.

I just use git clone repo address and never use --depth 1 when I just want to review the code or use the end version of it or etc for example change and commit. but you know Where I decided to commit this ? whe I want to just download and compile bitcoin and linux repo and not using the depth parameter. increased the time of download very much and I think to that what if all india and china work without --depth param for this case of usage?!I think git Wastes bunch of mony energy and computional sources bouth server and client side and yes I care about that also I care about who just wasted his time every day for downloading un needed thigs! please open this pr for thinking and talking more. thnks

mdrahulkhan2021:
thenks

PR #1074, opened 2021-08-21
Make ident dynamic, not just a hardcoded value of "$Id".

sobomax:
This allows ident to be something like $FreeBSD$ so it provides matching functionality for repos migrated from CVS / SVN.

This works by allowing ident to have a parameter, i.e.:

`* ident=MyCustomId`

In .gitattributes.

cc: Matheus Tavares <matheus.bernardino@usp.br>
cc: Đoàn Trần Công Danh         <congdanhqx@gmail.com>

### DIFF
--- a/Documentation/gitattributes.txt
+++ b/Documentation/gitattributes.txt
@@ -382,6 +382,14 @@ sign `$` upon checkout.  Any byte sequence that begins with
 `$Id:` and ends with `$` in the worktree file is replaced
 with `$Id$` upon check-in.
 
+The `ident` attribute can also provide an optional value,
+which if supplied is going to be used for expansion instead of
+the string `Id`.
+
+------------------------
+*.[ch]		ident=FreeBSD
+------------------------
+
 
 `filter`
 ^^^^^^^^
--- a/builtin/checkout--worker.c
+++ b/builtin/checkout--worker.c
@@ -9,7 +9,7 @@ static void packet_to_pc_item(const char *buffer, int len,
 			      struct parallel_checkout_item *pc_item)
 {
 	const struct pc_item_fixed_portion *fixed_portion;
-	const char *variant;
+	const char *variant, *ident_action;
 	char *encoding;
 
 	if (len < sizeof(struct pc_item_fixed_portion))
@@ -19,7 +19,8 @@ static void packet_to_pc_item(const char *buffer, int len,
 	fixed_portion = (struct pc_item_fixed_portion *)buffer;
 
 	if (len - sizeof(struct pc_item_fixed_portion) !=
-		fixed_portion->name_len + fixed_portion->working_tree_encoding_len)
+		fixed_portion->name_len + fixed_portion->working_tree_encoding_len +
+		fixed_portion->ident_action_len)
 		BUG("checkout worker received corrupted item");
 
 	variant = buffer + sizeof(struct pc_item_fixed_portion);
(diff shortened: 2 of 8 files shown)


gitster:
The name of the "ident_action" structure may need rethinking. It appears it _is_ used to specify how the custom "ident" is spelled, so the shortest and workable alternative might be to call it just "ident".

Having said that, I have a feeling that this is not such a great feature, not because the way it is made customizable by this commit is bad, but simply because the original "$Id$" is not a very good match to the data model of Git. We probably are better off leaving the (mis-)feature as it is now, without spending any more brain-cycles on it.

sobomax:
@gitster thanks, Junio for a review/feedback!

> The name of the "ident_action" structure may need rethinking. It appears it _is_ used to specify how the custom "ident" is spelled, so the shortest and workable alternative might be to call it just "ident".

Other similar data structures are called xyz_action, so it's basically named for consistency and to allow future extensions. Also there is ident_filter.ident which is different and is used in same contexts. However, if it's the only culprit, I'd change it happily.

> Having said that, I have a feeling that this is not such a great feature, not because the way it is made customizable by this commit is bad, but simply because the original "$Id$" is not a very good match to the data model of Git. We probably are better off leaving the (mis-)feature as it is now, without spending any more brain-cycles on it.

Coming from purely git user experience and until recently being quite oblivious about its inner working I don't feel qualified to comment on the fitness of the feature for Git in general. However, I can tell you that there are probably thousands of users that have been affected by the loss of $FreeBSD$ tag expansion after the project has switched from SVN to Git few months ago (myself included). You see, over the years the community has build useful tools to help managing configuration files and now those tools are broken. And while Git ident expansion does not match CVS/SVN, but it seems to be good enough for many of those use cases. The issue is important enough to warrant official Wiki page explaining the situation: https://wiki.freebsd.org/VCSFeatureDollarFreeBSD.

mdrahulkhan2021:
thenks boro

sobomax:
> /allow

Thanks, Đoàn!

sobomax:
> > From: Maksym Sobolyev <sobomax@sippysoft.com>
> >
> > This allows ident to be something like $FreeBSD$ so it provides matching
> > functionality for repos migrated from CVS / SVN.
> >
> > This works by allowing ident to have a parameter, i.e.:
> >
> > * ident=MyCustomId
> >
> > In .gitattributes.
> 
> I would have expected that, just like cvs uses its configuration
> mechanism for LocalKeyword, this would belong to the .git/config
> file.
> 
> The attribute mechanism is inherently for stuff that you would want
> to specify per-path, so the "do we do keyword expansion for this
> file (yes/no)?" is a good match to be an attribute, and that
> question is what the original 'ident' attribute asks.
> 
> I do not think "what keyword do we want to expand as if it were
> $Id$?" is a question we want to be giving different answers per
> path, and it does not smell like a good match to be an attribute.
> 
> By allowing a potentially separate values for
> 
>     *.txt ident=FreeBSD
>     *.cc  ident=OpenBSD
> 
> this gives long rope to create a mess to the users.

Except it actually might be quite useful. Let's say you are some project that integrates stuff from different open-source projects (something like FreeRTOS), so it  may actually be desirable to set different keywords for a different parts of the tree. Say you get some .[ch] files from FreeBSD and add your own custom build glue/business logic on top of that, or mix it with OpenBSD sources.

If anything my 30 years journey into software development has taught me, is that more flexibility in tools is always better, especially if it does not cost much extra to have in this case.

> I also think that the ident expansion is overrated, and have a
> feeling that we should discourage its use, not encourage it, but
> that is a separate story.

I guess that's where I we respectfully disagree. I think original use of the Id tag was to have some cheap way to reconcile source from the built binary and the GIT ident works just fine for that purpose. Both file name and commit hash can be recovered from the blob ID. It takes a little of effort to go from the blob ID, but the amount of work needed fits into 10 lines of the shell script and possibly could be compacted even further.

-Max

CC: Junio C Hamano <gitster@pobox.com>

> [On the Git mailing list](https://lore.kernel.org/git/xmqqa6l8qbgc.fsf@gitster.g), Junio C Hamano wrote ([reply to this](https://github.com/gitgitgadget/gitgitgadget/wiki/ReplyToThis)):
> 
> ```

dscho:
> 
> 
> /allow ikke

This has little to do with _this_ particular PR, it's just that @ikke is active on IRC, and when asked to `/allow` here [said they couldn't](https://colabti.org/irclogger/irclogger_log/git-devel?date=2021-08-21#l21), and now they could.

ShettyGaneshprasad (approved):

dscho:
@sobomax let's close this?

PR #1080, opened 2021-08-29
Add a new --remerge-diff capability to show & log

newren:
Here are some patches to add a --remerge-diff capability to show & log, which works by remerging merge commits (without touching the index or working tree), and then comparing the recorded merge commit against the automatic remerge result.  Since merges are often clean and unmodified, this would result in an empty diff for most merges.  However, the --remerge-diff option can show the edits users made to resolve the conflict -- removing conflict markers, which hunks they picked from the various conflicts to keep, and changes made outside of conflict markers (which might reflect changes needed to resolve semantic conflicts, or additional changes an integrator felt were useful).

This new option is perhaps most like the existing --cc option, so some comparisons:

Advantages relative to --cc:
  * Seems to be much more understandable to most users than --cc output.
  * Faster than --cc (!!)
Disadvantages relative to --cc:
  * Does not (currently) work for octopus merges[1]
  * Will not work on a read-only or full filesystem[2]

[1] I have nebulous ideas of writing an Octopus-centric ORT strategy (basically just repeatedly invoking ort, but trying to be smart about conflicts), but my ideas are still only half baked.  For octopus merges, a simple warning is printed instead.
[2] New blobs/trees can be written by the three-way merging step.  These are written to a temporary area under the git object store that is cleaned up at the end of the operation.


demaford:
newren:remerge-diff

newren:
I could not change the base branch, so I'm closing this in favor of https://github.com/gitgitgadget/git/pull/1103

PR #1082, opened 2021-08-30
userdiff: support enum keyword in PHP hunk header

zonuexe:
"enum" keyword will be introduced in PHP 8.1.
https://wiki.php.net/rfc/enumerations

### DIFF
--- a/t/t4018/php-enum
+++ b/t/t4018/php-enum
@@ -1,0 +1,4 @@
+enum RIGHT: string
+{
+    case Foo = 'ChangeMe';
+}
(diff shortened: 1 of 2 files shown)


t/t4018/php-enum
```diff
@@ -0,0 +1,3 @@
+enum RIGHT: string {
+    case Foo = 'ChangeMe';
```
dscho, line 2: This seems to break the test suite: https://github.com/git/git/pull/1082/checks?check_run_id=3465586494#step:5:3769

zonuexe: Fixed at 1fe4aa4e692c099e5d7098b1e4076482b7e18eb4 (squashed)

mohsin1980 (commented):
https://ghcr.io/v2/homebrew/core/git-remote-hg/blobs/sha256:3903ddefc5ed6142943aa33ba298ac51d054159f0c401bcde044934494202a19

PR #1085, opened 2021-09-07
Removal of untracked files/directories and the current working directory

newren:
Here's a series of patches that started with a bug report about the current working directory being deleted by rebase[1], and is based on a slightly modified idea from Peff to make the code handle such cases more carefully[2].  While working on it, I discovered that my premise that codepaths already carefully protect untracked (but not ignored) files was incorrect, and another rebase bug report came in at the time about that issue as well[3].  I'm using Junio's declaration about `checkout -f` and `reset --hard` (and also presuming that since `read-tree --reset` is porcelain that its behavior should be left alone)[4] in this series.  The fun didn't end there either; I also discovered some other bugs in am & stash as a side effect and which affected my ability to write testcases for those commands.  And I noticed a few places that the documentation should be improved.

So, a long series.  I could potentially break it up into five (partially dependent) series if others prefer:
  * Patches 1-3: fix up stash bugs I found while doing this work
  * Patches 4-5: document some am bugs and fix one of them
  * Patch 6: testcase cleanups
  * Patch 7-10: new testcases for nuking of untracked files, fixes for some them, and better docs
  * Patches 11-18: new testcases for nuking cwd and fixes for all these testcases
The first three series would be independent, and the last two would depend on all the previous ones.  Let me know if you'd rather see them submitted this way.

SIDENOTE (since it was brought up in the threads I reference):

There's another related topic here of treating ignored files as precious instead of deleting them.  I do not try to handle that here, but I believe that would actually be relatively easy to handle.  If you leave unpack_trees_options->dir as NULL, then they are treated as precious (my original patch 8 made that mistake).  There's a few other locations that already optionally set up unpack_trees_options->dir (a quick search for "overwrite_ignore" and "overwrite-ignore" will find them), so we'd just need to copy that optionality to these new callsites (and perhaps make a global core.overwrite_ignored config option to affect all of these).  Of course, doing so would globally treat ignored files as precious rather than allowing them to be configured on a per-path bases, but honestly I think the idea of configuring ignored files as precious on a per-path basis sounds like insanity (we have enough bugs with untracked and ignored files without adding yet another type, tla/baz was excessively confusing to me due in part to the number of types of files and I'd rather not see such ideas ported to git, configuring per-path rules sounds like lots of work for end users to configure, etc.), so if someone wants to pursue the precious-ignored concept then I'd much rather see it done as a global setting.  Just my $0.02.

[1] https://lore.kernel.org/git/xmqqv93n7q1v.fsf@gitster.g/
[2] https://lore.kernel.org/git/YS8eEtwQvF7TaLCb@coredump.intra.peff.net/
[3] https://lore.kernel.org/git/CABPp-BFyR19ch71W10oJDFuRX1OHzQ3si971pMn6dPtHKxJDXQ@mail.gmail.com/
[4] https://lore.kernel.org/git/xmqqr1e2ejs9.fsf@gitster.g/

cc: Johannes Sixt j6t@kdbg.org
cc: Jeff King peff@peff.net
cc: Taylor Blau me@ttaylorr.com
cc: Fedor Biryukov <fedor.birjukov@gmail.com>
cc: Yuri <yuri@rawbw.com>


newren:
I decided to split this review up into separate series:
   * https://github.com/git/git/pull/1086
   * https://github.com/git/git/pull/1087
   * https://github.com/git/git/pull/1088
   * https://github.com/gitgitgadget/git/pull/1036
   * https://github.com/gitgitgadget/git/pull/1037

PR #1094, opened 2021-09-22
win32: drop -Wno-incompatible-pointer-types when pedantic

carenas:
27e0c3c6cf (win32: allow building with pedantic mode enabled, 2021-09-03)
adds -W as a workaround to the use of `void *` to hold a pointer to a
function, punting the final solution.

The type used in Windows is a generic function type, but assigning it
from a different function interface will require a cast that was
missing from the original implementation, so lets add that.

Sadly; that is not enough, as gcc will helpfully raise a -Wcast-function-type
warnig when casting between functions that might have incompatible
return types (ex: GetUserNameExW returns bool which is only half the
size of the return type from FAPROC), so create a new type that could be
used as a completely generic function pointer, which then could be assigned
to the correct function pointer through a cast.

This has the advantage that the generic pointer functions then will be
opaque and not useful as functions without being assigned to the right
function pointer variable (just like `void *` was meant to be in POSIX).

Signed-off-by: Carlo Marcelo Arenas Belón <carenas@gmail.com>

### DIFF
--- a/compat/win32/lazyload.h
+++ b/compat/win32/lazyload.h
@@ -15,10 +15,12 @@
  *                        source, target);
  */
 
+typedef void (*FARVOIDPROC)(void);
+
 struct proc_addr {
 	const char *const dll;
 	const char *const function;
-	FARPROC pfunction;
+	FARVOIDPROC pfunction;
 	unsigned initialized : 1;
 };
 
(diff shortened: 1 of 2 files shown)


carenas:
corresponding to patches in list from thread:

  https://lore.kernel.org/git/CAPUEspjaKUxq4_kptMC305Rkc7mEUEY_JGTxrRTiTaw5km87oQ@mail.gmail.com/T/#u

PR #1097, opened 2021-09-30
rebase -r: Allow '/' in generated label names

rwe:
The `git rebase -r` generated-label sanitization introduced in commit
cd55222 (rebase -r: let `label` generate safer labels, 2019-11-17)
was intended to avoid dangerous NTFS/FAT characters, especially
colon/double-quote/pipe chars. It exempted non-ASCII UTF-8 characters,
but of the ACII set only permitted alphanumerics.

However, it's fairly common for organizations to use nested branch names,
e.g. "bug/1234", and `git rebase -i -r` began to render those as `label
bug-1234`. This is despite the original name clearly being a valid ref
name, as evidenced by already being one.

That sanitization logic is taught here to permit non-terminal slashes
which follow an alphanumeric character.


rwe:
Yes, but that was already a hazard of having those two branch names to begin with.

dscho:
> Yes, but that was already a hazard of having those two branch names to begin with under `refs/heads`.

Not at all. You can easily run into such a situation by accepting PRs.

rwe:
> Not at all. You can easily run into such a situation by accepting PRs.

This is a good point. My claim was not that those labels cannot show up, but that they're an existing hazard of the way branches and other namespaced refs are currently handled. You can run into the same problem by just fetching from a remote, too, if they create `foo/bar`, you fetch, they delete `foo/bar` and create `foo`, and you fetch again.

Avoiding that was not (I believe) among the goals behind the commit (cd5522271f6b985114b33332e148bff2283b0440) which broke the very convenient label ↔︎ branch name heuristic, which is certainly the more common case. And with rebase, it's easy to avoid those rare conflicts by changing the label in `git-rebase-todo`.

But it's possible that this would be a bigger issue than I think, so I've prepared another branch based on this which deconflicts path prefixes, avoiding the issue: https://github.com/rwe/git/tree/permit-and-deconflict-slash-in-rebase-label

dscho:
I don't think that the benefits are worth the incredible pain of introducing d/f conflicts that will then have to be deconflicted with a rather fragile logic (slash after dot on Windows, anyone?).

In other words, I think that you can only "worsimprove" the design from here. The very idea has this fundamental flaw that it is too complex. For example, why stop at slashes? Why not also allow other special characters? Except of course on Windows, where they cannot be used because some are not valid in filenames.

So you see, you want to replace a simple, easy-to-explain design by something that is the opposite.

If you insist, you can submit this to the mailing list, I will then repeat the same objection there, where I fully expect other regulars to agree with my reasoning.

rwe:
> By your own reasoning, dots should be allowed, though.

I only said that the behaviour of dots was not changed here: they're still always replaced with `-`, and so any concerns about interacting with `/` aren't relevant. I didn't claim one way or another about whether they _should_ be filtered.

Slashes are more important: slash separators, when used in an organization, become ubiquitous or even mandatory in branch names, and are frequently used in CI policies or branch permission patterns on shared repos. Again, not defending that practice, just dealing with it. As far as I know, dots are not usually used in that way, and so I don't feel is worthwhile to spend attention on it by suggesting restoring that behaviour.

Though, labels still go through ref name validation, so this "sanitization" is just to avoid later rejections, as far as I can tell. And we're only talking about the _generated_ labels here; you can still use a dot and a slash now, if you just do so in the editor. This isn't about label validation, it's just generation. Also, even the rebase manpage uses a dot in its example labels!

> Why would you want to create a new merge and lose the entire original merge commit message?

By "a new merge", there's no original merge commit message. What I mean is introducing an intermediate merge between branches, which is useful for keeping finer-grained conflict resolutions:

```diff
 reset onto
 pick aaaaaaa # update formatting rules
 label task/new-fmt
 
 reset onto
 pick bbbbbbb # fix terrible bug
 merge -C abababa # Merge branch 'task/new-fmt' into 'bug/1234'
 label bug/1234
 
 reset onto
 pick ccccccc # factor out bug generator
+merge task/new-fmt
 label task/5678
 
 reset onto
 merge -C abdabda bug/1234 # Merge branch 'bug/1234' into 'dev'
 merge -C cdcdcdc task/5678 # Merge branch 'task/5678' into 'dev'
```

Prior to git 2.25, when doing the above rebase, the labels in the generated `git-rebase-todo` would match the inferred branch names, so that the default merge message of that new merge would include "Merge branch 'task/new-fmt'". After 2.25, the labels are prepared such that the new label would be `task-new-fmt`, leading to a default message of `Merge branch 'task-new-fmt'` for the unwary new merge. It's annoyance: so in those orgs, I do something like a crappy and error-prone `:%s/^(label|reset|(merge[^#]+) )(task|bug|…)-/\1\2\//` to have `git-rebase-todo` set up with original labels.

> They are not meant to be an accurate reflection of the original branch name, by any stretch of imagination.

I know they cannot uniquely or unambiguously correspond in general. But in local development, the heuristic correspondence was a big convenience. Not only for having the default merge message, but also that it was often possible to use the labels to parameterize editor macros like `label foo` → "append `x git update-ref refs/{heads,rewritten}/foo` to end" in order to maintain selected local branch heads.

It was previously "good enough" to fairly smoothly use that kind of workflow, but then became _much_ worse for repos using slashes, and requires very little to restore that. The change here is intended to restore those use-cases in the simplest way possible.

dscho:
> Slashes are more important

That's just _your_ opinion, man.

rwe:
Not really an opinion. Certainly subject to more data though. You think that dots are used in branch-name based CI or permissions patterns as often as slashes? I don't think I've bumped into it personally, but could imagine for versioned branches. And why would that be relevant here at all, unless you're asking to expand be the scope of this change? Is that what you're saying?

PR #1107, opened 2021-10-15
Adding GNU License reference

rafaelstz:
I linked the file to the GNU License reference mentioned in the README file, it's the same approach with the INSTALL and other links mentioned.

Screen Shot 2021-10-14 at 8 33 38 PM


rimrul:
The change looks reasonable, but the commit message neeeds work. Please read [Documentation/SubmittingPatches](https://github.com/git/git/blob/master/Documentation/SubmittingPatches).

PR #1114, opened 2021-10-19
RFC: Server side merges (no ref updating, no commit creating, no touching worktree or index)

newren:
(NOTE for Junio: Depends on en/remerge-diff to avoid a minor conflict.  That series has no outstanding feedback for about three weeks so far, so it should be stable (and mergeable to next).  Besides, this series is still RFC, so if you want to wait to pick it up...)

Stuff intentionally NOT included in this v3, but which others seemed to feel strongly about; they'd need to convince me more on these:
  * Any form of diff output[1]
  * A way to omit printing the generated tree hash[2][3]
See also the new "Mistakes to avoid" section I added to Documentation/git-merge-tree.txt.
 
[1] https://lore.kernel.org/git/nycvar.QRO.7.76.6.2201101427440.339@tvgsbejvaqbjf.bet/ (section starting with "Providing a tree")
[2] https://lore.kernel.org/git/CABPp-BHvXrP0sTTmuTYfACoJTCcm9+wk_f441nj4TstrmQdqMQ@mail.gmail.com/ (sections starting with "avoid printing" and "Where did I suggest")
[3] https://lore.kernel.org/git/CABPp-BGdbh=HM7jA+_RTwSWVcMr_qvEY7RoNXooeBT2NB4Ubzw@mail.gmail.com/ (section starting with "Providing a tree object")

Updates since v2 (thanks to Christian, Dscho, Ramsay, and René for suggestions and comments on v2):
  * Rebased on top of en/remerge-diff
  * Changed output format -- includes more information by default for conflicts
  * Instead of just conflicted filenames, provide (mode, oid, stage, file) tuples (i.e. `ls-files -u` style of information), though there's a flag to pick just the filenames.
  * Much more thorough manual for git-merge-tree.txt
  * Renamed option from `--real` to `--write-tree`
  * Accept an optional `--trivial-merge` option to get old style merge-tree behavior
  * Allow both `--write-tree` and `--trivial-merge` to be omitted since we can deduce which from number of arguments
  * Document exit code when the merge cannot be run (so we can distinguish error from both clean and from conflicted)
  * testcase cleanups: test_tick, early skip of test when using recursive backend, variable renames, etc.
  * various minor code cleanups
  * Add a new --allow-unrelated-histories option (with same meaning as the one used in `git merge`)

Updates since v1 (thanks to Johannes Altmanninger and Fabian for suggestions on v1):
  * Fixed a bad patch splitting, and a style issue pointed out by Johannes Altimanninger
  * Fixed misleading commit messages in new test cases
  * Fixed my comments about how commit-tree could be used to correctly use two `-p` flags

NOTE2: A preliminary version of this series was discussed here: https://lore.kernel.org/git/nycvar.QRO.7.76.6.2110211147490.56@tvgsbejvaqbjf.bet/

NOTE3: An alternative has been implemented by Christian, over here: https://lore.kernel.org/git/20220105163324.73369-1-chriscool@tuxfamily.org/

This series introduces a new option to git-merge-tree: --write-tree.  This new option is designed to allow a server-side "real" merge (or allow folks client-side to do merges with branches they don't even have checked out), with the resulting top-level tree being returned.  Real merges differ from trivial merges in that they handle:
  * three way content merges
  * recursive ancestor consolidation
  * renames
  * proper directory/file conflict handling
  * etc.

The reason this is different from `merge` is that `merge-tree` does NOT:
  * Read/write/update any working tree (and assumes there probably isn't one)
  * Read/write/update any index (and assumes there probably isn't one)
  * Create a commit object
  * Update any refs

cc: Christian Couder <chriscool@tuxfamily.org>
cc: Ramsay Jones <ramsay@ramsayjones.plus.com>
cc: René Scharfe <l.s.r@web.de>

### DIFF
--- a/Documentation/diff-options.txt
+++ b/Documentation/diff-options.txt
@@ -34,7 +34,7 @@ endif::git-diff[]
 endif::git-format-patch[]
 
 ifdef::git-log[]
---diff-merges=(off|none|on|first-parent|1|separate|m|combined|c|dense-combined|cc)::
+--diff-merges=(off|none|on|first-parent|1|separate|m|combined|c|dense-combined|cc|remerge|r)::
 --no-diff-merges::
 	Specify diff format to be used for merge commits. Default is
 	{diff-merges-default} unless `--first-parent` is in use, in which case
@@ -63,6 +63,14 @@ ifdef::git-log[]
 	This makes merge commits show the full diff with respect to
 	each of the parents. Separate log entry and diff is generated
 	for each parent.
++
+--diff-merges=remerge:::
+--diff-merges=r:::
+--remerge-diff:::
+	With this option, two-parent merge commits are remerged to
+	create a temporary tree object -- potentially containing files
+	with conflict markers and such.  A diff is then shown between
+	that temporary tree and the actual merge commit.
 +
 --diff-merges=combined:::
 --diff-merges=c:::
--- a/Documentation/git-merge-tree.txt
+++ b/Documentation/git-merge-tree.txt
@@ -58,6 +58,7 @@ simply one line:
 Whereas for a conflicted merge, the output is by default of the form:
 
 	<OID of toplevel tree>
+	<Conflicted file list>
 	<Informational messages>
 
 These are discussed individually below.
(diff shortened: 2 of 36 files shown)


newren:
Closing in favor of gitgitgadget#1122, since I can't change the base for the pull request (at least not to a branch within the gitgitgadget repo).

PR #1125, opened 2021-11-05
Consolidate find_header logic into one function

john-cai:
This addresses the NEEDSWORK comment in builtin/receive-pack.c:

```
 /**
   * NEEDSWORK: reuse find_commit_header() from jk/commit-author-parsing
   * after dropping "_commit" from its name and possibly moving it out
   * of commit.c
   **/
```

There are two functions that have very similar logic of finding a header
value. find_commit_header, and find_header. We can conslidate the logic
by introducing a new function find_header_mem, which is equivalent to
find_commit_header except it takes a len parameter that determines how
many bytes will be read. find_commit_header and find_header can then both
call find_header_mem.

This reduces duplicate logic, as the logic for finding header values
can now all live in one place.

Changes since v4:
- adjust verbiage of NEEDSWORK comment block

Changes since v3:
- added NEEDSWORK block detailing what needs to be done to clean up find_header_mem
- fixed verbiage in commit message
- adjusted style of an if block (based on Junio's feedback)

### DIFF
--- a/builtin/receive-pack.c
+++ b/builtin/receive-pack.c
@@ -581,32 +581,19 @@ static char *prepare_push_cert_nonce(const char *path, timestamp_t stamp)
 	return strbuf_detach(&buf, NULL);
 }
 
-/*
- * NEEDSWORK: reuse find_commit_header() from jk/commit-author-parsing
- * after dropping "_commit" from its name and possibly moving it out
- * of commit.c
- */
 static char *find_header(const char *msg, size_t len, const char *key,
 			 const char **next_line)
 {
-	int key_len = strlen(key);
-	const char *line = msg;
+	size_t out_len;
+	const char *val = find_header_mem(msg, len, key, &out_len);
 
-	while (line && line < msg + len) {
-		const char *eol = strchrnul(line, '\n');
+	if (!val)
+		return NULL;
 
-		if ((msg + len <= eol) || line == eol)
-			return NULL;
-		if (line + key_len < eol &&
-		    !memcmp(line, key, key_len) && line[key_len] == ' ') {
-			int offset = key_len + 1;
-			if (next_line)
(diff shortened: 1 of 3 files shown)


dscho:
This will need a much better commit message in order to avoid being perceived as a refactoring for refactoring's sake.

jackiebrewster307:
Rootaccess

john-cai:
@dscho could I ask for a review on this one? I tried fixing a NEEDSWORK but I'm not sure if this is the right idea. Could you let me know?

dscho:
> @dscho could I ask for a review on this one?

Sorry, I am really pressed for time this weekend. Maybe you can send it to the list for review? Or if you want a pre-review, ask on the git-mentoring list?

builtin/receive-pack.c
```diff
@@ -581,32 +581,26 @@ static char *prepare_push_cert_nonce(const char *path, timestamp_t stamp)
...
-			if (next_line)
-				*next_line = *eol ? eol + 1 : eol;
-			return xmemdupz(line + offset, (eol - line) - offset);
-		}
-		line = *eol ? eol + 1 : NULL;
+	size_t out_len;
+	const char *eol;
+	char *ret;
+
+	const char *val = find_header(msg, key, &out_len);
```
rimrul, line 590: Does the first commit compile without the second? Isn't `find_header()` still called `find_commit_header()` at this point in time?

john-cai: I think you're right. I'll re-roll these commits so both compile

john-cai: @rimrul would you be able to take another look and let me know if this is the right idea? 🙏

rashid212rm:
Null

rashid212rm:
Gplv3 salutation please

PR #1128, opened 2021-11-10
Default git-p4 branch changes from master -> main

bccbrendan:
Also adding some more forgiving unicode decoding
on p4 changeset descriptions.

### DIFF
--- a/git-p4.py
+++ b/git-p4.py
@@ -181,7 +181,7 @@ def prompt(prompt_text):
 if bytes is not str:
     # For python3, always encode and decode as appropriate
     def decode_text_stream(s):
-        return s.decode() if isinstance(s, bytes) else s
+        return s.decode('utf-8', 'backslashreplace') if isinstance(s, bytes) else s
     def encode_text_stream(s):
         return s.encode() if isinstance(s, str) else s
 else:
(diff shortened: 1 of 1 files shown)


git-p4.py
```diff
@@ -4035,7 +4041,7 @@ def run(self, args):
                 print("p4-git branches: %s" % self.p4BranchesInGit)
                 print("initial parents: %s" % self.initialParents)
             for b in self.p4BranchesInGit:
-                if b != "master":
+                if b != "main":
```
dscho, line 4044: Should this be `not b in ["main", "master"]`?

git-p4.py
```diff
@@ -769,13 +769,19 @@ def p4CmdList(cmd, stdin=None, stdin_mode='w+b', cb=None, skip_info=False,
...
+                    key = key.decode('utf-8', 'backslashreplace')
                     if isinstance(value, bytes) and not (key in ('data', 'path', 'clientFile') or key.startswith('depotFile')):
-                        value = value.decode()
+                        try:
+                            value = value.decode('utf-8', 'backslashreplace')
+                        except UnicodeDecodeError as e:
+                            if key == 'desc':
+                                value = str(value)[2:-1]  # work around unicode decode errors in description (hex escapes will just be printed as \x7a for instance.
+                            else:
+                                raise e
```
dscho, line 780: This hunk, as well as the ones immediately preceding and following it, probably want to live in their own commit. Maybe even in their own PR.

dscho (commented):
I like where this is going!

To ensure that this works as advertised, _especially_ with existing setups, I think we will have to add a test case that emulates such an existing setup (by renaming `p4/main` to `p4/master`) and then performing at least one typical operation that now still needs to find `p4/master` (because there is no `p4/main`).

I would also like to ask you to have a closer look at [the commit history of `git-p4.py`](https://github.com/git/git/commits/main/git-p4.py) and to imitate the commit messages' style. Look e.g. at 6b79818bfbd39a5d41d15003375a9f382f70ad0e, it has an outstanding commit message.

PR #1139, opened 2021-11-14
subtree: ignore merge.ff setting

koutcher:
When `merge.ff` is set to `only` in .gitconfig, `git subtree pull` will
fail with error `fatal: Not possible to fast-forward, aborting.`. This
fix ignores the `merge.ff` setting when using `git merge` within subtree.

Signed-off-by: Thomas Koutcher <thomas.koutcher@online.fr>
cc: Thomas Koutcher <thomas.koutcher@online.fr>

### DIFF
--- a/contrib/subtree/git-subtree.sh
+++ b/contrib/subtree/git-subtree.sh
@@ -976,10 +976,10 @@ cmd_merge () {
 
 	if test -n "$arg_addmerge_message"
 	then
-		git merge -Xsubtree="$arg_prefix" \
+		git merge --no-ff -Xsubtree="$arg_prefix" \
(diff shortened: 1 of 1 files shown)


koutcher:
Thanks for your reviews Ævar and Junio. I'll update the patch taking your comments into account.

dscho:
> 
> 
> Thanks for your reviews Ævar and Junio. I'll update the patch taking your comments into account.

Please note that the mentioned reviewers won't see this reply unless you send it to the Git mailing list as per https://github.com/gitgitgadget/gitgitgadget/wiki/ReplyToThis.

PR #1142, opened 2021-11-18
mingw: fix 64 bit build because of missing *time_r() function definitions

carenas:
This is a cherry-pick from git-for-windows/git#3398, that is needed to be able to build recent master with the Git for Windows SDK or recent MinGW64.

It was discussed recently on list[1], and might need further adjustements if the 32-bit Git for Windows SDK also updates their winpthread headers, requiring a similar fix, but since without it a plain build from master wouldn't work in Windows I think it is worth reviewing on its own merits, and had withstood the test of time in the Git for Windows fork since it was originally merged there late August.

Changes since v1:
- Fixed grammar in the commit message (suggested by Junio)
- Added ACK (proposed by dscho)

[1] https://lore.kernel.org/git/20211005063936.588874-1-mh@glandium.org/

### DIFF
--- a/git-compat-util.h
+++ b/git-compat-util.h
@@ -127,7 +127,9 @@
 /* Approximation of the length of the decimal representation of this type. */
 #define decimal_length(x)	((int)(sizeof(x) * 2.56 + 0.5) + 1)
 
-#if defined(__sun__)
+#ifdef __MINGW64__
+#define _POSIX_C_SOURCE 1
+#elif defined(__sun__)
  /*
   * On Solaris, when _XOPEN_EXTENDED is set, its header file
   * forces the programs to be XPG4v2, defeating any _XOPEN_SOURCE
--- a/compat/mingw.c
+++ b/compat/mingw.c
@@ -1083,6 +1083,7 @@ int pipe(int filedes[2])
 	return 0;
 }
 
+#ifndef __MINGW64__
 struct tm *gmtime_r(const time_t *timep, struct tm *result)
 {
 	if (gmtime_s(result, timep) == 0)
@@ -1096,6 +1097,7 @@ struct tm *localtime_r(const time_t *timep, struct tm *result)
 		return result;
 	return NULL;
 }
+#endif
 
 char *mingw_getcwd(char *pointer, int len)
 {


dscho:
Is this really your first time with GitGitGadget, @carenas?

git-compat-util.h
```diff
@@ -127,7 +127,9 @@
 /* Approximation of the length of the decimal representation of this type. */
 #define decimal_length(x)	((int)(sizeof(x) * 2.56 + 0.5) + 1)
 
-#if defined(__sun__)
+#ifdef __MINGW64__
+#define _POSIX_C_SOURCE 1
```
dscho, line 131: I wonder how much of a fall-out this has on things other than `*time_r()`...

dscho:
> WARNING: carenas has no public email address set on GitHub

GitGitGadget currently has only one way to find out what email address you have (which it needs in order to Cc: you on the mails it sends): it looks at the GitHub profile for a public email address.

I have a few ideas how to remedy that, but no time to go forward with any of them at the moment.

git-compat-util.h
carenas: I chose 1 to minimize the fall-out, but I can only see it as an advantage for compatibility with a codebase that targets POSIX anyway.

The only possible undesired effect I was able to track was its interaction with __USE_MINGW_ANSI_STDIO, but that not a problem unless the settings for it in the Makefile/cmake change as well AFAIK.

There is also the fact, this has been in use in the windows fork for a while without any known sideffects.

But I could change it to use the less aggresive _POSIX_THREAD_SAFE_FUNCTIONS which is mentioned in the list and which would make this even safer

carenas: Alternatively we could revert this change and go with the one Mike propossed on list instead, which will also make the MSVC and gcc builds more similar (both using the fallback), and which might be a good change to do early in this release cycle as well.

dscho: > There is also the fact, this has been in use in the windows fork for a while without any known sideffects.

Oh, so this has withstood the test of time in Git for Windows? _goes and checks_ ... indeed, it is in git-for-windows/git@adf0cd8d14262e56db39573287a0c16555eb892b.

That information would have been quite good in the cover letter or even in the commit message.

dscho:
> mingw-w64's pthread_unistd.h had a bug that mistakenly (because there is no support for the *lockfile() functions required[1]) defined _POSIX_THREAD_SAFE_FUNCTIONS and that was being worked around since [3ecd153](https://github.com/git/git/commit/3ecd153a3bdbefb25a4bf451d7b91cf8f1ece2a9) (compat/mingw: support MSys2-based MinGW build, 2016-01-14).
> 
> the bug was fixed in winphtreads, but as a sideeffect, leaves the reentrant functions from time.h not longer visible and therefore breaks the build.
> 
> since the intention all along was to avoid using the fallback functions, formalize the use of POSIX by setting the corresponding feature flag and to make the intention clearer compile out the fallback functions.
> 
> [1] https://unix.org/whitepapers/reentrant.html
> 
> Signed-off-by: Carlo Marcelo Arenas Belón [carenas@gmail.com](mailto:carenas@gmail.com)
> 
> Thanks for taking the time to contribute to Git! Please be advised that the Git community does not use github.com for their contributions. Instead, we use a mailing list ([git@vger.kernel.org](mailto:git@vger.kernel.org)) for code submissions, code reviews, and bug reports. Nevertheless, you can use GitGitGadget (https://gitgitgadget.github.io/) to conveniently send your Pull Requests commits to our mailing list.
> 
> Please read the "guidelines for contributing" linked above!

Please edit this comment before sending the patch, the first comment of a GitGitGadget PR will be used as a cover letter. Therefore it would be best not to repeat the commit message, and certainly to omit the template that is intended to be replaced by the user opening the PR.

PR #1146, opened 2021-11-24
Add a workflow for creating GitHub release notes

mahozad:
Because this is not directly the git code and is related to GitHub CI, I post it here.

This pull request adds a new GitHub Actions workflow that automatically creates releases on GitHub repository when pushing a new tag to the repository.

It can be useful for GitHub users who want to subscribe to "Release" notifications so they are notified whenever a new release is created.


dscho:

Please fix this, GitGitGadget gave good advice about it:

> Also, it is a good idea to review the commit messages one last time, as the Git project expects them in a quite specific form:
> 
> * the lines should not exceed 76 columns,
> 
> * the first line should be like a header and typically start with a prefix like "tests:" or "revisions:" to state which subsystem the change is about, and
> 
> 

Also, please fix https://github.com/git/git/runs/4315074301?check_suite_focus=true#step:3:33:

> .github/scripts/generate-github-changelog.main.kts:22: new blank line at EOF.

mahozad:
@dscho Fixed the issues.

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,40 @@
...
+
+jobs:
+  create-gh-release:
+    name: Create a new release or update an existing release in the GitHub repository
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2  # OR '0' To retrieve all preceding commit.
```
rimrul, line 23: Why 2? And why ddo we need that comment?

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,40 @@
+name: Create GH release
+
+# Create a GitHub release for each new tag.
+# The release notes are taken from the release notes file
+# modified in that commit located in Documentation/RelNotes directory.
+
+on:
+  push:
+    tags:
+      - v[0-9]+.*
```
rimrul, line 10: Schoudn't we filter out release candidates?

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Create the release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref_name }}
+          release_name: ${{ github.ref_name }}
+          body_path: changelog.txt
+          draft: false
+          prerelease: false
```
rimrul, line 40: If we want to include release candidates, shouldn't they be marked as pre-releases?

.github/scripts/generate-github-changelog.main.kts
```diff
@@ -0,0 +1,21 @@
...
+    println("\t${index + 1}- $name")
+}
+
+val notesFile = args
+    .map(::File)
+    .singleOrNull { "RelNotes" in it.parent }
+
+notesFile
+    ?.copyTo(File("changelog.txt"))
+    ?: println("No release notes file modified in this commit")
```
rimrul, line 21: Why do we care if the file was changed? We know its location either way.

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,40 @@
...
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2  # OR '0' To retrieve all preceding commit.
+      - name: Get changed files
+        uses: tj-actions/changed-files@v11.7
+        id: changed-files
+        with:
+          separator: ' '
+      - name: Generate the changelog
+        run: kotlin .github/scripts/generate-github-changelog.main.kts ${{ steps.changed-files.outputs.all_changed_and_modified_files }}
```
rimrul, line 30: Do we need to spin up a jvm and run 20 lines of kotlin to copy a single file? What does kotlin get us that `cp` doesn't?

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Create the release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref_name }}
+          release_name: ${{ github.ref_name }}
+          body_path: changelog.txt
```
rimrul, line 38: Do we even need to copy the file? Can't we just use `Documentation/RelNotes/${GITHUB_REF#refs/tags/v}.txt` directly here and drop the previous two steps completely?

rimrul (commented):
This seems a little convoluted for what it's trying to achieve.

.github/workflows/create-release.yml
mahozad: The comment is not needed. The action only fetches a single revision by default (`fetch-depth: 1`). To get the list of modified files, *tj-actions/changed-files@v11.7* needed last two revisions.

.github/workflows/create-release.yml
mahozad: This is up to you or other repository owners.
But I think there is no problem publishing release candidates too.

.github/workflows/create-release.yml
dscho: That will give you the list of files modified in the tip commit. Not between tags.

.github/workflows/create-release.yml
mahozad: I don't know how to do that :)

.github/workflows/create-release.yml
dscho: Yes.

.github/workflows/create-release.yml
dscho: Even more importantly, why introduce yet another language into a code base that does not suffer from a shortage of languages used: C, Perl, Shell, Python, Makefile, Javascript, ...

dscho:
@rimrul Maybe you can also comment on-list, as @gitster strongly prefers that over reviews on GitHub PRs...

.github/scripts/generate-github-changelog.main.kts
mahozad: I copied the  created/modified file to *changelog.txt* and then used that in `body_path: changelog.txt`.
I'm not good at command line.
Maybe there is a better way so the file name can be directly inserted in `body_path: ` parameter.

.github/workflows/create-release.yml
mahozad: You're right.

.github/workflows/create-release.yml
dscho: It did not take a lot of research on my part to find https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-ignoring-branches-and-tags...

dscho: > It did not take a lot of research on my part to find https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-ignoring-branches-and-tags...

Sorry, that was meant as a comment on https://github.com/git/git/pull/1146#discussion_r756847622

About `prerelease: false` conditional on full releases: you add a `run:` step before this one that sets an environment variable to `true` when running on an `-rc` version and otherwise `false`, then use `${{ env.<name> }}` here.

rimrul:
I know, I know. But sending a reply to the list requires me to be at home, but I can write GitHub reviews from my phone.

.github/workflows/create-release.yml
dscho: >  But I think there is no problem publishing release candidates too.

I do see a rather big problem publishing them as full releases: https://github.com/git/git/releases/latest would then point to a pre-release, not the latest full release. That's not desirable.

.github/workflows/create-release.yml
mahozad: Again, you're right. The whole workflow can probably be simplified.
Could anyone please do that?

I just wanted to be notified of new releases.

dscho: >  Could anyone please do that?

Yes: you can do that.

> I just wanted to be notified of new releases.

Sure. And you can still make this happen.

.github/workflows/create-release.yml
mahozad: Yes, exactly.

mahozad:
Also see https://github.com/github/roadmap/issues/247.

dscho:
https://semgrep.dev/blog/2025/popular-github-action-tj-actionschanged-files-is-compromised/ seems relevant to this PR...

PR #1150, opened 2021-12-01
terminal: refactor read_key_without_echo() to simplify code and fix it at least for macOS

carenas:
The escape key detection process was not working in macOS because there poll() will always timeout as stdin is implemented via a device and therefore can't be polled.

Eitherway, the alternative while simplifying the code, add also support for UTF-8 keys and AFAIK seems to be as functional as the original code.

First patch is definitely code complete and likely useful all the way to maint, patches 2 and 3 implement features that might interact with other features and bugfixes in fly, hence why this is only posted as a draft.

### DIFF
--- a/add-interactive.c
+++ b/add-interactive.c
@@ -10,6 +10,7 @@
 #include "dir.h"
 #include "run-command.h"
 #include "prompt.h"
+#include "compat/terminal.h"
 
 static void init_color(struct repository *r, struct add_i_state *s,
 		       const char *section_and_slot, char *dst,
@@ -30,6 +31,7 @@ static void init_color(struct repository *r, struct add_i_state *s,
 void init_add_i_state(struct add_i_state *s, struct repository *r)
 {
 	const char *value;
+	int single_key = 0;
 
 	s->r = r;
 
@@ -69,7 +71,14 @@ void init_add_i_state(struct add_i_state *s, struct repository *r)
 	git_config_get_string("diff.algorithm",
 			      &s->interactive_diff_algorithm);
 
-	git_config_get_bool("interactive.singlekey", &s->use_single_key);
+	git_config_get_bool("interactive.singleKey", &single_key);
+
+	if (single_key && !terminal_support(CBREAK)) {
+		warning(_("missing single keystrock support, "
+			  "disabling interactive.singleKey"));
+		single_key = 0;
+	}
+	s->use_single_key = single_key;
 }
 
 void clear_add_i_state(struct add_i_state *s)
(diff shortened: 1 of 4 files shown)


add-interactive.c
```diff
@@ -69,7 +71,14 @@ void init_add_i_state(struct add_i_state *s, struct repository *r)
 	git_config_get_string("diff.algorithm",
 			      &s->interactive_diff_algorithm);
 
-	git_config_get_bool("interactive.singlekey", &s->use_single_key);
+	git_config_get_bool("interactive.singleKey", &single_key);
+
+	if (single_key && !terminal_support(CBREAK)) {
+		warning(_("missing single keystrock support, "
+			  "disabling interactive.singleKey"));
```
phillipwood, line 78: Translating the warning message is an improvement but I'd be tempted to use the original wording that is deleted below. It explains the problem in non-technical terms and avoids the issue that `interactive.singleKey` should not be translated within the message.

add-patch.c
```diff
@@ -1184,8 +1184,10 @@ static int read_single_character(struct add_p_state *s)
 {
 	if (s->s.use_single_key) {
 		int res = read_key_without_echo(&s->answer);
-		printf("%s\n", res == EOF ? "" : s->answer.buf);
-		return res;
+		if (res != EOF) {
+			printf("%s\n", s->answer.buf);
```
phillipwood, line 1188: This is a change of behavior when res == EOF - is that intended? If so maybe call it out in the commit message

compat/terminal.c
```diff
@@ -286,28 +278,36 @@ void restore_term(void)
...
+	return EOF;
+}
 
-	if (!warning_displayed) {
-		warning("reading single keystrokes not supported on this "
-			"platform; reading line instead");
-		warning_displayed = 1;
+#endif
+
+int terminal_support(enum terminal_mode mode)
```
phillipwood, line 299: I had trouble understanding the purpose of this function from its name. I think something like `test_terminal_support` would be clearer.

phillipwood (commented):
I think this is a good change, I'm not sure about the motivation being "perl compatibility" though - we want this change because it makes sense for our use case, not because we want to emulate what perl does per se.

phillipwood (commented):
I think the commit message and subject could be more explicit that this is reverting the commits mentioned in the message. I'm not against removing the escape support as I don't think we're using it but would it be possible to avoid `poll` and just set `VTIME` and `VMIN` in read_key_without_echo() instead?

dscho:
@carenas can't this be closed in favor of https://github.com/gitgitgadget/git/pull/1146 (which made it into v2.36.0 already)?

PR #1177, opened 2021-12-31
t4126: this test does not pass SANITIZE_LEAK; quit claiming it does

newren:
My builds in several different topics keep running into this, and I'm sure it's a false positive -- I didn't change anything affecting this test.  I've just been ignoring it and submitting anyway, and I suspect others are too.


newren:
Looks like Junio decided not to pick this one up.

PR #1204, opened 2022-01-21
blame: Add --ignore-revs-blob and blame.ignoreRevsBlob

davebarrau:
In a bare repository, there isn't a simple way to
ignore revisions via a file without extracting it
to a temporary file.

This patch provides a command-line option and
config variable, similar to --ignore-revs-file
and blame.ignoreRevsFile, which reads the list
of revisions to ignore from a blob in the
repository.

Signed-off-by: David Barr <git@davebarr.dev>

cc: Ævar Arnfjörð Bjarmason avarab@gmail.com


davebarrau:
@newren Can you "allow" me please? 🙇

davebarrau:
> /allow

Thanks!

PR #1206, opened 2022-01-26
repack: add --filter=<filter-spec>

john-cai:
Currently, repack does not work with partial clones. When repack is run
on a partially cloned repository, it grabs all missing objects from
promisor remotes. This also means that when gc is run for repository
maintenance on a partially cloned repository, it will end up getting
missing objects, which is not what we want.

In order to make repack work with partial clone, teach repack a new
option --filter, which takes a <filter-spec> argument. repack will skip
any objects that are matched by <filter-spec> similar to how the clone
command will skip fetching certain objects.

The final goal of this feature, is to be able to store objects on a
server other than the regular git server itself.

There are several scripts added so we can test the process of using a
remote helper to upload blobs to an http server:

- t/lib-httpd/list.sh lists blobs uploaded to the http server.
- t/lib-httpd/upload.sh uploads blobs to the http server.
- t/t0410/git-remote-testhttpgit a remote helper that uploads blobs onto
  an http server. Copied over from t/t5801/git-remote-testhttpgit and
  modified to upload blobs to an http server.
- t/t0410/lib-http-promisor.sh convenience functions for uploading
  blobs


gab:
If I may, while the ability to offload objects onto an http remote is indeed very interesting - and might open up the possibility of offloading large blobs to cheaper storage like git-lfs allows -, the main and very crucial use case I see for this feature is the possibility to **clean up a partial clone from blobs accumulated over time**!

Currently partial clones seem viable to deal (finally) with very large binary files in git, but the .git folder will grow over time to the point that you need to re-clone eventually, something that's disruptive to users, and tricky to do without losing any tag, stash, note or branch. Not very sustainable.

Ideally we'd have some garbage collection system that automatically gets rid of older blobs in a partial clone, but for now I'd happily settle for your feature, as it would provide a command team members can run occasionally that slims down their local repo's history.

A common use case will likely be to keep locally any blob belonging to the currently checked-out revision, so perhaps you could add an option for this and make it the default?

PR #1228, opened 2022-03-16
[RFC] rebase: reword: new feature change the subject in the todo list

tmaillart:
If the user only wants to modify the subject of a commit during
interactive rebase, now he can choose the "reword" command and
change the subject directly in the todo list.
If "reword" is selected and the subject has not changed, then we
use the default behavior.

This is probably not the correct way to implement this new feature,
or perhaps it is not desirable, so I would like some feedback.
Thank you


builtin/rebase.c
```diff
@@ -152,6 +152,7 @@ static struct replay_opts get_replay_opts(const struct rebase_options *opts)
 		oidcpy(&replay.squash_onto, opts->squash_onto);
 		replay.have_squash_onto = 1;
 	}
+	replay.subject_len = replay.content_len = 0;
```
dscho, line 155: This is unnecessary because of the initialization in line 120.

sequencer.c
```diff
@@ -990,6 +991,7 @@ static int run_git_commit(const char *defmsg,
 			  struct replay_opts *opts,
 			  unsigned int flags)
 {
+	char message[opts->content_len + opts->subject_len];
```
dscho, line 994: I do not think that C allows variable-sized arrays on the stack. The CI build [agrees](https://github.com/git/git/runs/5570793178?check_suite_focus=true#step:5:235). You will need `xmalloc()` and `free()` for that.

sequencer.c
```diff
@@ -1033,7 +1035,13 @@ static int run_git_commit(const char *defmsg,
 		strvec_push(&cmd.args, "--cleanup=strip");
 	if ((flags & VERBATIM_MSG))
 		strvec_push(&cmd.args, "--cleanup=verbatim");
-	if ((flags & EDIT_MSG))
+	if ((flags & EDIT_MSG) && (flags & INLINE_MSG)) {
+		memcpy(message, opts->subject, opts->subject_len);
+		memcpy(message + opts->subject_len, opts->content,
+				opts->content_len);
+		strvec_push(&cmd.args, "-m");
+		strvec_push(&cmd.args, message);
```
dscho, line 1043: This can easily run into command-line limits. That would not be an issue if we only ever pass the `cmd.args` to an internal function, but it _is_ a problem if we pass this to an external process. Have you looked into that?

sequencer.c
```diff
@@ -2145,6 +2153,22 @@ static int do_pick_commit(struct repository *r,
 		if (res || command != TODO_REWORD)
 			goto leave;
 		reword = 1;
+		if (strlen(msg.subject) != item->arg_len ||
+		    strncmp(msg.subject, item->arg_offset,
+			    item->arg_len) != 0) {
+			const char *commit_buf, *subject;
```
dscho, line 2159: Git's coding convention does not allow declarations after statements. The declarations of variables must come first.

sequencer.h
```diff
@@ -108,7 +113,8 @@ struct todo_item {
 	unsigned int flags;
 	int arg_len;
 	/* The offset of the command and its argument in the strbuf */
-	size_t offset_in_buf, arg_offset;
+	size_t offset_in_buf;
+	const char *arg_offset;
```
dscho, line 117: That's not possible because `arg_offset` points into a `strbuf` that can be reallocated at any time, which would render any pointer stale.

In any case, a change like this needs to be refactored into its own commit.

dscho (commented):
Thank you for the idea and for an initial patch!

Before sending this off, you will not only want to look into the issues pointed out by @gitgitgadget, but also introduce a flag that enables inline subject editing, and introduce tests and documentation for this feature.

sequencer.c
```diff
@@ -1033,6 +1034,24 @@ static int run_git_commit(const char *defmsg,
...
 	if ((flags & VERBATIM_MSG))
 		strvec_push(&cmd.args, "--cleanup=verbatim");
+	if ((flags & EDIT_MSG) && (flags & INLINE_MSG)) {
+		int ret, size = opts->content_len + opts->subject_len + 1;
+		char *message;
+
+		if (size <= 1)
+			return error(_("no valid message length"));
+
+		message = malloc(size);
```
dscho, line 1044: Please use `xmalloc()`; This will handle out-of-memory scenarios by exiting instead of crashing in the next line.

builtin/rebase.c
tmaillart: I have deleted this statement.

sequencer.c
tmaillart: As suggested in the other comment I use *stdin* to transfer the commit message now.
So allocation is not needed anymore.

sequencer.c
```diff
@@ -1033,6 +1034,24 @@ static int run_git_commit(const char *defmsg,
...
+		int ret;
+
+		strvec_push(&cmd.args, "-F");
+		strvec_push(&cmd.args, "-");
+		cmd.in = -1;
+		ret = start_command(&cmd);
+		if (ret)
+			return ret;
+		if (write(cmd.in, opts->subject, opts->subject_len) !=
+		    opts->subject_len);
```
dscho, line 1047: This semicolon is most certainly in need of being removed.

sequencer.h
dscho: I would strongly suggest to stick with offsets.

sequencer.c
tmaillart: I'm not sure I understand your point.
First line of this block there are declarations only.
What am I missing ?

sequencer.h
tmaillart: I cannot get a string reference from a `todo_item` if there are only offsets.
Maybe you mean I should keep the `arg_offset` as is, and introduce another field ?

tmaillart: I changed the type of `arg_offset` because in the function
```
static int do_pick_commit(struct repository *r,
			  struct todo_item *item,
			  struct replay_opts *opts,
			  int final_fixup, int *check_todo)
```
There is no reference to the current `todo_list` (required to get a reference to the `strbuf` or to call `todo_item_get_arg`).

Am I missing an obvious way to reach a pointer to the current todo_list's `strbuf` ?

dscho: Just add the `todo_list` as parameter ;-)

sequencer.c
```diff
@@ -2331,11 +2331,10 @@ static int do_pick_commit(struct repository *r,
...
 fast_forward_edit:
-			if (item->arg_len > 0 && (strlen(msg.subject) !=
-			    item->arg_len ||
-			    strncmp(msg.subject, todo_item_get_arg(todo_list,
-				    item),
-				    item->arg_len) != 0)) {
+			if (opts->allow_inline_reword && item->arg_len > 0 &&
+			    (strlen(msg.subject) != item->arg_len ||
+			     strncmp(msg.subject, todo_item_get_arg(todo_list,
+				     item),item->arg_len) != 0)) {
```
dscho, line 2337: It would probably make most sense to move the new `allow_inline_reword` attribute as well as this hunk to the first commit, so that the patch series does not read like "let's introduce that feature! Oh never mind, let's make it opt-in!"

t/t3404-rebase-interactive.sh
```diff
@@ -955,6 +955,17 @@ test_expect_success 'rebase -ix with --autosquash' '
 	test_cmp expected actual
 '
 
+test_expect_success 'rebase -i with --allow-inline-reword' '
```
dscho, line 958: Since this verifies that `--allow-inline-reword`, this test should probably be squashed into the second commit that adds support for that option.

dscho (commented):
Good work! I think this does not need much work before it can be `/submit`ed.

sequencer.h
```diff
@@ -56,6 +56,11 @@ struct replay_opts {
 	enum commit_msg_cleanup_mode default_msg_cleanup;
 	int explicit_cleanup;
 
+	const char *content;
+	int content_len;
+	const char *subject;
+	int subject_len;
+
```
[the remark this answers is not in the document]
tmaillart: You are right, I'm sorry about that, I did not mean any offense (just a mistranslation from french).
I will fix this.

sequencer.c
```diff
@@ -2311,9 +2331,30 @@ static int do_pick_commit(struct repository *r,
 		*check_todo = !!(flags & EDIT_MSG);
 		if (!res && reword) {
 fast_forward_edit:
+			if (item->arg_len > 0 && (strlen(msg.subject) !=
+			    item->arg_len ||
+			    strncmp(msg.subject, todo_item_get_arg(todo_list,
+				    item),
```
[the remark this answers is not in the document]
tmaillart: I agree I have only tested on simple cases.
Regarding the shortening, I think it should be ok as there is this test `strlen(msg.subject) != item->arg_len`.
I have recorded a little demo https://asciinema.org/a/YyEFhNzEshxu7FxczsS4yjrPq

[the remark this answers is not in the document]
dscho: True, that length check addresses part of my concern.

But what if you have a [beautiful commit message](https://www.youtube.com/watch?v=4qLtKx9S9a8) (remember, such messages consist of substantially more than just a one-liner), and change the first line that is presented in the todo list? That will drop the rest of that commit message, making it a non-beautiful commit message.

dscho:
@tmaillart while I believe that this feature encourages too-short commit messages, you should feel free to contribute this to the Git mailing list and see whether you can find more support for this feature there. I am not the Git maintainer, after all.

sequencer.c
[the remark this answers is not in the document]
tmaillart: I didn't realize that the link to the demo expired, sorry.
So I've made a new one using **linux** source tree to illustrate this new feature with *beautiful commit messages*: https://asciinema.org/a/T9tEmUjjl4dyDuaXalPzXNzo1
Originally, the goal is just to be able to quickly fix a typo in multiple commit subjects at a time.

Soheilafsha:
درود و احترام به توسعه دهندگان محترم .
من از کد نویسی و پیدا کردن مکان مناسب برای انتقال ایده های ذهنم اگاهی زیادی ندارم، امیدوارم از همین تریبون بتونم افکارم رو به شما انتقال دهم و شما عزیزان کمک کنید هر کلمه و جمله پ پیشنهاد من در پروژه های مربوطه که وجود دارند قرار گیرند . یا در صورت نیاز پروژه جدید پیدایش گردد ، از نظرات و پیشنهادات کلیه علاقمندان استفاده شود ، با هم ترکیب شود و به پیشرفت کمک کند.
هدف و انگیزه من فقط و فقط ارتقا و نواوری در علم و تکنولوژی با بهره بری از قدرت تمام هوش بشریت و ادغام با هوش مصنوعی و رسیدن به بهترین گزینه و تصمیم گیری .
از کسانی که دوست دارند در هر بخش گروهبندی و سازماندهی کنند ، مدیریت برنامه ریزی های مختلف رو به افراد شایسته اموزش واستخدام کنند ، دعوت به همکاری میکنم .
پروژه من جهانی و کلی هست ، با منطقه بندی و بخش بخش های مختلف در تمام حوضه های اجتماعی ، کارافرینی، اشتغال زایی اموزشی و غیره میباشد که در طول پروژه ، مرحله به مرحله بخش های مربوطه را فرا میخوانم و به همکاری دعوت مینمایم.
با تشکر از تک تک افراد که دنیا رو به جای بهتری تبدیل میکنند.

dscho:
@Soheilafsha please note that you need to communicate in English, not in Persian, in this project, and that there is little you can do in this here PR.

PR #1229, opened 2022-03-18
git-p4: fix issue with perforce branch synchronized multiple times.

kfrolov:
Perforce branch might be sync'ed multiple times with different
revision numbers (by the command "git p4 sync --branch=NAME //perforce/path...")

In such situation, git p4 will see multiple perforce branches with
same name and different revision numbers. The problem is that to make
a shelve, git-p4 script will try to find "origin" branch, if not
specified in command line explicitly. And previously script selected
any branch with same name and don't mention particular revision number.
Later this may cause failure of the command "git diff-tree -r $rev^ $rev",
so shelve can't be created (due to wrong origin branch/commit).

This commit fixes the heuristic by which git p4 selects origin branch:
first it tries to select branch with same perforce path and perforce
revision, and if it fails, then selects branch with only same perforce
path (ignoring perforce revision number).


dscho:
> Thanks for taking the time to contribute to Git! Please be advised that the Git community does not use github.com for their contributions. Instead, we use a mailing list ([git@vger.kernel.org](mailto:git@vger.kernel.org)) for code submissions, code reviews, and bug reports. Nevertheless, you can use GitGitGadget (https://gitgitgadget.github.io/) to conveniently send your Pull Requests commits to our mailing list.
> 
> Please read the "guidelines for contributing" linked above!

@kfrolov could you please remove this before `/submit`ing?

PR #1233, opened 2022-04-07
cat-file: wip: use packfile order in -batch-command --unordered

john-cai:
When git cat-file --batch-command --unordered --buffer is used, we can
optimize the fetch of objects by going through the packfile in packfile
order.

This works by keeping a hashmap of oids that are seen from stdin that map
to an entry with a corresponding function of the command. When flush
is issued, we iterate through the packfile and check each oid against the map,
and if it's found then we execute the function for that oid.

For all the oids that are not in the packfile, we issue the normal command
to retrieve it from the object database.

### DIFF
--- a/builtin/cat-file.c
+++ b/builtin/cat-file.c
@@ -16,6 +16,7 @@
 #include "packfile.h"
 #include "object-store.h"
 #include "promisor-remote.h"
+#include "hashmap.h"
 
 enum batch_mode {
 	BATCH_MODE_CONTENTS,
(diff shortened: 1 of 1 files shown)


john-cai:
@avar when I tested this on my machine, I was seeing it perform 2x worse--I'm not sure why. When you get a chance, could you take a look as it might be obvious to you. thanks!

ThoonsanEain:
https://github.com/git/git/pull/1233/commits/15c821477a0f6dd9e3d65a322b7f34d637bc6cc3

ThoonsanEain:
https://github.com/git/git/pull/1233/commits/15cm

PR #1235, opened 2022-04-13
setup: fix safe.directory key not being checked

Matheus28:
It seems that nothing is ever checking to make sure the safe directories in the configs actually have the key safe.directory, so some unrelated config that has a value with a certain directory would also make it a safe directory.

This is a fix to https://github.com/git/git/commit/8959555cee7ec045958f9b6dd62e541affb7e7d9

### DIFF
--- a/setup.c
+++ b/setup.c
@@ -1098,6 +1098,9 @@ struct safe_directory_data {
 
 static int safe_directory_cb(const char *key, const char *value, void *d)
 {
+	if (strcmp(key, "safe.directory"))
+		return 0;
+
 	struct safe_directory_data *data = d;
 
 	if (!value || !*value)


Matheus28:
@dscho since this is a fix to a patch of yours, I'm @'ing you here since I don't have permissions with the bot yet. Could you /allow me?

db48x:
I guess you’ll want to move line 1104 back to the first line of the function where it “belongs”.

rimrul:
Please move the `struct safe_directory_data *data = d;` line up to be the first line of `safe_directory_cb()` and ammend your commit with a sign-off, then force push this branch.

Matheus28:
Sounds good, will do it tomorrow morning

Matheus28:
@derrickstolee beat me to it, so I'll be closing this pull request. Thank you!

derrickstolee:
Hi @Matheus28: I adapted your commit into a larger series here: https://github.com/gitgitgadget/git/pull/1215

The biggest issue is that we didn't have tests for this area, and now we do. I also wanted to add an extra "*" value to allow a blanket opt-out of the feature.

derrickstolee:
Oh, @Matheus28: could you confirm that it is OK for me to user your sign-off? The second commit has your authorship and this line:

```
Signed-off-by: Matheus Valadares <me@m28.io>
```

Please see [this portion of the documentation](https://github.com/git/git/blob/11cfe552610386954886543f5de87dcc49ad5735/Documentation/SubmittingPatches#L225-L283) for what this means.

Matheus28:
That is fine

PR #1253, opened 2022-04-20
git-p4: Properly initialize the largeFiles set so that it includes the already existing large files

tumata:
`git-p4 sync` isn't taking into account existing files that are managed by git-lfs. This is fine when doing `git p4 clone` but doing `git p4 sync` needs to take into account the files that are already managed by git-lfs.

This change reads-in the .gitattributes and re-generate the data so that sync can iteratively add files to the git-lfs. It takes care of initializing the largeFiles with the already existing files referenced in the .gitattributes instead of initializing with an empty set by default.

We had issues cloning the first commit, then syncing the remaining commits of a really large Perforce repo (making hundreds of smaller `sync` instead of one massive `sync`). The sync was deleting the git-lfs each time. We were able to fix this issue by first reading the current content of git-lfs instead of initializing to an empty set.

Signed-off-by: tumata <5684571+tumata@users.noreply.github.com>

### DIFF
--- a/git-p4.py
+++ b/git-p4.py
@@ -1195,8 +1195,27 @@ class LargeFileSystem(object):
     """Base class for large file system support."""
 
     def __init__(self, writeToGitStream):
-        self.largeFiles = set()
+        self.largeFiles = self.parseLargeFiles()
         self.writeToGitStream = writeToGitStream
+
+    def parseLargeFiles(self):
+        """Parse large files in order to initially populate 'largeFiles'"""
+        paths = set()
+        try:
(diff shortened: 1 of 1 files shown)


phil-blain:
Hi @tumata, the Git projects requires your real email address be used when creating commits. This is why GitGitGadget complains. You can use `git commit --amend --reset-author` on the command line to fix this.

tumata:
Hi there 👋  
Thanks for the tip @phil-blain, I updated my PR use now. I'll proceed with GitGitGadget once everything looks right.

tumata:
I updated my git commit message to follow the requested standard. Ready for /submit

PR #1264, opened 2022-05-12
grep: add --max-count command line option

00xc:
This patch adds a command line option analogous to that of GNU
grep(1)'s -m / --max-count, which users might already be used to.
This makes it possible to limit the amount of matches shown in the
output while keeping the functionality of other options such as -C
(show code context) or -p (show containing function), which would be
difficult to do with a shell pipeline (e.g. head(1)).

Signed-off-by: Carlos López <00xc@protonmail.com>
cc: Martin Ågren <martin.agren@gmail.com>
cc: Paul Eggert <eggert@cs.ucla.edu>


Aliabdo112 (commented):
Okay

PR #1266, opened 2022-05-16
fetch: limit shared symref check only for local branches

orgads:


orgads:
@dscho You're the original author. Can you please review?

dscho:
> dscho You're the original author. Can you please review?

Seems that you already got good reviews.

PR #1270, opened 2022-05-27
ssh signing: better error message when key not in agent

adaszko:
When signing a commit with a SSH key, with the private key missing from
ssh-agent, a confusing error message is produced:

    error: Load key "/var/folders/t5/cscwwl_n3n1_8_5j_00x_3t40000gn/T//.git_signing_key_tmpkArSj7": invalid format?
    fatal: failed to write commit object

The temporary file .git_signing_key_tmpkArSj7 created by git contains a
valid *public* key.  The error message comes from `ssh-keygen -Y sign'
and is caused by a fallback mechanism in ssh-keygen whereby it tries to
interpret .git_signing_key_tmpkArSj7 as a *private* key if it can't find
in the agent [1].  A fix is scheduled to be released in OpenSSH 9.1. All
that needs to be done is to pass an additional backward-compatible
option -U to 'ssh-keygen -Y sign' call.  With '-U', ssh-keygen always
interprets the file as public key and expects to find the private key in
the agent.

As a result, when the private key is missing from the agent, a more
accurate error message gets produced:

    error: Couldn't find key in agent

[1] https://bugzilla.mindrot.org/show_bug.cgi?id=3429

cc: Adam Szkoda <adaszko@gmail.com>
cc: Fabian Stelzer <fs@gigacodes.de>

### DIFF
--- a/gpg-interface.c
+++ b/gpg-interface.c
@@ -998,6 +998,7 @@ static int sign_buffer_ssh(struct strbuf *buffer, struct strbuf *signature,
 	char *ssh_signing_key_file = NULL;
 	struct strbuf ssh_signature_filename = STRBUF_INIT;
 	const char *literal_key = NULL;
+	int literal_ssh_key = 0;
 
 	if (!signing_key || signing_key[0] == '\0')
 		return error(
@@ -1005,6 +1006,7 @@ static int sign_buffer_ssh(struct strbuf *buffer, struct strbuf *signature,
 
 	if (is_literal_ssh_key(signing_key, &literal_key)) {
 		/* A literal ssh key */
+		literal_ssh_key = 1;
 		key_file = mks_tempfile_t(".git_signing_key_tmpXXXXXX");
 		if (!key_file)
 			return error_errno(
@@ -1039,8 +1041,10 @@ static int sign_buffer_ssh(struct strbuf *buffer, struct strbuf *signature,
 		     "-Y", "sign",
 		     "-n", "git",
 		     "-f", ssh_signing_key_file,
-		     buffer_file->filename.buf,
 		     NULL);
+	if (literal_ssh_key)
+		strvec_push(&signer.args, "-U");
+	strvec_push(&signer.args, buffer_file->filename.buf);
 
 	sigchain_push(SIGPIPE, SIG_IGN);
 	ret = pipe_command(&signer, NULL, 0, NULL, 0, &signer_stderr, 0);


adaszko:
My email address should now be public and the CI failures in this PR are unrelated to my changes.  Is there anything else I need to do?

dscho:
Looks good to me, next step is `/submit`

adaszko:
Hi.  The patch on the OpenSSH side has been released in [OpenSSH 9.1](https://www.openssh.com/txt/release-9.1):

> * ssh-keygen(1): improve error message when 'ssh-keygen -Y sign' is
>   unable to load a private key; bz3429
> * ssh-keygen(1): allow the existing -U (use agent) flag to work
>   with "-Y sign" operations, where it will be interpreted to require
>   that the private keys is hosted in an agent; bz3429

Can this PR be scheduled to be merged in the next git release?

gpg-interface.c
```diff
@@ -1019,6 +1021,7 @@ static int sign_buffer_ssh(struct strbuf *buffer, struct strbuf *signature,
 		ssh_signing_key_file = strbuf_detach(&key_file->filename, NULL);
 	} else {
 		/* We assume a file */
+		literal_ssh_key = 0;
```
phillipwood, line 1024: As we initialize `literal_ssh_key` to zero when we declare it I don't think we need this line.

gpg-interface.c
```diff
@@ -1035,12 +1038,23 @@ static int sign_buffer_ssh(struct strbuf *buffer, struct strbuf *signature,
...
+			NULL);
+	}
+	else {
+	    strvec_pushl(&signer.args, use_format->program,
+			"-Y", "sign",
+			"-n", "git",
+			"-f", ssh_signing_key_file,
+			buffer_file->filename.buf,
+			NULL);
+	}
```
phillipwood, line 1057: Rather than duplicating everything, it would be nicer to do

    strvec_pushl(&signer.args, /* options */, NULL);
    if (literal_ssh_key)
        strvec_push(&signer.args, "-U");
    strvec_push(buffer_file->filename.buf);

gpg-interface.c
adaszko: Removed that line 👍

gpg-interface.c
adaszko: Corrected

phillipwood:
This is looking good, let's submit this to the list so it can be reviewed there and hopefully merged soon.

PR #1274, opened 2022-06-07
config: document and test the 'worktree' scope

chooglen:
While I was digging through the docs on config scopes for
discovery.bare, I noticed that Documentation/git-config.txt doesn't
mention the 'worktree' scope, but the usage string does.

One might assume that since we have CONFIG_SCOPE_WORKTREE, the omission
from Documentation/git-config.txt is obviously accidental. But, when we
first added "--show-scope" in 145d59f (config: add '--show-scope' to
print the scope of a config value, 2020-02-10), we noted that
'submodule' would never be output from "--show-scope", even though
CONFIG_SCOPE_SUBMODULE exists.

This patch adds a test that shows "git config -l --show-scope" can
indeed output 'worktree', and adds the 'worktree' scope to
Documentation/git-config.txt.

This does not conflict with gc/bare-repo-discovery.

Cc: Matthew Rogers <mattr94@gmail.com>, Philip Oakley <philipoakley@iee.email>

### DIFF
--- a/Documentation/git-config.txt
+++ b/Documentation/git-config.txt
@@ -248,7 +248,7 @@ Valid `<type>`'s include:
 --show-scope::
 	Similar to `--show-origin` in that it augments the output of
 	all queried config options with the scope of that value
-	(local, global, system, command).
+	(worktree, local, global, system, command).
 
 --get-colorbool <name> [<stdout-is-tty>]::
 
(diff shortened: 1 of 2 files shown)


prasertKealampat (commented):
@[Documentation/git-config.txt](Documentation/git-config.txt) #1210

saa18877433:
في الثلاثاء، ١٤ يونيو ٢٠٢٢ ٢:٠٣ ص gitgitgadget-git[bot] <
***@***.***> كتب:

dscho:
@saa18877433 Please use your own fork for practicing how to use GitHub.

PR #1275, opened 2022-06-09
[RFC] submodule update: parse all options in C

chooglen:
As a follow up to ar/submodule-update [1] and its successors
gc/submodule-update-part* [2] [3], this series converts the last remaining
piece of "git submodule update" into C, namely, the option parsing in
git-submodule.sh.

As a result, git-submodule.sh::cmd_update() is now an (almost)
one-liner:

  cmd_update()
  {
    git ${wt_prefix:+-C "$wt_prefix"} submodule--helper update \
      ${wt_prefix:+--prefix "$wt_prefix"} \
      "$@"
  }

and best of all, "git submodule update" now shows a usage string for its
own subcommand instead of a giant usage string for all of "git
submodule" :)

Given how many options "git submodule update" accepts, this series takes
a gradual approach:

1. For each option, if "git submodule--helper update" already
   understands the string option, just pass the string option instead of
   setting a temporary 'boolean' variable (1/6).
2. Otherwise, if the option makes sense, teach "git submodule--helper
   update" to understand the option. Goto 1. (2-3/6).
3. Otherwise, if the option makes no sense, drop it (4/6).
4. When we've processed all options, delete all the option parsing code
   (5/6) and clean up (6/6).

I read over the other "git submodule" subcommands very briefly to see
how much work this would be if we were to repeat the process for all
subcommands. It should be relatively simple - most "git
submodule--helper" commands already understand every CLI option that
they need to, so we can skip step 2 (i.e. the tedious one).

I'm sending this as an RFC because even though this series is pretty
low-risk, the previous topics [1] [2] [3] had their fair share of
regressions (even when the conversions seemed 'obviously correct') and
showed that the test coverage of "git submodule "update" is quite poor.
I'd like to get feedback on the general approach before I invest more
time into improving the test coverage and/or checking the before/after
output for regressions [4].

[1] https://lore.kernel.org/git/20210907115932.36068-1-raykar.ath@gmail.com/
[2] Part 1 https://lore.kernel.org/git/20220305001401.20888-1-chooglen@google.com/
[3] Part 2 https://lore.kernel.org/git/20220315210925.79289-1-chooglen@google.com/
[4] Ævar suggested a way to do this using the test suite and "--tee" in
https://lore.kernel.org/git/220214.86h791xsza.gmgdl@evledraar.gmail.com/


chooglen:
Ævar has taken over this work :)

sfdfsdfsdf:
Ah that might explain some things. I'm still not sure how removing
SUPPORT_SUPER_PREFIX gives the result we see here, but here are some
observations using the extra info I learned while converting
"--recursive-prefix" -> "--super-prefix":

- 29a5e9e1ffe and its parent teach init_submodule() how to use the
  "--recursive-prefix" flag as a substitute for "--super-prefix".
  "--recursive-prefix" is used exclusively by "git submodule update" and
  does exactly the same thing as "--super-prefix".

  Those commits pass "--recursive-prefix" to
  do_get_submodule_displaypath(), which is an escape hatch to
  get_submodule_displaypath() (which always uses "--super-prefix").

- 75df9df0f81 introduced a regression because we now pass
  update_data.prefix to get_submodule_displaypath() instead of prefix,
  and update_data.prefix was never set (oops, that's my mistake). This
  was rectified in its child (b3c5f5cb048), where we start to set
  update_data.prefix when the two subcommands were combined.

  As a sidenote, in 75df9df0f81's diff, we see that recursive_prefix is
  used in conjunction with get_submodule_displaypath(), which means that
  we could theoretically use *both* "--recursive-prefix" AND
  "--super-prefix" at the same time. We don't do that though, because
  "--super-prefix" was never used by "git submodule update".

  (This prefixed_path stuff is going away when I convert
  "--recursive-prefix" to "--super-prefix", so don't feel the need to
  clean it up).

PR #1276, opened 2022-06-15
push: Fix capitalisation of the option name autoSetupMerge

fangyi-zhou:
This was found during l10n process by Jiang Xin.

Signed-off-by: Fangyi Zhou <me@fangyi.io>


dscho:
> `push: Fix capitalisation of the option name autoSetupMerge`
> Prefixed commit message must be in lower case

This means that GitGitGadget suggests "push: fix [...]" instead of "push: Fix [...]".

PR #1279, opened 2022-06-20
i18n: mark message helpers prefix for translation

DJm00n:
Some messages prefixes like 'usage:'/'fatal:'/'warning:'/'error:'
were not translated.

CC: Jiang Xin <zhiyou.jx@alibaba-inc.com>, Junio C Hamano <gitster@pobox.com>
cc: Jiang Xin <worldhello.net@gmail.com>

### DIFF
--- a/usage.c
+++ b/usage.c
@@ -33,7 +33,7 @@ static void vreportf(const char *prefix, const char *err, va_list params)
 
 static NORETURN void usage_builtin(const char *err, va_list params)
 {
-	vreportf("usage: ", err, params);
+	vreportf(_("usage: "), err, params);
 
 	/*
 	 * When we detect a usage error *before* the command dispatch in
@@ -58,7 +58,7 @@ static NORETURN void usage_builtin(const char *err, va_list params)
 static void die_message_builtin(const char *err, va_list params)
 {
 	trace2_cmd_error_va(err, params);
-	vreportf("fatal: ", err, params);
+	vreportf(_("fatal: "), err, params);
 }
 
 /*
@@ -78,14 +78,14 @@ static void error_builtin(const char *err, va_list params)
 {
 	trace2_cmd_error_va(err, params);
 
-	vreportf("error: ", err, params);
+	vreportf(_("error: "), err, params);
 }
 
 static void warn_builtin(const char *warn, va_list params)
 {
 	trace2_cmd_error_va(warn, params);
 
-	vreportf("warning: ", warn, params);
+	vreportf(_("warning: "), warn, params);
 }
 
 static int die_is_recursing_builtin(void)


DJm00n:
@jiangxin can you please approve?

jiangxin:
@DJm00n , the prefix of the subject line should be "i18n:", not "l10n:", for you are marking strings for translation. And the subject line of your commit should not end with period.

dscho:
> the display width of the prefix is calculated using `strlen`, but fails for localized prefixes.

@DJm00n The display width is not the number of characters, there Unicode characters that are wider than others. See https://en.wikipedia.org/wiki/Halfwidth_and_fullwidth_forms for details.

This is not limited to Asian characters, even. Take for example "０", which is the fullwidth form of "0".

@jiangxin Having said that, are you sure that `vreportf()` uses `strlen()` to determine the _display_ width? To me, it looks more as if it tries to determine the number of bytes.

jiangxin:
> 



> > the display width of the prefix is calculated using `strlen`, but fails for localized prefixes.
> 
> @DJm00n The display width is not the number of characters, there Unicode characters that are wider than others. See https://en.wikipedia.org/wiki/Halfwidth_and_fullwidth_forms for details.
> 
> This is not limited to Asian characters, even. Take for example "０", which is the fullwidth form of "0".
> 
> @jiangxin Having said that, are you sure that `vreportf()` uses `strlen()` to determine the _display_ width? To me, it looks more as if it tries to determine the number of bytes.

My mistake. `strlen` is used to count the number of bytes to fill prefix in the msg string, this patch is OK.

dscho:
> /submit

@DJm00n please note that the Git mailing list regulars frown upon seeing iterations of patch series being submitted too frequently. The preference I saw was to leave at least one day for reviewers to provide feedback.

In this instance, I think it was merited, and the range-diff might explain why you did it, but it would be a wonderful thing if you could add a note to your cover letter when sending subsequent iterations what necessitated the new iteration (edit the first comment in this PR for that, it will be used as "cover letter" between commit message and diffstat).

DJm00n:
@dscho sorry for that. This is my first contribution via gitgitgadget. Their docs are a little confusing to me.

>edit the first comment in this PR for that, it will be used as "cover letter" between commit message and diffstat

Thank you for suggestion. This was not obvious to me. I thought that there is some params for `/submit` instead. Also I was surpriced that it automagically adds `v2` to next submits and diffs commit desc. Nice.

dscho:
> This is my first contribution via gitgitgadget. Its docs are a little confusing to me.

Those docs would be my fault, sorry! If you have any ideas how to improve them, I'd love to have your help.

> > edit the first comment in this PR for that, it will be used as "cover letter" between commit message and diffstat
> 
> Thank you for suggestion. This was not obvious to me. I thought that there is some params for `/submit` instead.

If you have any ideas how to make this more obvious, that would be awesome.

PR #1282, opened 2022-06-27
submodule: remove "--recursive-prefix"

chooglen:
Thanks Ævar for the feedback! I've incorporated all of the suggestions,
except breaking up the tests in 1/6 into their own patch - I wasn't
convinced of the value of having prescriptive tests (with
test_expect_failure), and I found it difficult to discuss descriptive
tests (with test_expect_success) without also having the code in the
same diff. FWIW, this version (and the previous one) "git rebase -x
'make test'" cleanly :)

Note for Junio: this version is based on ab/submodule-cleanup (and so
will future versions).
  
= Description

This series is a refactor of "git submodule--helper update" that
replaces "--recursive-prefix" with "--super-prefix" (see Background).
This was initially motivated by:

- Junio's suggestion to simplify the code [1] (in response to a memory
  leak found by Johannes Schindelin [2]).
- A desire to use the module_list API + get_submodule_displaypath()
  outside of builtin/submodule--helper.c (I expect to use this to check
  out branches in each submodule).

But it also happens to remove some overly complicated/duplicated code
that was literally converted from shell :)

= Background 

When recursing into nested submodules, Git commands keep track of the
path from the superproject to the submodule in order to print a "display
path" to the user, e.g.

  Submodule path '../super/sub/nested-sub': checked out 'abcdef'

For historical reasons, "git submodule--helper update" uses
"--recursive-prefix" for this purpose, but it should use
"--super-prefix" instead because:

- That's what every other command uses (not just submodule--helper
  subcommands).
- Using the "--super-prefix" helper functions makes the "git
  submodule--helper update" code simpler

= Patch organization

- Patch 1/6 makes sure that display paths are only computed using
  display path helper functions ([do_]get_submodule_displaypath()) and
  fixes some display paths that we never realized were broken.
- Patches 2-3/6 simplify and deduplicate some display path computations.
- Patch 4/6 (authored by Ævar) removes SUPPORT_SUPER_PREFIX where it's
  not needed.
  - This doesn't affect correctness, but we want to do this eventually,
    so do it now to make 5/6 a bit cleaner.
- Patch 5/6 replaces "--recursive-prefix" with "--super-prefix", making
  do_get_submodule_displaypath() obsolete.
  - GGG outputs an odd diff; I recommend viewing it with "--histogram"
- Patch 6/6 removes do_get_submodule_displaypath().

= Series history

Changes in v2:
- Rebase onto ab/submodule-cleanup (previously master)
- Cherry pick https://github.com/avar/git/commit/f445c57490d into 4/6
- Style fixes in .c and tests

= Future work

At the end of this series, get_submodule_displaypath() can be moved to
submodule.h, which would make submodule.c:get_super_prefix_or_empty()
obsolete. I have a patch that demonstrates this (CI run: [4]), but I
decided to keep this series more focused on "git submodule--helper
update" so that it's easier to review.

[1] https://lore.kernel.org/git/xmqq35g5xmmv.fsf@gitster.g
[2] https://lore.kernel.org/git/877a45867ae368bf9e053caedcb6cf421e02344d.1655336146.git.gitgitgadget@gmail.com
[3] https://lore.kernel.org/git/patch-v3-02.12-082e015781e-20220622T142012Z-avarab@gmail.com
[4] https://github.com/chooglen/git/actions/runs/2572557584

Cc: "Atharva Raykar" <raykar.ath@gmail.com>, "Ævar Arnfjörð Bjarmason" <avarab@gmail.com>, Johannes Schindelin <johannes.schindelin@gmx.de>


Shahbazahammed:
Close for my account

PR #1283, opened 2022-06-29
mv: fix out-of-cone file/directory move logic

ffyuanda:
Changes since [PATCH v4](https://lore.kernel.org/git/20220623114120.12768-1-shaoxuan.yuan02@gmail.com/)

1. Added t1092 test for `mv` directory from out-of-cone to in-cone, this is suggested by Derrick [here](https://lore.kernel.org/git/17a34570-f823-7908-1ea5-8e7bf8680b23@github.com/).

2. Fixed style nits.

Looking forward to hearing your reviews! @vdye @derrickstolee

### DIFF
--- a/builtin/mv.c
+++ b/builtin/mv.c
@@ -20,6 +20,13 @@ static const char * const builtin_mv_usage[] = {
 	NULL
 };
 
+enum update_mode {
+	BOTH = 0,
+	WORKING_DIRECTORY = (1 << 1),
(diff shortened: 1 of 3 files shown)


derrickstolee:
@ffyuanda: If you were intending to create a PR for early (not on-list) review, then you might want to create this PR in your fork. You can use another one of your branches as a base so we can see only the commits that are new to that PR and not already being discussed upstream.

PR #1285, opened 2022-07-05
git-p4: Fix bug with encoding of P4 client name

cohomology:
CC: Tao Klerks <tao@klerks.biz>

### DIFF
--- a/git-p4.py
+++ b/git-p4.py
@@ -854,12 +854,12 @@ def p4CmdList(cmd, stdin=None, stdin_mode='w+b', cb=None, skip_info=False,
             if bytes is not str:
                 # Decode unmarshalled dict to use str keys and values, except for:
                 #   - `data` which may contain arbitrary binary data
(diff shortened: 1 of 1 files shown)


cohomology:
@TaoK:
@DJm00n : 

Can you allow me?

SpaldingGreene:
PR #1285- Client name is- Bradley Lee green

cohomology:
Sorry, I was in vacation. I tried to refactor the coding a bit, I hope it is more clear now. Not sure about the camel- vs. underscore naming convention in git-p4.py

cohomology:
Actually I do not know how to clean up my mess. My original commit contained two different email addresses of mine, which was correctly complained by Junio. The only possiblity was a force commit which is not allowed either ;-). I also lost my original commit which can only be seen in the github user interface but can't be checked out. I now try to do an additional commit of a cherry-pick taken from the seen branch.

PR #1286, opened 2022-07-07
Do not use CURLOPT_RESOLVE for libcurl < 7.21.3

atoomic:
Starting with 2.37.0 via 511cfd3bffa685fda0e7c25bfa08082aa0de3a30
CURLOPT_RESOLVE was added to http.c.

But as described by https://curl.se/libcurl/c/CURLOPT_RESOLVE.html
> this was "Added in 7.21.3. Removal support added in 7.42.0."
not all curl version provide this feature.

Note: http.curloptresolve need to be marked as unsupported when
using an older curl version.

### DIFF
--- a/git-curl-compat.h
+++ b/git-curl-compat.h
@@ -42,6 +42,12 @@
 #define GITCURL_HAVE_CURLOPT_TCP_KEEPALIVE 1
 #endif
 
+/**
+ * CURLOPT_RESOLVE was added in 7.21.3, released in December 15 2010
+ */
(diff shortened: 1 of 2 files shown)


dscho:
Isn't this essentially the same as https://github.com/gitgitgadget/git/pull/1277?

atoomic:
Sorry for the noise and I agree this is a dup of https://github.com/gitgitgadget/git/pull/1277
I ve not noticed it earlier, we can close this case

PR #1294, opened 2022-07-19
git-p4: fix crlf handling for utf16 files on Windows

mbs-c:
CC: Tao Klerks <tao@klerks.biz>, Junio C Hamano <gitster@pobox.com>


fvoidCN:
Can someone allow me to use gitgitgadget?

PR #1295, opened 2022-07-19
Add git switch suggestion in checkout error hint

victorlin:
It would be nice suggest a `git switch` command when facing the error from running `git switch <name>` where `<name>` matches multiple remote tracking branches.

---

### DIFF
--- a/builtin/checkout.c
+++ b/builtin/checkout.c
@@ -1224,6 +1224,10 @@ static const char *parse_remote_branch(const char *arg,
 			     "\n"
 			     "    git checkout --track origin/<name>\n"
 			     "\n"
+			     "or\n"
+			     "\n"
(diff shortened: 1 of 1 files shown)


dscho:
> Invalid author email in [8271505](https://github.com/git/git/commit/8271505d7bdf97ff23abd669c3096d75fd8fb252): "[13424970+victorlin@users.noreply.github.com](mailto:13424970+victorlin@users.noreply.github.com)"

@victorlin This means that GitGitGadget tried to point out (before any human reviewer even had a look) that the Git project [requires the real name and the real email address to be used](https://git-scm.com/docs/SubmittingPatches#sign-off). The reason is that you might be contacted in case some follow-up work touches the same parts as your patches. And that simply requires your real information to be available. Otherwise the Git project cannot reach you.

PR #1297, opened 2022-07-20
git-p4: fix two undefined variables

mbs-c:
CC: Luke Diamand <luke@diamand.org>

### DIFF
--- a/git-p4.py
+++ b/git-p4.py
@@ -4369,19 +4369,16 @@ class P4Unshelve(Command):
     def renameBranch(self, branch_name):
         """Rename the existing branch to branch_name.N ."""
 
-        found = True
         for i in range(0, 1000):
             backup_branch_name = "{0}.{1}".format(branch_name, i)
             if not gitBranchExists(backup_branch_name):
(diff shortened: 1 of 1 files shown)


git-p4.py
```diff
@@ -4369,19 +4369,16 @@ def __init__(self):
...
                 gitUpdateRef(backup_branch_name, branch_name)
                 gitDeleteRef(branch_name)
-                found = True
                 print("renamed old unshelve branch to {0}".format(backup_branch_name))
                 break
-
-        if not found:
-            sys.exit("gave up trying to rename existing branch {0}".format(sync.branch))
+        else:
+            sys.exit("gave up trying to rename existing branch {0}".format(branch_name))
```
rimrul, line 4381: I don't see any `if` for this new `else`. I think the correct fix here is to just replace `sync.branch` with `branch_name` and not change anything else.

rimrul: Ok, weird construct, but it seems to exist and work in the way you describe. No objections then.

PR #1300, opened 2022-07-27
WIP: Blame Trees

john-cai:

### DIFF
--- a/blame.c
+++ b/blame.c
@@ -1,4 +1,6 @@
 #include "cache.h"
+
+#include "pathspec.h"
 #include "refs.h"
 #include "object-store.h"
 #include "cache-tree.h"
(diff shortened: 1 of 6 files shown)


blame.c
```diff
@@ -1381,22 +1398,25 @@ static struct blame_origin *find_origin(struct repository *r,
 		       PATHSPEC_LITERAL_PATH, "", paths);
 	diff_setup_done(&diff_opts);
 
-	if (is_null_oid(&origin->commit->object.oid))
+	if (is_null_oid(&origin->commit->object.oid)) {
 		do_diff_cache(get_commit_tree_oid(parent), &diff_opts);
```
john-cai, line 1408: @pks-t I think something is going wrong here. I can't quite follow how the diff is happening here, but it feels we need to change something here so that it diffs the tree object content instead of the blobs.

PR #1309, opened 2022-08-18
trailer: allow spaces in tokens

tekknolagi:
The docs (for example, https://git-scm.com/docs/git-interpret-trailers)
say that whitespace should be allowed inside tokens:

> There can also be whitespaces inside the token and the value.

The code since e4319562bc2834096fade432fd90c985b96476db does not allow
that, so re-enable and add a test.

Signed-off-by: Max Bernstein <max@bernsteinbear.com>


dscho:
> /preview

Please try again.

dscho:
[There you go](https://dev.azure.com/gitgitgadget/git/_build/results?buildId=149353&view=results)!

PR #1313, opened 2022-08-23
grep: fix multibyte regex handling under macOS

dspinellis:
The 2013 commit 29de20504e9 fixed t0070-fundamental.sh under Darwin
(macOS) by adopting Git's regex library.  However, this library is
compiled with NO_MBSUPPORT, which causes git-grep to work incorrectly
on multibyte (e.g. UTF-8) files.  Current macOS versions pass
t0070-fundamental.sh with the native macOS regex library, which
also supports multibyte characters.

Adjust the Makefile to use the native regex library, and call
setlocale(3) to set CTYPE according to the user's preference.  The
setlocale(3) call is required on all platforms, but in platforms
supporting gettext(3), setlocale(3) is called as a side-effect of
initializing gettext(3).

To avoid running the new tests on platforms still using the
compatibility library, which is compiled without multibyte
support, store the corresponding NO_REGEX setting in the
GIT-BUILD-OPTIONS file.  This makes it available to the test
scripts.  In addition, adjust the test-tool regex command to
work with multibyte regexes to further test a platform's
support for them.

Signed-off-by: Diomidis Spinellis <dds@aueb.gr>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1427,7 +1427,6 @@ ifeq ($(uname_S),Darwin)
 		APPLE_COMMON_CRYPTO = YesPlease
 		COMPAT_CFLAGS += -DAPPLE_COMMON_CRYPTO
 	endif
-	NO_REGEX = YesPlease
 	PTHREAD_LIBS =
 endif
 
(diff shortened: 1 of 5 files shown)


dscho:
@dspinellis I fear that this patch is not quite as platform-specific as the PR title claims: it [breaks the tests on Windows](https://github.com/git/git/runs/7972935856?check_suite_focus=true#step:5:201).

dscho:
> it [breaks the tests on Windows](https://github.com/git/git/runs/7972935856?check_suite_focus=true#step:5:201).

... and [on 32-bit Linux, too](https://github.com/git/git/runs/7972783545?check_suite_focus=true#step:5:1763).

It probably has nothing to do with Windows or Linux, but more with what PCRE2 version is used/not used and with what options it was compiled. Or maybe it's `regex()` dependent.

dscho:
@dspinellis would you kindly also edit [the initial comment](https://github.com/git/git/pull/1313#issue-1347666512), which is used as a cover letter by GitGitGadget? If in doubt, send a preview to yourself using the `/preview` command.

dspinellis:
@dscho Thank you for the `/allow`. The failing tests are the ones I introduced to check this fix. They expose a pre-existing issue, and need to be skipped on the specific configurations until these are also fixed (if possible). To do that I need to better understand the corresponding build configurations.

dscho:
@dspinellis if you're familiar with SSH/tmux, I recommend using https://github.com/mxschmitt/action-tmate/ to debug interactively. Note: There have been reports that using this Action might flag new accounts as abusing GitHub's services, so please do use it judiciously.

dspinellis:
Switching from [gitgitgadget](https://github.com/apps/gitgitgadget-git) to [send-email](https://git-scm.com/docs/git-send-email). Further exchanges will appear only in the list thread.

dscho:
Would you mind closing the PR, then?

PR #1324, opened 2022-09-12
chainlint: colorize problem annotations and test delimiters

sunshineco:
This is a re-roll of [1] which colorizes the output of "chainlint.pl"
when it detects problems in Git test definitions. During discussion, it
was noted that the eye could sometimes glide right over[2] the bold-red
"?!FOO?!" annotations, so Junio suggested using reverse video, which is
what v2 does.

Reverse video certainly makes the "?!FOO?!" annotations pop out and draw
the reader's attention. I find that I don't have a strong preference
between this version and v1 which merely used bold-red, but I suspect
that v2 with its reverse video is probably the better approach.

[1] https://lore.kernel.org/git/pull.1324.git.git.1663023888412.gitgitgadget@gmail.com/
[2] https://lore.kernel.org/git/CAPig+cTq3j5M7cz3T14h9U6e+H5PAu8JJ_Svq87W3WviwS6_qA@mail.gmail.com/
[3] https://lore.kernel.org/git/xmqqo7vkazuh.fsf@gitster.g/

Cc: Jeff King <peff@peff.net>
Cc: Junio C Hamano <gitster@pobox.com>

### DIFF
--- a/t/chainlint.pl
+++ b/t/chainlint.pl
@@ -585,12 +585,14 @@ sub check_test {
 	my $parser = TestParser->new(\$body);
 	my @tokens = $parser->parse();
 	return unless $emit_all || grep(/\?![^?]+\?!/, @tokens);
+	my $c = main::fd_colors(1);
 	my $checked = join(' ', @tokens);
 	$checked =~ s/^\n//;
 	$checked =~ s/^ //mg;
 	$checked =~ s/ $//mg;
+	$checked =~ s/(\?![^?]+\?!)/$c->{rev}$c->{red}$1$c->{reset}/mg;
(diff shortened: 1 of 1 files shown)


Nice7118:
This is what I need. I need to track everyone of these developers. They
need to know that they are not aloud to do what they are doing to my
devices! They need to stop with the fake litigation and information. They
can not pull data from my devices m they are not aloud to remove documents
. They do not have legal permission to have illegal monitoring on my
devices! They have set up communications web versions on my devices. They
installed a fake Google mirror. They are cloning apps and setting data
transfers. They set up something that wares down my battery. They are able
to lock my phone to where I can never use it. They can turn it off. They
can switch users. Not onemof the have permission .

PR #1333, opened 2022-09-17
unicode: update the width tables to Unicode 15

fcharlie:
Unicode 15 Released[0] and including several additional zero/double width characters.

[0] https://home.unicode.org/announcing-the-unicode-standard-version-15-0/

unicode-width.h regenerated by contrib/update-unicode/update_unicode.sh


fcharlie:
In fact, I re-run contrib/update-unicode/update_unicode.sh, 
which downloads the latest Unicode data and then generates 
a new unicode-width.h. The process can be retried.

dscho:
> In fact, I re-run contrib/update-unicode/update_unicode.sh, which downloads the latest Unicode data and then generates a new unicode-width.h. The process can be retried.

@fcharlie please reply on the list, as instructed in https://github.com/gitgitgadget/gitgitgadget/wiki/ReplyToThis

fcharlie:
> > In fact, I re-run contrib/update-unicode/update_unicode.sh, which downloads the latest Unicode data and then generates a new unicode-width.h. The process can be retried.
> 
> @fcharlie please reply on the list, as instructed in https://github.com/gitgitgadget/gitgitgadget/wiki/ReplyToThis

Both my Outlook email and qq email were blocked, and I ended up using charlieio@icloud.com to reply to this

my reply: 
https://lore.kernel.org/git/A502C1AE-CD49-46FB-B5ED-0513BABF7AD1@icloud.com/

PR #1335, opened 2022-09-19
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

`check-whitespace.yml` is triggered by `pull_request` only and receives restricted token already.
`l10n.yml` has permissions on  job level already. So I didn't make any changes to them.
In both cases it is possible to add explicit global lever permissions just for consistency if you prefer. Let me know.

### DIFF
--- a/builtin/fetch.c
+++ b/builtin/fetch.c
@@ -1617,9 +1617,21 @@ static int do_fetch(struct transport *transport,
 				break;
 			}
 		}
-	} else if (transport->remote && transport->remote->fetch.nr)
-		refspec_ref_prefixes(&transport->remote->fetch,
-				     &transport_ls_refs_options.ref_prefixes);
+	} else {
+		struct branch *branch = branch_get(NULL);
+
+		if (transport->remote->fetch.nr)
+			refspec_ref_prefixes(&transport->remote->fetch,
+					     &transport_ls_refs_options.ref_prefixes);
+		if (branch_has_merge_config(branch) &&
+		    !strcmp(branch->remote_name, transport->remote->name)) {
+			int i;
+			for (i = 0; i < branch->merge_nr; i++) {
+				strvec_push(&transport_ls_refs_options.ref_prefixes,
+					    branch->merge[i]->src);
+			}
+		}
+	}
 
 	if (tags == TAGS_SET || tags == TAGS_DEFAULT) {
 		must_list_refs = 1;
--- a/t/t5520-pull.sh
+++ b/t/t5520-pull.sh
@@ -218,6 +218,23 @@ test_expect_success 'fail if upstream branch does not exist' '
 	test_cmp expect file
 '
 
+test_expect_success 'fetch upstream branch even if refspec excludes it' '
+	# the branch names are not important here except that
+	# the first one must not be a prefix of the second,
+	# since otherwise the ref-prefix protocol extension
+	# would match both
+	git branch in-refspec HEAD^ &&
+	git branch not-in-refspec HEAD &&
+	git init -b in-refspec downstream &&
+	git -C downstream remote add -t in-refspec origin "file://$(pwd)/.git" &&
+	git -C downstream config branch.in-refspec.remote origin &&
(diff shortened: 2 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -5,8 +5,14 @@ on: [push, pull_request]
...
   DEVELOPER: 1
 
+permissions:
+  contents: read
+
 jobs:
   ci-config:
+    permissions:
+      contents: read
+      actions: read # for github.actions.getWorkflowRun
```
dscho, line 15: Did you mean to add a comment after the `#`?

.github/workflows/main.yml
```diff
@@ -5,8 +5,14 @@ on: [push, pull_request]
 env:
   DEVELOPER: 1
 
+permissions:
+  contents: read
+
 jobs:
   ci-config:
+    permissions:
+      contents: read
```
dscho, line 14: Is this necessary even if those permissions are already configured globally?

dscho (commented):
Thank you for working on Git's CI!

I see you hardened `main.yml`. what about `l10n.yml` and `check-whitespace.yml`?

Also, please follow the guidance in https://github.blog/2022-06-30-write-better-commits-build-better-projects/ to populate the commit message with useful content.

.github/workflows/main.yml
sashashura: Yes

.github/workflows/main.yml
sashashura: According to the documentation permissions are not merged, but replaced. By defining new permission for `actions` we need to repeat the permission for `contents`.

sashashura:
Made the email public.

sashashura:
> I see you hardened `main.yml`. what about `l10n.yml` and `check-whitespace.yml`?

`check-whitespace.yml` is triggered by `pull_request` only and receives restricted token already.
`l10n.yml` has permissions on  job level already.

In both cases it is possible to add explicit global lever permissions just for consistency if you prefer. Let me know.

dscho:
> > I see you hardened `main.yml`. what about `l10n.yml` and `check-whitespace.yml`?
> 
> `check-whitespace.yml` is triggered by `pull_request` only and receives restricted token already. `l10n.yml` has permissions on job level already.

Could you mention this in [the cover letter](https://github.com/git/git/pull/1335#issue-1377545934)? That will help other reviewers by answering a question they are prone to have.

Other than that, I think this is good to `/submit`!

dscho:
> [force-pushed](/git/git/compare/1949ebdc27c38d85093d9c8483ed031ec1b273c5..ca20a44bc5dbed583b355c4a9a3c9ebf901a883d) the patch-2 branch from [ `1949ebd` ](/git/git/commit/1949ebdc27c38d85093d9c8483ed031ec1b273c5) to [ `ca20a44` ](/git/git/commit/ca20a44bc5dbed583b355c4a9a3c9ebf901a883d)

@sashashura was that inadvertent?

sashashura:
Yes, I screwed my branch while doing rebase :(

sashashura:
I opened a new one https://github.com/git/git/pull/1337

PR #1339, opened 2022-09-22
l10n: fr: don't say that merge is "the default strategy"

alexhenrie:
The text of this message was changed in commit 71076d0edde43a7672a9a0f555753ff078602a64 to avoid making any suggestion about which strategy is better for the situation at hand. Update the Franch translation to match.

I originally submitted this change at https://github.com/jnavila/git/pull/27, but the maintainer has not responded.


alexhenrie:
I accidentally sent this pull request to git/git instead of git-l10n/git-po. Resubmitted at https://github.com/git-l10n/git-po/pull/660

PR #1348, opened 2022-09-26
tmp-objdir: do not closedir() when handling a signal

john-cai:
We have recently observed a Git process hanging around for weeks. A backtrace revealed that a git-receive-pack(1) process was deadlocked when trying to remove the quarantine directory "incoming." It turns out that the tmp_objdir API calls opendir(3) and closedir(3) to observe a directory's contents in order to remove all the contents before removing the directory itself. These functions are not async signal save as they allocate and free memory.

The fix is to avoid calling these functions when handling a signal in order to avoid a deadlock. The implication of such a fix however, is that temporary object directories may not get cleaned up properly when a signal is being handled. The tradeoff this fix is making is to prevent deadlocks at the cost of temporary object directory cleanup.

This is similar to 58d4d7f1c5 (2022-01-07 fetch: fix deadlock when cleaning up lockfiles in async signals) The difference being in this case, the cleanup of the files will not happen when handling a signal.

Changes since v3:

- Updated commit message to clarify that we still clean up the tmp dir on atexit

cc: Jeff King <peff@peff.net>

### DIFF
--- a/tmp-objdir.c
+++ b/tmp-objdir.c
@@ -18,7 +18,7 @@ struct tmp_objdir {
 
 /*
  * Allow only one tmp_objdir at a time in a running process, which simplifies
- * our signal/atexit cleanup routines.  It's doubtful callers will ever need
+ * our atexit cleanup routines.  It's doubtful callers will ever need
  * more than one, and we can expand later if so.  You can have many such
  * tmp_objdirs simultaneously in many processes, of course.
  */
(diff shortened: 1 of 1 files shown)


Manuel1885:
{{{HTMRRPT}}}

Manuel1885:
{{{HTMRRPT}}}

PR #1349, opened 2022-09-28
Minor Refactors: follow the best python practicer

AhzamAhmed6:
Refactor the code by removing the unnecessary code blocks and replace with the best alternatives of python which reduce the complexity of the code and increase its readability.


dscho:
> the message is too short

Please also follow the best practices when it comes to [commit messages](https://github.blog/2022-06-30-write-better-commits-build-better-projects/) 😉

PR #1351, opened 2022-09-29
test-lib: fix UTF8_NFD_TO_NFC prereq while using ZFS' UTF-8 normalization

winterqt:
The commit message is too long, I presume. I'm not really sure how to go about making it shorter, though -- suggestions appreciated.


winterqt:
Going to close this as this isn't the right solution (see the macOS failures), though I'd still appreciate a `/allow` so I can submit the proper solution ASAP :)

PR #1359, opened 2022-10-12
archive: Add --recurse-submodules to git-archive command

Alphadelta14:
This makes it possible to include submodule contents in an archive command.

The inspiration for this change comes from this Github thread, https://github.com/dear-github/dear-github/issues/214, with at least 160 👍🏻 's at the time of writing.
(I stumbled upon it because I wanted it as well).

I figured the underlying implementation wouldn't be too difficult with most of the plumbing already in place, so I decided
to add the relevant logic to the client git-archive command.

One of the trickier parts of this implementation involved teaching read_tree about submodules.
Some of the troublesome areas were still using the `the_repository` references to look up commit or tree or oid information.
I ended up deciding that `read_tree_fn_t` would probably be best off having a concrete repo reference
since it allows changing the context to a subrepo where needed (even though some of the usages did not need it specifically).

I am open to feedback since this is all quite new to me :) 

TODO:
- [x] working implementation
- [x] valgrind
- [ ] add regression tests
- [ ] update documentation with new flag
- [ ] submit to mailing list


axxel:
Me being a humble git user who had high hopes this would make his live easier, may I ask: was this abandoned just because there was no-one who would be willing and able to look into and fix a segfault on the win+VS build? If so, that would be really sad.

PR #1371, opened 2022-10-26
[PATCH] [OUTREACHY] t7001-mv.sh :Use test_path_is_* functions in test script

fobiasic07:
Older test scripts use the command 'test -*' to verify the presence or absence of features such as files, directories and symbolic links. This however requires slightly complicated uses, such as 'test ! -f <path>'. The helper functions 'test_path_is_*' located in t/test-lib-functions.sh have taken into account all scenarios of the 'test -*' to reduce errors. This was a microproject to replace them with the helper functions.

Test script to verify the presence/absence of files, paths, directories,symboliclinks and many other features in mv command are using the command format

'test -(e|f|s|h|...).

Replace them with helper functions of format

'test_path_is_*'

Signed-off-by: fobiasic07 <debraobondo@gmail.com>

### DIFF
--- a/t/t7001-mv.sh
+++ b/t/t7001-mv.sh
@@ -60,8 +60,8 @@ test_expect_success 'checking the commit' '
 
 test_expect_success 'mv --dry-run does not move file' '
 	git mv -n path0/COPYING MOVED &&
-	test -f path0/COPYING &&
-	test ! -f MOVED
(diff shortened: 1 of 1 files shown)


dscho:
@fobiasic07 please note that you can force-push the branch to update the PR, no need to open a new one. Not a big deal in this instance, but something to remember in the future.

PR #1373, opened 2022-10-28
Fix autocorrect in work tree for bare repository

simu:
Currently, auto correction doesn't work reliably for commands which must run in a work tree (e.g. `git status`) in Git work trees which are created from a bare repository.

This patch adds a test case illustrating the issue and proposes a fix which adjusts the usage of `read_early_config()` in `help_unknown_cmd()` to match other usages of `read_early_config()`. In particular the patch removes the call to `git_default_config()` in the read config callback.

Changes since v1 (both suggested by Junio):
- Moved test to 9003
- Squashed change and test into a single commit

### DIFF
--- a/help.c
+++ b/help.c
@@ -563,7 +563,7 @@ static int git_unknown_cmd_config(const char *var, const char *value, void *cb)
 	if (skip_prefix(var, "alias.", &p))
 		add_cmdname(&aliases, p, strlen(p));
 
-	return git_default_config(var, value, cb);
+	return 0;
 }
 
 static int levenshtein_compare(const void *p1, const void *p2)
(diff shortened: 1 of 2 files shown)


simu:
I created the PR against `maint` since I based my changes on `maint` -- please let me know if I should rebase this onto `master`

apidwalinsubport:
> Currently, auto correction doesn't work reliably for commands which must run in a work tree (e.g. `git status`) in Git work trees which are created from a bare repository.
> 
> This patch adds a test case illustrating the issue and proposes a fix which adjusts the usage of `read_early_config()` in `help_unknown_cmd()` to match other usages of `read_early_config()`. In particular the patch removes the call to `git_default_config()` in the read config callback.
> 
> Changes since v1 (both suggested by Junio):
> 
> * Moved test to 9003
> * Squashed change and test into a single commit

PR #1393, opened 2022-12-11
git: fix word "does" sticking to "not" in message

Sheepolution:
When a repository is on a FAT32 file system it will send a message to the user explaining they should use `git config --global --add safe.directory`. To prevent a long line in the file the message is made up of two string literals. In this change the first literal has a space added to it to fix the issue that the last and the first word of the two literals stick together like the example image below.

image

### DIFF
--- a/compat/mingw.c
+++ b/compat/mingw.c
@@ -2752,7 +2752,7 @@ int is_path_owned_by_current_sid(const char *path, struct strbuf *report)
 			/*
 			 * On FAT32 volumes, ownership is not actually recorded.
 			 */
-			strbuf_addf(report, "'%s' is on a file system that does"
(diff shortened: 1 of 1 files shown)


dscho:
Could you please add a `Signed-off-by` line with your real name in it? The Git project insists on that. You can fix it by configuring `user.name`, then `git commit --amend --reset-author --signoff` (deleting the superseded `Signed-off-by` line), and force-pushing. See https://git-scm.com/docs/SubmittingPatches#sign-off for more details.

dscho:
@Sheepolution ⬆️

Sheepolution:
@dscho Done. Sorry for the delay.

dscho:
I guess it's `/submit` time!

PR #1432, opened 2023-01-09
git: replace two checks with one not equal check

AtariDreams:
(version < 2 || version > 2) looks silly
considering this is an integer.


noahbroyles (commented):
This makes sense

PR #1433, opened 2023-01-12
feat(git-prompt): customizable git-prompt symbols

aleeraser:


aleeraser:
Hi @dscho, I'm tagging you since I see you've been active on previous PRs.

I'd like to be allowed to contribute, and also ask for a little guidance with the CI failure.
It looks to me that the failure is with some `curl` options being deprecated, and unrelated from this PR.
I opened #1434 for those deprecation warnings. Does it make sense?

dscho:
> It looks to me that the failure is with some `curl` options being deprecated, and unrelated from this PR.
> I opened #1434 for those deprecation warnings. Does it make sense?

@aleeraser Yes!

WombleWoo7547:
I'm no expert in .sh files or any of the C, but I saw the alpine failed due to some expressions being deprecated, might want to look at it, might be an easy fix. Doesn't look like it though, I had a look and saw nothing "easy" to fix.

PR #1434, opened 2023-01-13
curl: fix deprecation warnings

aleeraser:
Various deprecation warnings have been issued by a [CI run](https://github.com/git/git/actions/runs/3906618785) for linux-musl (alpine), for some `curl` options.

Deprecated options are:
- CURLOPT_REDIR_PROTOCOLS
- CURLOPT_PROTOCOLS
- CURLOPT_IOCTLFUNCTION
- CURLOPT_IOCTLDATA
- CURLOPT_PUT

This PR aims to fix the warning.

### DIFF
--- a/http-push.c
+++ b/http-push.c
@@ -198,7 +198,7 @@ static void curl_setup_http(CURL *curl, const char *url,
 		const char *custom_req, struct buffer *buffer,
 		curl_write_callback write_fn)
 {
-	curl_easy_setopt(curl, CURLOPT_PUT, 1);
+	curl_easy_setopt(curl, CURLOPT_UPLOAD, 1);
 	curl_easy_setopt(curl, CURLOPT_URL, url);
 	curl_easy_setopt(curl, CURLOPT_INFILE, buffer);
 	curl_easy_setopt(curl, CURLOPT_INFILESIZE, buffer->buf.len);
--- a/http-push.c
+++ b/http-push.c
@@ -203,8 +203,8 @@ static void curl_setup_http(CURL *curl, const char *url,
 	curl_easy_setopt(curl, CURLOPT_INFILE, buffer);
 	curl_easy_setopt(curl, CURLOPT_INFILESIZE, buffer->buf.len);
 	curl_easy_setopt(curl, CURLOPT_READFUNCTION, fread_buffer);
-	curl_easy_setopt(curl, CURLOPT_IOCTLFUNCTION, ioctl_buffer);
-	curl_easy_setopt(curl, CURLOPT_IOCTLDATA, buffer);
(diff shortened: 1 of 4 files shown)


dscho:
Hmm. It [looks](https://github.com/git/git/actions/runs/3911454812/jobs/6684870561#step:4:354) as if the new symbols might not be available yet in Ubuntu's version of cURL...

dscho:
@aleeraser  https://curl.se/libcurl/c/CURLOPT_REDIR_PROTOCOLS_STR.html says that the symbol was introduced only in cURL v7.85.0, which is _very_ recent (we're only two versions later, on v7.87.0).

I fear that we'll need to play games with defining a symbol in `git-curl-compat.h` depending on the cURL version, and fall back to the current code for older versions.

In addition, I would like to suggest using a `strbuf` instead of a fixed-size `char` buffer. Something like this:

```c
struct strbuf allowed_protocols = STRBUF_INIT;

[...]
get_curl_allowed_protocols(0, &allowed_protocols);
curl_easy_setopt(result, CURLOPT_REDIR_PROTOCOLS_STR,
    allowed_protocols.buf);
strbuf_reset(&allowed_protocols);
get_curl_allowed_protocols(-1, &allowed_protocols);
curl_easy_setopt(result, CURLOPT_PROTOCOLS_STR,
    allowed_protocols.buf);
strbuf_reset(&allowed_protocols);
```

And in `get_curl_allowed_protocols()`, the `xsnprintf()` calls will need to be replaced by `strbuf_addf()` ones.

dscho:
On second thought, it might be easier to modify (and rename) `get_curl_allowed_protocols()` to also do the `curl_easy_setopt()` calls. That way, there could be just one `#ifdef [...] ... #else [...] #endif` block (which the Git project tries _very_ much to avoid, but I do not see any alternative here).

aleeraser:
I see there is an [ongoing discussion](https://public-inbox.org/git/xmqqv8l9n5fj.fsf@gitster.g/T/) about this topic on the mailing list. Right now the proposed solution is to avoid treating the deprecation warning as error.
While typically I would disagree with such a solution, the argument in support of that are pretty much valid:

> On the other hand, I don't know how useful those deprecations are going
to be, as it depends on the timeframe. If the deprecation is added for
the same version of libcurl that implements the alternative (which is
roughly the case here), then we'd always be stuck supporting the old and
new forms (old for backwards compatibility, and new to silence the
deprecation warning). We care a lot more about the deprecation once the
alternative has been around for a while, and/or the old way of doing
things is about to be removed. And if we just wait until that removal,
then we do not have to rely on deprecation warnings. The build will
break just fine on its own. :)

I'm closing this PR, the ongoing discussion is certainly more helpful.
Thanks for your help!

PR #1438, opened 2023-01-18
Replace deprecated GetVersion with RtlGetVersion

AtariDreams:
GetVersion has its behavior changed in
Windows 8 and above anyway,
so this is the right way to do it now.

The previous way returns the wrong value in Windows 8 and up, returning the manifest Windows data as opposed to the actual Windows data.


dscho:
@AtariDreams since reviewing patches comes at quite the cost, it would be advisable to:

- avoid sending patches too frequently (otherwise potential reviewers will assume that none of your patches are ready for review yet and start ignoring them)
- mark patches that aren't supposed to be finished (e.g. still need work on the commit message or even on the code changes) as RFC
- wait an appropriate amount of time to give reviewers a chance to respond (I see already the fourth version of this patch on the mailing list, with not so much as a single review in between)
- discuss the need for the patch in the cover letter (the current code seems to detect the Windows version just fine in Git for Windows, thanks to the included manifest, so why is this patch even necessary? Or is it not necessary, but merely intended to address a pet peeve?)

PR #1440, opened 2023-01-21
win32: prefer beginthreadex over CreateThread

AtariDreams:
Use pthread_exit instead of async_exit.

This means we do not have
to deal with Windows's implementation
requiring an unsigned exit coded
despite the POSIX exit code requiring
a signed exit code.

Use _beginthreadex instead of CreateThread
since we use the Windows CRT.

Finally, check for NULL handles, not "INVALID_HANDLE," as _beginthreadex guarantees a valid handle in most cases

cc: Jeff Hostetler <git@jeffhostetler.com>

### DIFF
--- a/compat/winansi.c
+++ b/compat/winansi.c
@@ -340,7 +340,7 @@ enum {
 	TEXT = 0, ESCAPE = 033, BRACKET = '['
 };
 
-static DWORD WINAPI console_thread(LPVOID unused)
+static unsigned int WINAPI console_thread(LPVOID unused)
 {
 	unsigned char buffer[BUFFER_SIZE];
 	DWORD bytes;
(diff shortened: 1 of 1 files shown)


j6t:
The review comments on earlier rounds of this patch suggest that it is not needed. If you think that these reviews miss a point, please explain it.

To express your view, you can use the issue tracker on Github if you do not like to write emails.

If you have nothing to add, please stop submitting the same patch again and again.

-- Hannes

AtariDreams:
> The review comments on earlier rounds of this patch suggest that it is not needed. If you think that these reviews miss a point, please explain it.
> 
> To express your view, you can use the issue tracker on Github if you do not like to write emails.
> 
> If you have nothing to add, please stop submitting the same patch again and again.
> 
> -- Hannes

This had to be rebased because of conflicts. Nothing to do with my feelings towards anyone involved. I resent the patch just in case as a result.

dscho:
What about the review comments that suggest that he patch is not needed in the first place? Did you plan on addressing them at some stage?

dscho:
So no reply at all?

LoboVerga:
Obtener Outlook para Android

PR #1446, opened 2023-02-01
git-gui: fix inability to quit after closing another instance

orgads:
If you open 2 git gui instances in the same directory, then close one of them and try to close the other, an error message pops up, saying: 'error renaming ".git/GITGUI_BCK": no such file or directory', and it is no longer possible to close the window ever.

Fix by catching this error, and proceeding even if the file no longer exists.


orgads:
Hi,

It's been more than a year since I opened a PR to git-gui upstream. I pinged several times, but looks like nobody's home.

Is it possible to accept it to git nevertheless? This is an annoying issue, and the fix is trivial.

orgads:
Forgot to include the link: https://github.com/prati0100/git-gui/pull/91

PR #1453, opened 2023-02-04
pathsec: make check for PATHSPEC_LITERAL more readable

AtariDreams:
This check is designed to die if global_magic
has the PATHSPEC_LITERAL and any other setting.

This can be written is a much more obvious way:
if global_magic has PATHSPEC_LITERAL, it can only
BE PATHSPEC_LITERAL, and if it isn't then there are other settings.

### DIFF
--- a/pathspec.c
+++ b/pathspec.c
@@ -281,8 +281,7 @@ static int get_global_magic(int element_magic)
 	if (get_icase_global())
 		global_magic |= PATHSPEC_ICASE;
 
-	if ((global_magic & PATHSPEC_LITERAL) &&
-	    (global_magic & ~PATHSPEC_LITERAL))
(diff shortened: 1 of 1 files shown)


dscho:
> @dscho is there a way to like not be CC? Can I remove my email from the CC I just did the submit?
> 
> Not sign the commit or author it?

It is the Git project's rule that they want to know how to contact you in case of follow-up questions. Details are found e.g. here: https://git-scm.com/docs/SubmittingPatches#sign-off

dscho:
If you are contributing the patch on someone else's behalf _without_ GitGitGadget, you are still Cc:ed. That's the rule. I did not make that rule. You may want to discuss this with someone who made that rule.

PR #1454, opened 2023-02-06
WIP: Make "git archive gzip" stream the same as gzip

cielavenir:
In https://github.com/git/git/commit/4f4be00d302bc52d0d9d5a3d4738bb525066c710 , `git archive` has been changed to use internal implementation. However, it caused different stream and a bunch of utilities were affected due to the checksum change.
This pull request will fill the gap and generate the same stream as gzip.

Todos:

- [] Port the change to zlib 1.2.13, I used 1.2.11 due to some implementation change in 1.2.12
- [] Robust testing, there might be failing cases

Note that need to move cielavenir/zlib submodule to git/zlib before merging.

----

I'm getting busy nowadays, so I save my current state as WIP pull request so that others can see it.

----

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,3 +2,6 @@
 	path = sha1collisiondetection
 	url = https://github.com/cr-marcstevens/sha1collisiondetection.git
 	branch = master
+[submodule "zlib"]
+	path = zlib
+	url = https://github.com/cielavenir/zlib.git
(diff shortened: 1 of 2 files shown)


cielavenir:
(thank you for /allow-ing, I will get back when I'm not busy)

(also I'm wondering what email address to use)

dscho:
Well, don't you know, you run into all kinds of interesting people all the time. Today, I researched `.zstd` support in 7-Zip and found https://github.com/cielavenir/p7zip... 😃

PR #1463, opened 2023-02-27
config.c: use struct for config reading state

chooglen:
Note to Junio: 8/8 (which renames "cs" -> "set") conflicts with
ab/config-multi-and-nonbool. I previously said that I'd rebase this, but
presumably a remerge-diff is more ergonomic + flexible (let me know if
I'm mistaken), so I'll send a remerge-diff in a reply (I don't trust GGG
not to mangle the patch :/).

After sending this out, I'll see if cocci can make it easy enough to
change config_fn_t. If so, I'll probably go with that approach for the
future 'libified' patches, which should also line up nicely with what
Ævar suggested.

= Changes in v3

- Adjust the *_push() function to be unconditional (like the *_pop()
  function)

- Various commit message and comment cleanups

= Changes in v2

- To reduce churn, don't rename "struct config_source cf" to "cs" early
  in the series. Instead, rename the global "cf" to "cf_global", and
  leave the existing "cf"s untouched.

  Introduce 8/8 to get rid of the confusing acronym "struct
  config_source cf", but I don't mind ejecting it if it's too much
  churn.

- Adjust 5/8 so to pass "struct config_reader" through args instead of
  "*data". v1 made the mistake of thinking "*data" was being passed to a
  callback, but it wasn't.

- Add a 7/8 to fix a bug in die_bad_number(). I included this because it
  overlaps a little bit with the refactor here, but I don't mind
  ejecting this either.

- Assorted BUG() message clarifications.

= Description

This series prepares for config.[ch] to be libified as as part of the
libification effort that Emily described in [1]. One of the first goals
is to read config from a file, but the trouble with how config.c is
written today is that all reading operations rely on global state, so
before turning that into a library, we'd want to make that state
non-global.

This series doesn't remove all of the global state, but it gets us
closer to that goal by extracting the global config reading state into
"struct config_reader" and plumbing it through the config reading
machinery. This makes it possible to reuse the config machinery without
global state, and to enforce some constraints on "struct config_reader",
which makes it more predictable and easier to remove in the long run.

This process is very similar to how we've plumbed "struct repository"
and other 'context objects' in the past, except:

* The global state (named "the_reader") for the `git` process lives in a
  config.c static variable, and not on "the_repository". See 3/6 for the
  rationale.

* I've stopped short of adding "struct config_reader" to config.h public
  functions, since that would affect non-config.c callers.

Additionally, I've included a bugfix for die_bad_number() that became
clear as I did this refactor.

= Leftover bits

We still need a global "the_reader" because config callbacks are reading
auxiliary information about the config (e.g. line number, file name) via
global functions (e.g. current_config_line(), current_config_name()).
This is either because the callback uses this info directly (like
builtin/config.c printing the filename and scope of the value) or for
error reporting (like git_parse_int() reporting the filename of the
value it failed to parse).

If we had a way to plumb the state from "struct config_reader" to the
config callback functions, we could initialize "struct config_reader" in
the config machinery whenever we read config (instead of asking the
caller to initialize "struct config_reader" themselves), and config
reading could become a thread-safe operation. There isn't an obvious way
to plumb this state to config callbacks without adding an additional arg
to config_fn_t and incurring a lot of churn, but if we start replacing
"config_fn_t" with the configset API (which we've independently wanted
for some time), this may become feasible.

And if we do this, "struct config_reader" itself will probably become
obsolete, because we'd be able to plumb only the relevant state for the
current operation, e.g. if we are parsing a config file, we'd pass only
the config file parsing state, instead of "struct config_reader", which
also contains config set iterating state. In such a scenario, we'd
probably want to pass "struct key_value_info" to the config callback,
since that's all the callback should be interested in anyway.
Interestingly, this was proposed by Junio back in [2], and we didn't do
this back then out of concern for the churn (just like in v1).

[1]
https://lore.kernel.org/git/CAJoAoZ=Cig_kLocxKGax31sU7Xe4==BGzC__Bg2_pr7krNq6MA@mail.gmail.com
[2]
https://lore.kernel.org/git/CAPc5daV6bdUKS-ExHmpT4Ppy2S832NXoyPw7aOLP7fG=WrBPgg@mail.gmail.com/

cc: Jonathan Tan <jonathantanmy@google.com>
cc: Emily Shaffer <nasamuffin@google.com>
cc: Calvin Wan <calvinwan@google.com>

### DIFF
--- a/config.c
+++ b/config.c
@@ -62,6 +62,9 @@ struct config_reader {
 static struct config_reader the_reader;
 
 /*
+ * FIXME The comments are temporarily out of date since "cf_global" has been
+ * moved to the_reader, but not current_*.
+ *
  * These variables record the "current" config source, which
  * can be accessed by parsing callbacks.
  *
@@ -76,11 +79,7 @@ static struct config_reader the_reader;
  * at the variables, it's either a bug for it to be called in the first place,
  * or it's a function which can be reused for non-config purposes, and should
  * fall back to some sane behavior).
- *
- * FIXME "cf_global" has been replaced by "the_reader.source", remove
- * "cf_global" once we plumb "the_reader" through all of the callback functions.
  */
-static struct config_source *cf_global;
 static struct key_value_info *current_config_kvi;
 
 /*
@@ -98,8 +97,6 @@ static inline void config_reader_push_source(struct config_reader *reader,
 {
 	top->prev = reader->source;
 	reader->source = top;
-	/* FIXME remove this when cf_global is removed. */
-	cf_global = reader->source;
 }
 
 static inline struct config_source *config_reader_pop_source(struct config_reader *reader)
@@ -109,8 +106,6 @@ static inline struct config_source *config_reader_pop_source(struct config_reade
 		BUG("tried to pop config source, but we weren't reading config");
 	ret = reader->source;
 	reader->source = reader->source->prev;
-	/* FIXME remove this when cf_global is removed. */
-	cf_global = reader->source;
 	return ret;
 }
 
@@ -175,6 +170,7 @@ struct config_include_data {
 	void *data;
 	const struct config_options *opts;
 	struct git_config_source *config_source;
+	struct config_reader *config_reader;
 
 	/*
 	 * All remote URLs discovered when reading all config files.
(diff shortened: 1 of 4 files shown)


Mado007:
This reduces the direct dependence on the "cf" static variable, which
will make it easier to remove in a later commit. The plumbed arg is
named "cs" to differentiate between it and the static variable.

The remaining references to "cf" will be tackled in later commits, since
they fall into the one following categories that need more work to
remove:

- Functions that assign to "cf"
- Public functions
- Callback functions

Mado007:
`git_config_check_init(repo);`

PR #1476, opened 2023-03-21
branch: improve error log on branch not found by checking remotes refs

ctmbl:
when failing to delete a branch with `git branch -d <branch>` because of branch not found, try to find a remote refs matching `<branch>` and if so add an hint:
`Did you forget --remote?` to the error message

cc: Clement Mabileau <mabileau.clement@gmail.com>

### DIFF
--- a/builtin/branch.c
+++ b/builtin/branch.c
@@ -216,10 +216,11 @@ static int delete_branches(int argc, const char **argv, int force, int kinds,
 	struct string_list refs_to_delete = STRING_LIST_INIT_DUP;
 	struct string_list_item *item;
 	int branch_name_pos;
+	const char *fmt_remotes = "refs/remotes/%s";
 
 	switch (kinds) {
 	case FILTER_REFS_REMOTES:
-		fmt = "refs/remotes/%s";
+		fmt = fmt_remotes;
 		/* For subsequent UI messages */
 		remote_branch = 1;
 		allowed_interpret = INTERPRET_BRANCH_REMOTE;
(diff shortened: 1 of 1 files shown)


builtin/branch.c
```diff
@@ -263,9 +265,25 @@ static int delete_branches(int argc, const char **argv, int force, int kinds,
...
-			      : _("branch '%s' not found."), bname.buf);
+			char* MISSING_REMOTE_REF_ERROR_MSG = "remote-tracking branch '%s' not found.";
+			char* MISSING_BRANCH_ERROR_MSG = "branch '%s' not found.";
+			char* MISSING_BRANCH_HINT_MSG = "branch '%s' not found.\n"
+											"Did you forget --remote?";
+
+			if (remote_branch) {
+				error(_(MISSING_REMOTE_REF_ERROR_MSG), bname.buf);
+			} else {
+				char* virtual_name = mkpathdup(FMT_REMOTES, bname.buf);
```
dscho, line 276: I would wager a bet that this `virtual_name` leaks and is the reason for the failures in `linux-leaks`, e.g. [here](https://github.com/git/git/actions/runs/4481323437/jobs/7877870324?pr=1476#step:4:1833).

@ctmbl could you try whether adding a `FREE_AND_NULL(virtual_name)` after the `resolve_refdup()` call fixes it?

ctmbl: Oh yes ofc I should have think about it! Tks for the review, I'll add it right away.

PR #1491, opened 2023-04-08
ko.po: fix typo in korean translation

K021:
I fixed a simple typo error in korean translation of `ko.po` file.


rimrul:
Since this is a l10n fix, you should open a PR with the korean l10n team at https://github.com/git-l10n-ko/git-l10n-ko/.

K021:
> Since this is a l10n fix, you should open a PR with the korean l10n team at https://github.com/git-l10n-ko/git-l10n-ko/.

Thank you for advice! But is that repository maintained? There is a PR opend in 2021, exactly same as this, but no one reviewed and not it is not merged. If I would like to fix this, opening PR in the repo(https://github.com/git-l10n-ko/git-l10n-ko/) is the only way to do that?

@rimrul

rimrul:
> Thank you for advice! But is that repository maintained? There is a PR opend in 2021, exactly same as this, but no one reviewed and it is not merged.

Right that repo is unmaintained. I had taken a quick glance at that, but somehow missread the `2.24.0` in the latest commit message as `2.40.0`.

> If I would like to fix this, opening PR in the repo(https://github.com/git-l10n-ko/git-l10n-ko/) is the only way to do that?

No, either a PR against 
https://github.com/git-l10n/git-po (the repo of the overall l10n) coordinator or sending this PR to the mailing list could work.

K021:
Thanks a lot! @rimrul

PR #1498, opened 2023-04-21
l10n: about box, use `©` directly instead of the escape

SixthDot:
the former is not substitued at all

cp-sym


SixthDot:
How is the real name determined. The commit is signed-off, matching address was set primary + public before sending the submit command... what have I missed ?

dscho:
> How is the real name determined ? The commit is signed-off, matching address was set primary + public before sending the submit command... what have I missed ?

[This](https://github.com/gitgitgadget/gitgitgadget/blob/08b8ef558ff4551b724b5be1ba03c5d47b806b56/lib/ci-helper.ts#L878-L888C25) is how GitGitGadget determines the full name. I believe the info is taken from your GitHub Profile, which you can [edit as described here](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-github-profile/customizing-your-profile/personalizing-your-profile#changing-your-profile-name).

PR #1501, opened 2023-05-04
pack-refs: allow users control over which refs to pack

john-cai:
git-pack-refs does not currently give much control over which refs are packed. By default, all tags and already packed refs are packed. --all allows the user to pack all refs. Beyond this the user does not have control. Introduce a pair of options --include and --exclude that will allow users full control over which refs end up in the packed-refs file.

Changes since v3:
- got rid of extending ref_exclusions
- use a strvec instead of a string_list to keep track of included refs

Changes since v2:
- repurpose ref_exclusions to be used for ref inclusions
- fixed test formatting
- adjusted --include to not include default of all tags

Changes since v1:
- Clarify that --all packs not just branch tips but all refs under refs/ in the documentation in patch 1
- Add --include in patch 3

It's worth noting that [1] discussed a proposal for a pack refs v2 format that would improve deletion speeds. The ref-table backend would also improve performance of deletions. However, both of those proposals are still being discussed.

1. https://lore.kernel.org/git/pull.1408.git.1667846164.gitgitgadget@gmail.com/

CC: Junio C Hamano <gitster@pobox.com>
CC: Christian Couder <christian.couder@gmail.com>

### DIFF
--- a/Documentation/git-pack-refs.txt
+++ b/Documentation/git-pack-refs.txt
@@ -8,7 +8,7 @@ git-pack-refs - Pack heads and tags for efficient repository access
 SYNOPSIS
 --------
 [verse]
-'git pack-refs' [--all] [--no-prune]
+'git pack-refs' [--all] [--no-prune] [--exclude <pattern>]
 
 DESCRIPTION
 -----------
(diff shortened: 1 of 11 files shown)


elivan2017:
ok thanks for your comments as you see the merger of an ETH AND ISU
TOGETHER WITH A BTC DO YOU THINK A NEW COIN COULD WORK?

PR #1514, opened 2023-05-20
tests: modernize test format part 2

john-cai:
There are a number of tests that still follow an old convention of using four-space indentation. Update a bunch of tests to use the new convention with tab indentation.

This is a continued effort from [1]

1. https://lore.kernel.org/git/pull.1513.git.git.1684440205.gitgitgadget@gmail.com/


precha1989 (commented):
#1521

PR #1519, opened 2023-06-03
cherry-pick: use trailer instead of free-text for `-x`

vermiculus:
When recording the origin commit during a cherry-pick, the current label used is not understood by git-interpret-trailers. Standardize onto the 'normal' trailer format that can be reasonably/reliably parsed and used by external tooling leveraging git-interpret-trailers.

This also somewhat improves the readability of resulting commit messages in some scenarios where trailers are already in use. Consider the example already present in cd650a4e (2023-02-12, "recognize '(cherry picked from ...' as part of s-o-b footer"):

   Signed-off-by: A U Thor <author@example.com>
   (cherry picked from commit da39a3ee5e6b4b0d3255bfef95601890afd80709)
   Signed-off-by: C O Mmitter <committer@example.com>

This would now show as

   Signed-off-by: A U Thor <author@example.com>
   Cherry-Picked-From-Commit: da39a3ee5e6b4b0d3255bfef95601890afd80709
   Signed-off-by: C O Mmitter <committer@example.com>

Most tests are adjusted for the new format. A test is added to demonstrate that the old free-text format in existing commit data is still considered part of the trailer block (i.e., the problem fixed by the above commit has not been re-introduced).

---

I considered (but did not pursue) a new configuration option for two reasons:

  1. Without regard to historical data, using a 'real' trailer seems inherently better than the current free-text state.

  2. Regarding historical data, adding a user-configurable option doesn't make things simpler for systems maintainers; those systems still have to handle both formats if they have such a need to begin with. As it's still a clear and readable format, end-user developers are unlikely to care to change it back.

The maintenance and cognitive costs of a new configuration option are not worth the minimal benefit it seems it would bring.


vermiculus:
GitGitGadget seems to be choking on the `--` in my message. Removing it.

cwsacnpodseven:
> Sincere apologies for the very quick v2; while I've been sitting on this patch for a while in one form or another, I neglected to update the documentation. This has now been addressed, as well as addressing another reference to the old format in trailer.c. (I've now evaluated results of regexp searches of `cherry.picked` and `picked.from`; there is nothing left that seems to be necessary to update.)

vaharikrishna is a jsfskdks jdksdkanslkdan

t/t3510-cherry-pick-sequence.sh
```diff
@@ -548,10 +548,10 @@ test_expect_success '--continue respects opts' '
...
 	git cat-file commit HEAD~2 >unrelatedpick_msg &&
 	git cat-file commit HEAD~3 >initial_msg &&
-	! grep "cherry picked from" initial_msg &&
-	grep "cherry picked from" unrelatedpick_msg &&
-	grep "cherry picked from" picked_msg &&
-	grep "cherry picked from" anotherpick_msg
+	! grep "Cherry-Picked-From-Commit" initial_msg &&
+	grep "Cherry-Picked-From-Commit" unrelatedpick_msg &&
+	grep "Cherry-Picked-From-Commit" picked_msg &&
+	grep "Cherry-Picked-From-Commit" anotherpick_msg
```
sunchao9, line 554: xxxxx

PR #1528, opened 2023-06-16
Add 'preserve' subcommand to 'git stash'

nadav96:
In this patch, we introduce a new subcommand preserve to git stash. The purpose of this subcommand is to save the current changes into the stash and then immediately re-apply those changes to the working directory.

Implementation-wise, this is achieved by adding a new branch to the conditional in the cmd_stash function, where we check if argv[0] is "preserve". If it is, we push_stash with the new argument that we added to it preserve=1. In all other cases we call push_stack/do_push_stack preserve=0

If the community will approve, I will modify the patch to include help messages for the new subcommand


Hefeweizen:
I'm not OP, just a drive-by commenter on this.  I chime in, because this functionality eerily echos an itch I've been scratching this week.  It's a motivation for this feature, though perhaps not OP's motivation for this feature.

feature: work-in-progess snapshot
I want git to capture the state of the working-tree in some temporary way.  The files aren't clean; I'm not ready for an "actual" commit.  Out of paranoia, I just want to know that the file contents are protected in version control so that if I bork my editing, I can go back.

For myself, I created this:
```
$ cat git/git-wip 
#!/bin/bash
# git wip - make a wip commit to capture intermediate changes


# if env var found, then abort saving wip
#
# This was specifically added to have git not interfere
# with --interactive patching/rebasing.
# ```
# $ git config core.editor
# NO_GIT_WIP=1 vim
# ```
if [[ ${NO_GIT_WIP} -eq 1 ]]; then
    exit
fi

current_branch_name=$(git rev-parse --abbrev-ref HEAD)

if git rev-parse --verify ${current_branch_name}-wip >/dev/null 2>&1; then
    git branch --delete --force ${current_branch_name}-wip >/dev/null
fi 
git checkout -b ${current_branch_name}-wip >/dev/null 2>&1
git commit --all -m "wip" >/dev/null
git checkout @{-1} >/dev/null 2>&1
git restore --worktree --source=${current_branch_name}-wip -- .
```
I then integrated this with vim, so that every file write is captured:
```
$ grep ls-files ~/.vimrc
autocmd BufWritePost * :!if git ls-files --error-unmatch % 1>/dev/null 2>&1; then git wip; fi
```


If a `git stash preserve` had existed, then my workflow could replace the last four lines with that one command (though, I'd still have to script some stash management, or it'd accrete endlessly).

Hefeweizen:
> I do agree that it would be reasonable to want to expose the first
> half (the other half is "now the local mod got saved in a stash,
> adjust the working tree and/or the index"), but then that means the
> code should cover the various operating modes we have, and let the
> users perform their first half, so that the second half (which by
> the way needs to be exposed by another series later) can be used on
> top of the result to emulate as if the combined two (i.e. "stash
> save/push") have been run, for the feature to be complete, no?

If I'm following along correctly, the assertion seems to be that:  functionality combined (A+B) exists; if we expose step A as a subcommand, then aren't we obliged to provide step B?

In this instance, I would say "no; it already exists."  The current functionality is a combination of: a) create stash commit; b) clean working directory.  B, as described is a straight-forward: `git reset --hard`.  A `git stash preserve --clean` seems redundant.

I do see an argument from a usage perspective that offering the subcomponent operations should be encompassing.  That is, if we introduce a `git stash preserve` as step A, then we need a `git stash preserve --clean` to expose step B.  But to rebut, I offer these points:
1) In the case of `git stash pop` and `git stash apply` there's no convenience alias of `git stash apply --clean` that aliases to `git stash drop`.   And, I think this is correct as the surface area of git commands is already large, do we add this alias to other functionality for the sake of symmetry?
2)  the man page, in its description of this new feature, will be differentiating itself from `git stash push`.  That contrast would be all the clearer if we provide the needed operation (i.e. `git reset --hard`).

dscho:
@Hefeweizen please reply on list. Almost nobody sees your replies here, which is frowned upon in the Git project.

PR #1551, opened 2023-07-17
config-parse: create config parsing library

chooglen:
Config parsing no longer uses global state as of gc/config-context, so the natural next step for libification is to turn that into its own library. This series does exactly that.


najmulsheik (commented):
99244816307b822bd8ffcbff8690ef449c797a23

lisa50loves:
Ok

rhoog-ine:
I guess this paves the way for alternate config parsing libraries to use with git that can:
- implement shell-style parameter expansion
- pre-processing with M4 or cpp
- use different config file syntax like json

PR #1557, opened 2023-07-30
Update sha1.h: Missing include stddef.h

lzsiga:
This header uses 'size_t' so including stddef.h seems reasonable.


lzsiga:
Also 'hasmap.h' uses both 'size_t' and 'ssize_t' so it should include 'sys/types.h'.

PR #1559, opened 2023-08-04
connect.c: disable ssh X11 forwarding by default

opohorel:
Prevent errors related to X11 forwarding by running 'ssh -x' command by default. Bugs in other components can sometimes cause git to hang when cloning via ssh with X11 forwarding enabled.

We have experienced this with dbus bug[0], when Git hanged while cloning from the server. This issue is easily fixed by setting an envar 'GIT_SSH_COMMAND' to 'ssh -x'.

As I don't think Git uses X11 forwarding when connecting to the server in any way, it makes sense to me to disable it by default in order to prevent any future hangs caused by other components.

[0]https://bugzilla.redhat.com/show_bug.cgi?id=1940067


connect.c
```diff
@@ -1376,7 +1376,7 @@ static void fill_ssh_args(struct child_process *conn, const char *ssh_host,
 
 		ssh = getenv("GIT_SSH");
 		if (!ssh)
-			ssh = "ssh";
+			ssh = "ssh -x";
 		variant = determine_ssh_variant(ssh, 0);
```
gitster, line 1380: The function builds ssh command line not as a single string but as an array similar to char **argv (notice the use of 0 as the second parameter to determine_ssh_variant; it tells the helper function that the string in the variable `ssh` is the name of the command, without any of its parameters.

And then later on, conn->args array is prepared by calling push_ssh_options() to fill the rest of it with various parameters, using the way that depends on the variant of the ssh client (e.g. the way to pass custom port may be different between openSSH and Putty).

I think disabling the X11 forwarding may be a good idea and the motivation behind the change is good, but the implementation is not quite there. A better way to do so may be to add another block at the end of connect.c:push_ssh_options() that looks more or less like

```
switch (variant) {
case VARIANT_FOO:
case VARIANT_BAR:
   strvec_push(args, "-x");
   break;
default:
  /* other variants are not capable of futzing with the X11 forarding */
  break;
}
```

without touching anything else.

gitster (changes requested):

connect.c
opohorel: Thanks for point this out. I've reimplemented it, as you suggested. 

I've looked through the documentation and I decided to disable X11 forwarding for every variant except of 'simple', as in the documentation there is a [note](https://github.com/git/git/blob/ac83bc5054c2ac489166072334b4147ce6d0fccb/Documentation/config/ssh.txt#L34) about not changing parameters for it. That being said, I'll add another commit changing the documentation later.

Another thing is that some tests fail now as they expect different parameters passed to ssh. I'll go through them and see if I didn't create any unexpected behavior

dscho:
@opohorel did you mean to `/submit` this PR again?

opohorel:
> @opohorel did you mean to `/submit` this PR again?

No, to be fair, I totally forgot about this PR and the whole gitgitgadget workflow with my other PR.
I'll try to take a look at this PR again, but I think last time I had some problems with rewriting tests.

PR #1562, opened 2023-08-09
sequencer: truncate lockfile and ref to NAME_MAX

mped-oticon:
Some commits may have unusually long subject lines, which can cause git error out.
Currently the sequencer and lockfile assumes these to be less than `NAME_MAX` which is the maximum length of a filename (on Linux).

When reproduced one is met by the error message:
```
$ git rebase --continue
error: cannot lock ref 'refs/rewritten/SANITIZED-SUBJECT': Unable to create '.git/refs/rewritten/SANITIZED-SUBJECT.lock': File name too long
```
- where `SANITIZED-SUBJECT` is very long

Affected repos can only be salvaged through filter-branch etc.


dscho:
@mped-oticon funnily enough, something similar to this came up in https://github.com/git-for-windows/git/issues/4544, so this is a greatly-timed PR!

One thing I am slightly concerned about is that this patch truncates at the byte boundary, which isn't the same as the character boundary in UTF-8. Can you think of an easy way to support UTF-8-encoded subject lines?

The easiest way I can think of is not really elegant: if the byte length is above the threshold and if `is_encoding_utf8()` indicates that the subject is encoded in UTF-8, then loop using [`utf8_width()`](https://github.com/git/git/blob/v2.41.0/utf8.c#L189-L202) to advance until the threshold is reached to avoid truncating in the middle of a multi-byte character.

By the way, it might make sense to decrement `GIT_MAX_LABEL_LENGTH` by a few, just to leave some room.

mped-oticon:
@dscho Cool! One method would be to simply take the hash of the raw subject/label. That would make things less readable ofc. 

To really do truncaction at UTF-8 level, it can be done by looping through the bytes and looking at their MSb. If set it is UTF-8, then basically count leading 1's gives the number of bytes. I'm sure there are tonnes of libraries around to do that.

refs/rewritten/ is pretty internal it seems, so perhaps the hash might be good enough?

dscho:
> One method would be to simply take the hash of the raw subject/label. That would make things less readable ofc.

@mped-oticon the readability issue would really hurt. I use those labels often to make sense of the rebase script. Remember, some people (like me) manage branch thickets with hundreds of commits and many dozens of merges. It would really harm my productivity if I could not make sense of those labels because they are hashes.

> To really do truncaction at UTF-8 level, it can be done by looping through the bytes and looking at their MSb.

That's essentially the `utf8_width()` method I described ;-)

Would you like me to help with that?

mped-oticon:
> Would you like me to help with that?

Sure!

mped-oticon:
I don't think `utf8_width()` can be used as that returns the width in rendered-columns - not the "width" in bytes.
But `pick_one_utf8_char()` can be useful.

mped-oticon:
@dscho This seems to work OK: 
https://gist.github.com/mped-oticon/8cdff3801365116c2360e6dff5d18f0a#file-adhoc_test-c-L169

dscho:
> I don't think `utf8_width()` can be used as that returns the width in rendered-columns - not the "width" in bytes.

Not directly, but then, but we get the pointer to the next character. The difference is the number of bytes. [This](https://github.com/git/git/compare/master...dscho:git:mped_bugfix_lockfile_maxname) is what I have currently (but I got interrupted multiple times, and tests are still missing):

```patch
commit cc6927e229417753b0e607c0c66c1b80462a6be9
Author: Johannes Schindelin <Johannes.Schindelin@gmx.de>
Date:   Thu Aug 10 11:16:15 2023 +0200

    amend! sequencer: truncate lockfile and ref to NAME_MAX
    
    sequencer: truncate labels to accommodate loose refs
    
    Some commits may have unusually long subject lines. When those subject
    lines are used as labels in the `--rebase-merges` mode of `git rebase`,
    they can cause errors when writing the corresponding loose refs because
    most file systems have a maximal file name length of 255 (`NAME_MAX`).
    The symptom looks like this:
    
            $ git rebase --continue
            error: cannot lock ref 'refs/rewritten/SANITIZED-SUBJECT': Unable to create '.git/refs/rewritten/SANITIZED-SUBJECT.lock': File name too long - where SANITIZED-SUBJECT is very long
    
    Let's accommodate this situation by truncating the labels.
    
    Care must be taken in case the subject line contains multi-byte
    characters so as not to truncate in the middle of a character.
    
    Signed-off-by: Mark Ruvald Pedersen <mped@demant.com>
    Signed-off-by: Johannes Schindelin <Johannes.Schindelin@gmx.de>

diff --git a/sequencer.c b/sequencer.c
index 863f4cb86cc..3fcc92f1cdc 100644
--- a/sequencer.c
+++ b/sequencer.c
@@ -51,9 +51,13 @@
 
 #define GIT_REFLOG_ACTION "GIT_REFLOG_ACTION"
 
-#define GIT_MAX_LABEL_LENGTH ((NAME_MAX) - (LOCK_SUFFIX_LEN))
-
-#define GIT_MIN(a, b) ((a) < (b) ? (a) : (b))
+/*
+ * To accommodate common filesystem limitations, where the loose refs' file
+ * names must not exceed `NAME_MAX`, the labels generated by `git rebase
+ * --rebase-merges` need to be truncated if the corresponding commit subjects
+ * are too long.
+ */
+#define GIT_MAX_LABEL_LENGTH ((NAME_MAX) - (LOCK_SUFFIX_LEN) - 16)
 
 static const char sign_off_header[] = "Signed-off-by: ";
 static const char cherry_picked_prefix[] = "(cherry picked from commit ";
@@ -3705,12 +3709,11 @@ static int do_label(struct repository *r, const char *name, int len)
 	struct strbuf msg = STRBUF_INIT;
 	int ret = 0;
 	struct object_id head_oid;
-	int len_trunc = GIT_MIN(len, GIT_MAX_LABEL_LENGTH);
 
 	if (len == 1 && *name == '#')
 		return error(_("illegal label name: '%.*s'"), len, name);
 
-	strbuf_addf(&ref_name, "refs/rewritten/%.*s", len_trunc, name);
+	strbuf_addf(&ref_name, "refs/rewritten/%.*s", len, name);
 	strbuf_addf(&msg, "rebase (label) '%.*s'", len, name);
 
 	transaction = ref_store_transaction_begin(refs, &err);
@@ -3774,12 +3777,11 @@ static const char *reflog_message(struct replay_opts *opts,
 static struct commit *lookup_label(struct repository *r, const char *label,
 				   int len, struct strbuf *buf)
 {
-	int len_trunc = GIT_MIN(len, GIT_MAX_LABEL_LENGTH);
 	struct commit *commit;
 	struct object_id oid;
 
 	strbuf_reset(buf);
-	strbuf_addf(buf, "refs/rewritten/%.*s", len_trunc, label);
+	strbuf_addf(buf, "refs/rewritten/%.*s", len, label);
 	if (!read_ref(buf->buf, &oid)) {
 		commit = lookup_commit_object(r, &oid);
 	} else {
@@ -5402,6 +5404,8 @@ static const char *label_oid(struct object_id *oid, const char *label,
 		}
 	} else {
 		struct strbuf *buf = &state->buf;
+		int label_is_utf8 = 1; /* start with this assumption */
+		size_t max_len = buf->len + GIT_MAX_LABEL_LENGTH;
 
 		/*
 		 * Sanitize labels by replacing non-alpha-numeric characters
@@ -5410,14 +5414,32 @@ static const char *label_oid(struct object_id *oid, const char *label,
 		 *
 		 * Note that we retain non-ASCII UTF-8 characters (identified
 		 * via the most significant bit). They should be all acceptable
-		 * in file names. We do not validate the UTF-8 here, that's not
-		 * the job of this function.
+		 * in file names.
+		 *
+		 * As we will use the labels as names of (loose) refs, it is
+		 * vital that the name not be longer than the maximum component
+		 * size of the file system (`NAME_MAX`). We are careful to
+		 * truncate the label accordingly, allowing for the `.lock`
+		 * suffix and for the label to be UTF-8 encoded (i.e. we avoid
+		 * truncating in the middle of a character).
 		 */
-		for (; *label; label++)
-			if ((*label & 0x80) || isalnum(*label))
+		for (; *label && buf->len + 1 < max_len; label++)
+			if (isalnum(*label) ||
+			    (!label_is_utf8 && (*label & 0x80)))
 				strbuf_addch(buf, *label);
+			else if (*label & 0x80) {
+				const char *p = label;
+
+				utf8_width(&p, NULL);
+				if (p) {
+					strbuf_add(buf, label, p - label);
+					label = p - 1;
+				} else {
+					label_is_utf8 = 0;
+					strbuf_addch(buf, *label);
+				}
 			/* avoid leading dash and double-dashes */
-			else if (buf->len && buf->buf[buf->len - 1] != '-')
+			} else if (buf->len && buf->buf[buf->len - 1] != '-')
 				strbuf_addch(buf, '-');
 		if (!buf->len) {
 			strbuf_addstr(buf, "rev-");
```

mped-oticon:
Looks good, `label_oid` seems to be the right place.
I'm not sure in the else-cond where `label_is_utf8` permanently set false, if that should be added verbatim: It is not an alnum, yet has MSb set but is not utf-8. Hmm.


FYI, https://github.com/git/git/commit/12defb06ab6e252db791ee9868477b927561241f

dscho:
> I'm not sure in the else-cond where `label_is_utf8` permanently set false, if that should be added verbatim: It is not an alnum, yet has MSb set but is not utf-8. Hmm.

Right, that's the fallback if we detect bytes with the high bit set which however do not conform to UTF-8. Then we fall back to assuming single-byte characters, like e.g. ISO-8859-15 does.

dscho:
> FYI, [12defb0](https://github.com/git/git/commit/12defb06ab6e252db791ee9868477b927561241f)

About this:

> ```c
> unsigned long utf8_bytes_truncate_line(const char *line, size_t max_bytes)
> ```

The 1970s called and want their data types back :-)

Seriously again, `unsigned long` is 32-bit on 64-bit Windows, and we already use `size_t` in this signature, so we should be consistent and use the latter throughout.

I was considering myself to add a function to `utf8.[ch]` instead, but decided against it because I want to _not_ require this to be UTF-8 encoded but fall back to assuming a non-multi-byte-character encoding, which seemed the wrong place for `utf8.h`.

dscho:
I found a bug! Fixed here: dscho/git@2b07c4d9df18a7337d3646b729ade6a6add79968

And I added a test case, and for good measure a config option to override the maximal label length: dscho/git@1ef7cbab2029867a32d8fc0cec86f9a397b72f2d

@mped-oticon could I ask you to run with this, polishing it more if you'd like, and then submitting it to the Git mailing list?

mped-oticon:
Sure. I'll also run it through our repo to verify.

mped-oticon:
Minor thing about missing default value. I'll try to fix it. A function in config.c perhaps.

dscho:
Oy, I introduced a bug that makes t3430 fail. I'm on it.

dscho:
> Minor thing about missing default value.

Ah, that's the bug ;-)

dscho:
I'd suggest this fix:

```diff
diff --git a/sequencer.c b/sequencer.c
index b2f8119a54d..d226530050f 100644
--- a/sequencer.c
+++ b/sequencer.c
@@ -5504,7 +5504,8 @@ static int make_script_with_merges(struct pretty_print_context *pp,
 	struct string_entry *entry;
 	struct oidset interesting = OIDSET_INIT, child_seen = OIDSET_INIT,
 		shown = OIDSET_INIT;
-	struct label_state state = { OIDMAP_INIT, { NULL }, STRBUF_INIT };
+	struct label_state state =
+		{ OIDMAP_INIT, { NULL }, STRBUF_INIT, GIT_MAX_LABEL_LENGTH };
 
 	int abbr = flags & TODO_LIST_ABBREVIATE_CMDS;
 	const char *cmd_pick = abbr ? "p" : "pick",
```

mped-oticon:
@dscho Thanks!   I assume I can send a command to the gitgitgadget bot to submit to the mailing list?

dscho:
> @dscho Thanks!   I assume I can send a command to the gitgitgadget bot to submit to the mailing list?

@mped-oticon yes: `/submit`

PR #1596, opened 2023-10-16
Chore: Typo fixed in multiple files

SusheelThapa:

### DIFF
--- a/Documentation/RelNotes/1.8.4.txt
+++ b/Documentation/RelNotes/1.8.4.txt
@@ -134,7 +134,7 @@ UI, Workflows & Features
    command, if "=whom" part is missing) and move the original author
    information to an in-body From: header as necessary.
 
- * The configuration variable "merge.ff" was cleary a tri-state to
+ * The configuration variable "merge.ff" was clearly a tri-state to
    choose one from "favor fast-forward when possible", "always create
    a merge even when the history could fast-forward" and "do not
    create any merge, only update when the history fast-forwards", but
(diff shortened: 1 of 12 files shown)


SusheelThapa:
@newren  Can you review my preview?

newren:
In the git project, the first line of commit messages are of the form: "area: short description".  "Chore" makes no sense there; I'd suggest "docs" for the area.

In the past, I've ignored typos in the RelNotes directory, because the release notes have already been published so there's not much value in changing them.  However, I do occasionally look at files in that directory, so I personally like seeing them fixed.  But there's a chance someone on the list complains that these are noise, and they do make up the majority of your patch.

The changes to git-daemon.txt are simply flipping American vs. British english.  We have many cases throughout the code of having used both kinds of spelling, so that one isn't really a fix. I have no idea whether that'll be accepted as is, or if you'd be asked to revert that piece.  Might bode better for your patch to drop that change, so we get one less complaint.

The fix to gc.txt is for a recent typo addition, and is the best part of the patch.  It may be that the gc.txt fix is good enough to just accept all the changes to RelNotes despite the normal comments about those, which would make me happy.  :-)

newren:
> > @SusheelThapa please do describe the context of your changes (how did you find them, how can we prevent similar issues in the future), fix your local Git user information (i.e. configure `user.name` and `user.email` correctly), and do [sign off on your patch](https://git-scm.com/docs/SubmittingPatches#dco) via `git commit --amend --reset-author -s`, then force-push to update the PR and fix the reported issues.
> 
> I have signed off my commit. Can you review it?

Please pay attention to gitgitgadget's review as well: Invalid author email in https://github.com/git/git/commit/5f8856fd494c6429322335545d4c8d7270bd569b: "[83917129+SusheelThapa@users.noreply.github.com](mailto:83917129+SusheelThapa@users.noreply.github.com)"

You need to fix that.  Also, my comments aren't a review, they're a preview.  You'll need to submit and have it sent off to the git mailing list before an official review can occur.

PR #1599, opened 2023-10-24
merge: --ff-one-only to apply FF if commit is one

QuAzI:
A new option to control the merging strategy. Plenty of developers want to simplify merge history. 

There are only two main merging strategies:
- Fast-forward (--ff) - There we lost merge commits for complex features and if we need to roll back some feature we can't revert just one commit.
- Merge (--no-ff) - There we have extra merges for extra simple changes

We prefer not to use squash. If we use squash we will lose some context and occasionally, we need multiple small features combined into one big feature. We would rather not mix it into one monolithic non-readable blob. For us, sometimes it is better to rebase something to make history more accurate than squash everything into one commit.

Currently, we have to juggle the parameters (--ff/--no-ff) every time to have history without extra merges for simple changes and to use merges for complex features

So, we prepared a small patch to add something like
- `--ff-one-only` option for `git merge`
- `git config merge.ff one-only`
- `git config branch.master.mergeoptions --ff-one-only`

A bit more context in [discussion](https://github.com/git-for-windows/git/discussions/4639)

### DIFF
--- a/Documentation/config/merge.txt
+++ b/Documentation/config/merge.txt
@@ -31,6 +31,9 @@ merge.ff::
 	a case (equivalent to giving the `--no-ff` option from the command
 	line). When set to `only`, only such fast-forward merges are
 	allowed (equivalent to giving the `--ff-only` option from the
+	command line). When set to `one-only`, fast-forward merge allowed
+	only for one commit, in other way extra merge commit should be
+	created (equivalent to giving the `--ff-one-only` option from the
 	command line).
 
 merge.verifySignatures::
--- a/builtin/merge.c
+++ b/builtin/merge.c
@@ -110,7 +110,8 @@ static const char *pull_twohead, *pull_octopus;
 enum ff_type {
 	FF_NO,
 	FF_ALLOW,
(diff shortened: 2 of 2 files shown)


PhilipOakley:
> Commit not signed off

I'm not seeing your sign-off as per the coding guidelines and submitting patches, but then I don't like GitHub's formatting of the PRs so it may jut be being 'hidden'

The commit message also needs to be proper plain text, with line breaks - I'm guessing you entered it as free-text that's interpreted as markdown and displayed (browser) as HTML, so doesn't have 'proper' line breaks.

Also the Git project does not like the `A bit more context in https://github.com/git-for-windows/git/discussions/4639` technique and prefers a proper summary of that discussion to be included in the commit message. There's no problem thereafter with referring to the discussion but the commit message should be 'complete'.

At some point you'll also need 'permission' to `/submit` the patches, once it passes all checks (do you have a 'test'?)

QuAzI:
@PhilipOakley 
I hope I fixed everything. But I don't have special tests yet.

PR #1610, opened 2023-11-16
Allow unset GIT_PS1_SHOWCONFLICTSTATE

wjandrea:
Previously, if `GIT_PS1_SHOWCONFLICTSTATE` was unset and `set -o nounset`/`set -u` was active, there would be an error.

I assume this is a pretty rare situation since it seems like `set -o nounset` isn't intended to be used interactively.

### DIFF
--- a/contrib/completion/git-prompt.sh
+++ b/contrib/completion/git-prompt.sh
@@ -511,7 +511,7 @@ __git_ps1 ()
 	fi
 
 	local conflict="" # state indicator for unresolved conflicts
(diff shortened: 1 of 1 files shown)


dscho:
> Phew, quite the process for a single-character change. Is it possible I could release the rights so that someone else can commit it?

Unfortunately, that's what the Git project wants, a [Developer's Certificate of Origin](https://git-scm.com/docs/SubmittingPatches#dco).

wjandrea:
Forget it then, I've already fixed the issue as far as I'm concerned by setting `GIT_PS1_SHOWCONFLICTSTATE` in my bashrc.

PR #1620, opened 2023-12-12
Use ^=1 to toggle between 0 and 1

AtariDreams:
If it is known that an int is either 1 or 0, doing an exclusive or to switch instead of a modulus makes more sense and is more efficient.

Signed-off-by: Seija Kijin doremylover123@gmail.com


builtin/fast-export.c
```diff
@@ -593,8 +593,8 @@ static void anonymize_ident_line(const char **beg, const char **end)
 	struct ident_split split;
 	const char *end_of_header;
 
-	out = &buffers[which_buffer++];
-	which_buffer %= ARRAY_SIZE(buffers);
+	out = &buffers[which_buffer];
+	which_buffer ^= 1;
```
axiac, line 597: This change does not take into account the fact that `which_buffer` is used as an index in `buffers`. If the size of `buffer` changes, this "hack" renders the change useless, producing confusion.

PR #1626, opened 2023-12-24
precious-files.txt: new document proposing new precious file type

newren:
A couple months ago, we had another in-depth discussion of precious files[1].  As with previous times, multiple strategies were discussed (including multiple new ones), meaning we keep making the possible solution space wider and never nail down an agreed path.  I also got the feeling we were potentially ignoring various edge cases or gotchas, and I thought it'd be good to have those better enumerated.

So, I went through the exercise of creating a design document to: (1) provide a specific design proposal and explore it, (2) cover at a high level the breadth of issues that an implementor needs to at least think about and which reviewers should be aware of in terms of readiness of a potential implementation, and (3) provide links to other discussions and alternative proposals for completeness.

I had some off-list discussions with Sebastian about this proposal, and he provided some helpful feedback.  The idea at this point is that if folks agree with the general direction, that he is going to be implementing at least the first cut basic capability.  I'll help review changes, but I'm mostly interested in just avoiding some of our past mistakes -- such as the large number of bugs we had in deleting untracked files unintentionally, not deleting ignored files when we should, and not documenting when untracked files were intended to be deleted by commands[2].

So...does the proposed direction seem reasonable to folks?

[1] https://lore.kernel.org/git/79901E6C-9839-4AB2-9360-9EBCA1AAE549@icloud.com/
[2] https://lore.kernel.org/git/pull.1036.v3.git.1632760428.gitgitgadget@gmail.com/

CC: Sebastian Thiel <sebastian.thiel@icloud.com>
CC: Josh Triplett <josh@joshtriplett.org>

### DIFF
--- a/Documentation/technical/precious-files.txt
+++ b/Documentation/technical/precious-files.txt
@@ -1,0 +1,540 @@
+Precious Files Design Document
+==============================
+
+Table of Contents
+  * Objective
+  * Background
+    * File categorization exceptions
+  * Proposal
+    * Precious file specification
+    * Breakdown of suggested behaviors by command
+  * Backward compatibility notes
+    * Slightly incompatible syntax
+    * Interaction with sparse-checkout parsing
+    * Behavior of traditional flags
+    * Interaction with older Git clients
+    * Commands with modified meaning
+  * Implementation hints
+    * Data structures
+    * Code areas
+    * Minimum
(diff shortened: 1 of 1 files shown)


newren:
I meant to open this in gitgitgadget/git, not git/git.

PR #1631, opened 2023-12-27
grep: default to posix digits with -P

carenas:
Since acabd2048e (grep: correctly identify utf-8 characters with \{b,w} in -P, 2023-01-08), PCRE2's UCP mode has been enabled when UTF content was expected and therefore muktibyte characters are included as part of all character classes (when defined).

note that if the locale used is not UTF enabled (specially: C, POSIX) or uses an extended charmap binary that is not unicode compatible, binary match will be used instead.

It was argued that doing so, at least for \d, was not a good idea, as that might not be what the user expected based on its historical meaning and was also slower, and indeed a similar change that was done to GNU grep was reverted and required further tweaks.

At that time, PCRE2 didn't have a way to disable UCP's character expansion selectively, but flags to do so, and that will be available in the next release (planned soon) were added, and one of them has been in use by GNU grep since their last release in May (only if built and linked against the prereleased PCRE2 library though).

Add flags to make sure that both \d and [:digit:] only include ASCII digits so that `git grep` will be closer to GNU grep and improve performance as a side effect, but add a configuration flag to allow keeping the current behaviour (which is closer to perl and ripgrep).


t/perf/p7822-grep-perl-character.sh
```diff
@@ -7,6 +7,8 @@ etc.) we will test the patterns under those numbers of threads.
 "
 
 . ./perf-lib.sh
+#. "$TEST_DIRECTORY"/lib-gettext.sh
+export LC_ALL=is_IS.utf8
```
carenas, line 11: this is not portable shell, but neither it is meant to be the final code.

ideally, there should be a way to define the right LOCALE (just like lib-gettext.sh) does by proving, and this test should be probably "skipped" if there is no unicode locale to use, as the result aren't "relevant" in that case (at least to compare it with the previous version).

truth is that this test is too noisy to be relevant anyway, so I'd been doing instead tests with `hyperfine` that show at least 20% improvement by restricting the sets to ASCII

majed08 (commented):
👍

majed08 (commented):
👍

PR #1644, opened 2024-01-14
clone: support cloning of filtered bundles

edigaryev:
f18b512bbb (bundle: create filtered bundles, 2022-03-09) introduced an incredibly useful ability to create filtered bundles, which advances the partial clone/promisor support in Git and allows for archiving large repositories to object storages like S3 in bundles that are:

* easy to manage
  * bundle is just a single file, it's easier to guarantee atomic replacements in object storages like S3 and they are faster to fetch than a bare repository since there's only a single GET request involved
* incredibly tiny
  * no indexes (which may be more than 10 MB for some repositories) and other fluff, compared to cloning a bare repository
  * bundle can be filtered to only contain the tips of refs neccessary for e.g. code-analysis purposes

However, in 86fdd94d72 (clone: fail gracefully when cloning filtered bundle, 2022-03-09) the cloning of such bundles was disabled, with a note that this behavior is not desired, and it the long-term this should be possible.

The commit above states that it's not possible to have this at the moment due to lack of remote and a repository-global config that specifies an object filter, yet it's unclear why a remote-specific config can't be used instead, which is what this change does.

CC: Derrick Stolee <stolee@gmail.com>

### DIFF
--- a/builtin/clone.c
+++ b/builtin/clone.c
@@ -1227,9 +1227,18 @@ int cmd_clone(int argc, const char **argv, const char *prefix)
 
 		if (fd > 0)
 			close(fd);
+
+		if (has_filter) {
+			strbuf_addf(&key, "remote.%s.promisor", remote_name);
+			git_config_set(key.buf, "true");
+			strbuf_reset(&key);
+
+			strbuf_addf(&key, "remote.%s.partialclonefilter", remote_name);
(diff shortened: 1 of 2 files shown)


Misschevious1 (commented):
Seriously cmon shane enough already ffs

PR #1646, opened 2024-01-17
Establish promises as a paradigm for asynchronous success/error tracking. Use `apply` code as a proving ground.

philip-peterson:
Similar to #1 but focuses on `apply` specifically. Also creates a "traceback" type output for failures.

```
% git apply -3 garbage.patch
error: 
	could not find header
caused by:
	no lines to read
error: No valid patches in input (allow with "--allow-empty")
```

To reproduce, use the following base (`not.txt`):

```
hello world
 
what is going on?
 
my name is

```

and the following patch (`garbage.patch`):

```
diff --git a/text.txt b/text.txt
index 741404d..95b791f 100644
--- a/text.txt
+++ b/text.txt
@@  @@
 
 what is going on?
 
-my name was
+my name is Joe
```

using the following invocation:

```
git apply -3 garbage.patch
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1107,6 +1107,7 @@ LIB_OBJS += preload-index.o
 LIB_OBJS += pretty.o
 LIB_OBJS += prio-queue.o
 LIB_OBJS += progress.o
+LIB_OBJS += promise.o
 LIB_OBJS += promisor-remote.o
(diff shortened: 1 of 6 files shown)


philip-peterson:
Apologies, erroneously opened this on the public repo when it was meant to be a private PR.

PR #1657, opened 2024-01-24
Remove the word "new" from during git switch -c new v2.6.13

BradleyYeo:
The new command does not work as it requires a space

### DIFF
--- a/Documentation/user-manual.txt
+++ b/Documentation/user-manual.txt
@@ -124,7 +124,7 @@ Create a new branch head pointing to one of these versions and check it
 out using linkgit:git-switch[1]:
 
 ------------------------------------------------
-$ git switch -c new v2.6.13
+$ git switch -c v2.6.13
 ------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


dscho:
> The new command does not work as it requires a space

I am fairly certain that this is an incorrect assessment. From https://git-scm.com/docs/git-switch#_synopsis:

image

Therefore, `git switch -c new v2.6.13` does what it is supposed to do: it creates a new branch called "new", starting at the commit that `v2.6.13` points to.

In contrast, `git switch -c v2.6.13` would _create a branch called `v2.6.13`_, and its initial tip commit would be the current commit. Definitely not what we'd want here.

dscho:
@BradleyYeo ping?

BradleyYeo:
Agree, you are right, my mistake Sent from my iPadOn 29 Jan 2024, at 3:51 AM, Johannes Schindelin ***@***.***> wrote:﻿
@BradleyYeo ping?

PR #1662, opened 2024-01-30
doc: Fix misspelled git-svn option: --minimize-url

vvs31415:


dscho:
Are you sure that the default isn't to minimize the URL and that the documented option works?

vvs31415:
A agree that the description is not very clear, and the name of the option does not make immediate sense to me.

vvs31415:
I guess the proposed change is wrong, therefore closing...

PR #1666, opened 2024-02-05
promise: introduce promises to track success or error

philip-peterson:
Hello all, this is my first patchset so thank you for being patient with me as I learn the process and conventions of your very fine project. These patches are intended as part of the libification effort, to show how we could use a Promise structure to help return values from functions.

## Problems

We seek to make libification easier by establishing a pattern for tracking whether a function errored in a rich way. Currently, any given function could immediately die(), or use error() to print directly to the console, bypassing any relevant verbosity checks. The use of die() currently makes use of Git as a library inconvenient since it is not graceful.

Additionally, returning using return error(...) (as is commonly done) always just returns a generic error value, -1, which provides little information.

## Approach

I solve this problem by splitting the single return value into two return values: error, and message. However, managing two output variables can require some coordination, and this coordination can be abstracted away by use of an existing pattern named Promise.

## Promise Concept

A promise is a contract representing "some task" that will eventually complete. Initially a promise is considered in a pending state. When it completes, one of two codepaths will eventually be entered: reject, or resolve. Once resolved or rejected, the promise enters a different state representing the result. Reject or resolve may only be called once on a given promise.

Until now, everything I described up to this point is consistent with other implementations, such as the ECMAScript standard for promises. However, this implementation departs from the complexity of those promises. In this implementation, promises are simple and canNOT be chained using .then(...) and do NOT have any notion of automatic bubbling (via re-entering the pending state).

## Sample output and reproduction

During an error, we can have richer feedback as to what caused the problem.

```
% git apply -3 garbage.patch
error: 
	could not find header
caused by:
	no lines to read
error: No valid patches in input (allow with "--allow-empty")
```

To reproduce this output, use the following base (`not.txt`):

```
hello world
 
what is going on?
 
my name is

```

and the following patch (`garbage.patch`):

```
diff --git a/text.txt b/text.txt
index 741404d..95b791f 100644
--- a/text.txt
+++ b/text.txt
@@  @@
 
 what is going on?
 
-my name was
+my name is Joe
```

using the following invocation:

```
git apply -3 garbage.patch
```

## Goals

I would love to get feedback on this approach. This patchset is kept small, so as to serve as a minimal proof of concept. It is intended to abstract to asynchronous use-cases even though this is only a synchronous one. Eventually, any top-level function, such as apply_all_patches(...) would return its output via a promise to make the library interface as clean as possible, but this patchset does not accomplish this goal. Hopefully it can provide a direction to go in to achieve that.

Thank you,
Philip

### DIFF
--- a/apply.c
+++ b/apply.c
@@ -27,6 +27,7 @@
 #include "object-file.h"
 #include "parse-options.h"
 #include "path.h"
+#include "promise.h"
 #include "quote.h"
 #include "read-cache.h"
 #include "rerere.h"
@@ -1317,13 +1318,14 @@ static int check_header_line(int linenr, struct patch *patch)
 	return 0;
 }
 
-int parse_git_diff_header(struct strbuf *root,
+void parse_git_diff_header(struct strbuf *root,
 			  int *linenr,
 			  int p_value,
 			  const char *line,
 			  int len,
 			  unsigned int size,
-			  struct patch *patch)
+			  struct patch *patch,
+			  struct promise_t* promise)
 {
 	unsigned long offset;
 	struct gitdiff_data parse_hdr_state;
@@ -1387,10 +1389,12 @@ int parse_git_diff_header(struct strbuf *root,
 			if (len < oplen || memcmp(p->str, line, oplen))
 				continue;
 			res = p->fn(&parse_hdr_state, line + oplen, patch);
-			if (res < 0)
-				return -1;
-			if (check_header_line(*linenr, patch))
-				return -1;
+			if (res < 0) {
+				PROMISE_THROW(promise, APPLY_ERR_GENERIC, "operation for \"%s\" failed with code: %d", p->str, res);
+			}
+			if (check_header_line(*linenr, patch)) {
+				PROMISE_THROW(promise, APPLY_ERR_GENERIC, "invalid header lines");
+			}
 			if (res > 0)
 				goto done;
 			break;
@@ -1400,25 +1404,25 @@ int parse_git_diff_header(struct strbuf *root,
 done:
 	if (!patch->old_name && !patch->new_name) {
 		if (!patch->def_name) {
-			error(Q_("git diff header lacks filename information when removing "
-				 "%d leading pathname component (line %d)",
-				 "git diff header lacks filename information when removing "
-				 "%d leading pathname components (line %d)",
-				 parse_hdr_state.p_value),
-			      parse_hdr_state.p_value, *linenr);
-			return -128;
+			PROMISE_THROW(promise, -128,
+				Q_("git diff header lacks filename information when removing "
+					"%d leading pathname component (line %d)",
+					"git diff header lacks filename information when removing "
+					"%d leading pathname components (line %d)",
+					parse_hdr_state.p_value),
+					parse_hdr_state.p_value, *linenr
+			);
 		}
 		patch->old_name = xstrdup(patch->def_name);
 		patch->new_name = xstrdup(patch->def_name);
 	}
 	if ((!patch->new_name && !patch->is_delete) ||
 	    (!patch->old_name && !patch->is_new)) {
-		error(_("git diff header lacks filename information "
+		PROMISE_THROW(promise, -128, _("git diff header lacks filename information "
 			"(line %d)"), *linenr);
-		return -128;
 	}
 	patch->is_toplevel_relative = 1;
-	return offset;
+	PROMISE_SUCCEED(promise, offset);
 }
 
 static int parse_num(const char *line, unsigned long *p)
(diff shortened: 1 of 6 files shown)


dscho:
@philip-peterson interesting idea!

> There are issues in commit [a44682b](https://github.com/git/git/commit/a44682be3d132a4fca324f11b7c9b88c74bea476):
> `apply: use new promise structures in git-apply logic as a proving ground`

Please review https://github.blog/2022-06-30-write-better-commits-build-better-projects/ for guidance how to improve the commit message so that the reviewers on the Git mailing list won't focus on a too-short commit message but instead focus on the much harder (but also much more important) task of reviewing the concept behind these patches.

> Commit not signed off

If you look at existing commits, you will see the `Signed-off-by:` footer. That's actually required, it is the [Developer’s Certificate of Origin](https://git-scm.com/docs/SubmittingPatches#dco).

apply.c
```diff
@@ -1591,15 +1595,24 @@ static int find_header(struct apply_state *state,
...
-				return -128;
+								size, patch, parse_git_diff_header_promise);
+			promise_assert_finished(parse_git_diff_header_promise);
+
+			if (parse_git_diff_header_promise->state == PROMISE_FAILURE) {
+				PROMISE_BUBBLE_UP(promise, parse_git_diff_header_promise,
+						 _("could not find file diff header"));
+			}
+
+			git_hdr_len = promise->result.success_result;
```
dscho, line 1610:
```suggestion
			git_hdr_len = parse_git_diff_header_promise->result.success_result;
```

Without this, you see tons of test failures.

This kind of finicky issue (where it is too easy to have perfectly compiling code even though the _wrong_ promise object was used by mistake) points to a problem with this approach that you may want to mention in the cover letter: this approach requires a lot of repetitions.

philip-peterson: Great point. Do you have any ideas by chance for how to reduce the repetition?

dscho: You could use the comma operator to combine the conditional return with the assignment of the return value but elegant it ain't...

philip-peterson:
@dscho Thanks for the comments. I addressed them mostly. Unfortunately now the `git add -p` tests are failing... any ideas as to why by chance ?

apply.c
```diff
@@ -4755,21 +4779,33 @@ static int apply_patch(struct apply_state *state,
...
+			error("\n\t%s", parse_chunk_promise->result.failure_result.message.buf);
+
+			nr = parse_chunk_promise->result.failure_result.status;
 			free_patch(patch);
 			if (nr == -128) {
 				res = -128;
 				goto end;
 			}
 			break;
 		}
```
dscho, line 4806: Please note that the original code prints _nothing_ if `nr < 0` unless it is `-128`. That's because it is not an error for a patch to have non-diff lines at the end; They are simply ignored. Therefore, you need to make sure to print the error only if we are actually erroring out, i.e.:

```diff
diff --git a/apply.c b/apply.c
index a5ad0355b5d..546ae4728bd 100644
--- a/apply.c
+++ b/apply.c
@@ -4794,11 +4794,10 @@ static int apply_patch(struct apply_state *state,
 
 		if (parse_chunk_promise->state == PROMISE_FAILURE) {
 			int nr;
-			error("\n\t%s", parse_chunk_promise->result.failure_result.message.buf);
-
 			nr = parse_chunk_promise->result.failure_result.status;
 			free_patch(patch);
 			if (nr == -128) {
+				error("\n\t%s", parse_chunk_promise->result.failure_result.message.buf);
 				res = -128;
 				goto end;
 			}
```

With that applied, t3701 passes again for me.

t/t4254-am-corrupt.sh
```diff
@@ -60,7 +60,13 @@ test_expect_success setup '
 test_expect_success 'try to apply corrupted patch' '
 	test_when_finished "git am --abort" &&
 	test_must_fail git -c advice.amWorkDir=false am bad-patch.diff 2>actual &&
-	echo "error: git diff header lacks filename information (line 4)" >expected &&
+	echo \
+"error:
```
dscho, line 64: As you saw, the `check-whitespace` job is really unhappy with that trailing space. There's code at the beginning of the file that provides inspiration how to deal with it:

```diff
diff --git a/t/t4254-am-corrupt.sh b/t/t4254-am-corrupt.sh
index 3de909a5f92..265bcc41bbd 100755
--- a/t/t4254-am-corrupt.sh
+++ b/t/t4254-am-corrupt.sh
@@ -60,8 +60,9 @@ test_expect_success setup '
 test_expect_success 'try to apply corrupted patch' '
 	test_when_finished "git am --abort" &&
 	test_must_fail git -c advice.amWorkDir=false am bad-patch.diff 2>actual &&
+	space=" " &&
 	echo \
-"error: 
+"error:$space
 	could not find header
 caused by:
 	could not find file diff header
```

philip-peterson:
@dscho  Thank you so much for figuring that out!! It will make things so much easier. I really appreciate it. 😊

PR #1682, opened 2024-02-23
userdiff: better method/property matching for C#

Whathecode:
Change since v1: I removed "from" from the list of keywords to skip.
First, I considered adding "await", but I discovered both "await" and
"from" are "contextual keywords", which unlike the other keywords
currently listed, aren't reserved, and can thus cause false negatives.
I.e., it is valid to have a method named "await" or "from".
In edge cases, this may lead to false positives, but a different
exclusion rule will need to be added to handle these.

Change since v2:
- Corrected comment formatting.
- Added `csharp-property-skip-body` test.
- Added comments in test code to explain sections not part of the test.
- Elaborated regex comments.
- Excluded math operators (+-*/%) in method pattern to not catch multiline
 operations, and tested for this in the `-skip-body` tests. Catching "-" only
 worked when it was defined at the end of the exclusion block for some
 reason. The regex matcher seems quite bugged.

Change since v3:
- Changed regex to better handle whitespace in types, making use of the
fact that it only appears after commas.
- Split regex into multiple lines with comments explaining structure.
- Split the "skip body" tests into more narrow `csharp-exclude-` tests.
- Added a test for generic methods: `csharp-exclude-generic-method-calls`.
- Added a test for array types used in methods: `csharp-method-array`.
- Added an addition property test: `csharp-property-braces-same-line`.
- Included a test for "( func(x)" case identified by Johannes in
`csharp-exclude-assignments`.

As before, I ran into many regex limitations (no possessive quantifiers,
no lookahead). It also seems different regex evaluators are used on
different test runs. Which one does `git diff` use? Maybe it is about time
to update this? E.g., if speed is a concern, possessive quantifiers can
speed up search.

Change since v4:
- Better matching of at least two "words".
- Better handling of generics by restricting commas within `< ... >`.
- Allow any spaces around commas in generics.
- Because of stricter use of comma, Johannes' identified failing cases
now pass.
- Updated tests to cover all of the above.

CC: Ævar Arnfjörð Bjarmason <avarab@gmail.com>, Jeff King <peff@peff.net>
cc: Linus Arver <linusa@google.com>

### DIFF
--- a/t/t4018/csharp-exclude-assignments
+++ b/t/t4018/csharp-exclude-assignments
@@ -1,0 +1,20 @@
+class Example
+{
+    string Method(int RIGHT)
+    {
+        var constantAssignment = "test";
+        var methodAssignment = MethodCall();
+        var multiLineMethodAssignment = MethodCall(
+        );
+        var multiLine = "first"
+            + MethodCall()
+            +
+            ( MethodCall()
+            )
+            + MethodCall();
+
+        return "ChangeMe";
+    }
+
+    string MethodCall(int a = 0, int b = 0) => "test";
+}
--- a/t/t4018/csharp-exclude-control-statements
+++ b/t/t4018/csharp-exclude-control-statements
@@ -1,0 +1,34 @@
+class Example
+{
+    string Method(int RIGHT)
+    {
+        if (false)
+        {
+            return "out";
+        }
+        else { }
+        if (true) MethodCall(
+        );
+        else MethodCall(
+        );
+        switch ("test")
+        {
+            case "one":
+            return MethodCall(
+            );
+            case "two":
+            break;
+        }
+        (int, int) tuple = (1, 4);
+        switch (tuple)
+        {
+            case (1, 4):
+              MethodCall();
+		      break;
+        }
+
+        return "ChangeMe";
+    }
+
+    string MethodCall(int a = 0, int b = 0) => "test";
+}
--- a/t/t4018/csharp-exclude-exceptions
+++ b/t/t4018/csharp-exclude-exceptions
@@ -1,0 +1,29 @@
+using System;
+
+class Example
+{
+    string Method(int RIGHT)
+    {
+        try
+        {
+            throw new Exception("fail");
+        }
+        catch (Exception)
+        {
+        }
+        finally
+        {
+        }
+        try { } catch (Exception) {}
+        try
+        {
+            throw GetException(
+            );
+        }
+        catch (Exception) { }
+
+        return "ChangeMe";
+    }
+
+    Exception GetException() => new Exception("fail");
+}
--- a/t/t4018/csharp-exclude-generic-method-calls
+++ b/t/t4018/csharp-exclude-generic-method-calls
@@ -1,0 +1,12 @@
+class Example
+{
+    string Method(int RIGHT)
+    {
+        GenericMethodCall<int, int>(
+            );
+
+        return "ChangeMe";
+    }
+
+    string GenericMethodCall<T, T2>() => "test";
+}
--- a/t/t4018/csharp-exclude-init-dispose
+++ b/t/t4018/csharp-exclude-init-dispose
@@ -1,0 +1,22 @@
+using System;
+
+class Example : IDisposable
+{
+    string Method(int RIGHT)
+    {
+        new Example();
+        new Example(
+            );
+        new Example { };
+        using (this)
+        {
+        }
+        var def =
+            this is default(
+                Example);
+
+        return "ChangeMe";
+    }
+
+    public void Dispose() {}
+}
--- a/t/t4018/csharp-exclude-iterations
+++ b/t/t4018/csharp-exclude-iterations
@@ -1,0 +1,26 @@
+using System.Linq;
+
+class Example
+{
+    string Method(int RIGHT)
+    {
+        do { } while (true);
+        do MethodCall(
+        ); while (true);
+        while (true);
+        while (true) {
+            break;
+        }
+        for (int i = 0; i < 10; ++i)
+        {
+        }
+        foreach (int i in Enumerable.Range(0, 10))
+        {
+        }
+        int[] numbers = [5, 4, 1, 3, 9, 8, 6, 7, 2, 0];
+
+        return "ChangeMe";
+    }
+
+    string MethodCall(int a = 0, int b = 0) => "test";
+}
--- a/t/t4018/csharp-exclude-method-calls
+++ b/t/t4018/csharp-exclude-method-calls
@@ -1,0 +1,20 @@
+class Example
+{
+    string Method(int RIGHT)
+    {
+        MethodCall();
+        MethodCall(1, 2);
+        MethodCall(
+            1, 2);
+        MethodCall(
+            1, 2,
+            3);
+        MethodCall(
+            1, MethodCall(),
+            2);
+
+        return "ChangeMe";
+    }
+
+    int MethodCall(int a = 0, int b = 0, int c = 0) => 42;
+}
--- a/t/t4018/csharp-exclude-other
+++ b/t/t4018/csharp-exclude-other
@@ -1,0 +1,18 @@
+class Example
+{
+    string Method(int RIGHT)
+    {
+        lock (this)
+        {
+        }
+        unsafe
+        {
+            byte[] bytes = [1, 2, 3];
+            fixed (byte* pointerToFirst = bytes)
+            {
+            }
+        }
+
+        return "ChangeMe";
+    }
+}
--- a/t/t4018/csharp-method
+++ b/t/t4018/csharp-method
@@ -1,0 +1,10 @@
+class Example
+{
+    string Method(int RIGHT)
+    {
+        // Filler
+        // Filler
+
+        return "ChangeMe";
+    }
+}
--- a/t/t4018/csharp-method-array
+++ b/t/t4018/csharp-method-array
@@ -1,0 +1,10 @@
+class Example
+{
+    string[] Method(int RIGHT)
+    {
+        // Filler
+        // Filler
+
+        return ["ChangeMe"];
+    }
+}
--- a/t/t4018/csharp-method-explicit
+++ b/t/t4018/csharp-method-explicit
@@ -1,0 +1,12 @@
+using System;
+
+class Example : IDisposable
+{
+    void IDisposable.Dispose() // RIGHT
+    {
+        // Filler
+        // Filler
+        
+        // ChangeMe
+    }
+}
--- a/t/t4018/csharp-method-generics
+++ b/t/t4018/csharp-method-generics
@@ -1,0 +1,11 @@
+class Example<T1, T2>
+{
+    Example<int, string> Method<TA, TB>(TA RIGHT, TB b)
+    {
+        // Filler
+        // Filler
+        
+        // ChangeMe
+        return null;
+    }
+}
--- a/t/t4018/csharp-method-generics-alternate-spaces
+++ b/t/t4018/csharp-method-generics-alternate-spaces
@@ -1,0 +1,11 @@
+class Example<T1, T2>
+{
+    Example<int,string> Method<TA ,TB>(TA RIGHT, TB b)
+    {
+        // Filler
+        // Filler
+        
+        // ChangeMe
+        return null;
+    }
+}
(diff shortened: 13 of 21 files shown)


Whathecode:
This is still a draft since I need guidance where/how to add tests.

I want to show which additional cases it now supports, and two cases I know of it doesn't support (and I'll gladly trade for overall better matching).

Specifically, constructors/destructors without access modifiers won't match. I.e., `Constructor()`, `~Destructor()`. You can view this as the current regex considering these normal method calls. I suppose an additional regex could be added which presumes these are defined on one line by matching the closing parenthesis, not followed by a semicolon. Just like the existing regex, it wouldn't work when the method definition straddles multiple lines.

Whathecode:
I tried to respond before through the mailing list, but never having done this before, and with limited or arcane documentation, it seems more trouble trying to respond to comments than create the original fix. 🙄

I don't have the tooling set up to respond, and neither are the current comments actionable. So, until I set time aside to figure out how to respond through the official channel, I'll just provide a quick response here.

> Roughly in other words, we assume that any file the end user throwsat us is a well formed program,
	
Yes. And within a well-formed program, the regex focuses on the distinguishing factors which sets a method definition apart from other lines of code, instead of aiming to have an exhaustive validation, which doesn't work either way since a method definition can (and commonly) straddles multiple lines, and which breaks (as the commit history shows) when new keywords are introduced.

> change method/property matching behavior, one thing at a time

Given that the nature/approach of the regex changed entirely, and was extremely broken to start with, this doesn't really make sense. I considered it while working on the code at first, but realized a different approach was needed instead.

Instead, as a reviewer I suggest to focus on the tests which pass instead, and weren't present before. You can revert the changes while keeping the tests and see how many fail.

dscho:
> I don't have the tooling set up to respond, and neither are the current comments actionable.

@listx FYI

listx:
@dscho Thanks for the ping

@Whathecode Generally speaking if a patch does more than 1 thing, it takes longer to review because the reviewer now has to be that much more of an expert in this area to give a proper review. If you are unable to break this patch down, I think the next best thing you could do is to explain why you didn't do so in your commit message or cover letter (the PR description will act as a cover letter which will be sent to the list but won't be part of the commit message(s) anywhere). Another thing you could do is CC the relevant people who've worked in this area most recently to signal-boost your patch:

```
 $ g shortlog -en --no-merges --since=24.months -- userdiff.c
Andrei Rybak <rybak.a.v@gmail.com> (3):
      userdiff: support Java type parameters
      userdiff: support Java record types
      userdiff: support Java sealed classes

Jeff King <peff@peff.net> (2):
      userdiff: mark unused parameter in internal callback
      userdiff: skip textconv caching when not in a repository

John Cai <johncai86@gmail.com> (2):
      diff: teach diff to read algorithm from diff driver
      attr: teach "--attr-source=<tree>" global option to "git"

René Scharfe <l.s.r@web.de> (2):
      userdiff: support regexec(3) with multi-byte support
      use xstrncmpz()

Calvin Wan <calvinwan@google.com> (1):
      git-compat-util: move alloc macros to git-compat-util.h

Elijah Newren <newren@gmail.com> (1):
      alloc.h: move ALLOC_GROW() functions from cache.h

Karthik Nayak <karthik.188@gmail.com> (1):
      attr: add flag `--source` to work with tree-ish
```

Whathecode:
@j6t 

Thank you for the review! I'll address what I can in code changes/cover letter, but the following I think may not belong in either of those, so I will comment on here (again, lacking experience in using mailing lists; apologies).

> So, this patch makes the choice to have some false positives instead of
> some false negatives. I have no experience in C#, so I trust your
> judgement that users are better served with this choice rather than the
> opposite.

I would argue false positives are preferable over false negatives in _any_ language.

False negatives can end up indicating a change happened in an entirely different context (e.g. different method) defined higher up in the file. Without looking at the actual code, you can't tell whether false negatives occurred. This is the exact reason I did this PR; the diff chunk headers shown on C# pull requests on GitHub are so unreliable, indicating properties/methods way up in source files, I simply stopped looking at them.

In contrast, false positives end up showing a line of code _within_ the current context as the diff chunk header. And, it is fairly easy to see when something is a false positive (e.g., it doesn't look like a method definition).

Specific to the keywords referred to here, note that the current version on `master` doesn't include `from` or `await` either. The cover letter documented a change I did since my first changeset. Originally, I added `from`, since I could think of it, but while going through [the full list of C# keywords](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/), I noticed it was listed under "contextual keywords", meaning, they _can_ be used as method names, which would invalidate the logic as commented right above that exclusion rule, which focuses on reserved keywords which can _never_ be method names.

Thus, to exclude `from` and `await` with certainty will require _different_ exclusion rules (maybe multiple), which could logically be added in a separate PR/commit. But to some extent, where these keywords may be used they are already excluded based on other indicators of non-method-things captured by the current regex. I will gladly add tests/additional rules for this later, but I don't think that should hold back this change set, as it doesn't invalidate the current changes which either way improve the status quo immensely.

> I am not 100% sure about the meaning of this expression. But would it
> not match this line, which looks like part of a multi-line expression:
>
>	+ func(x)

No. [`func(x)` doesn't match](https://regex101.com/r/Ob9Ppx/1), as it doesn't match "a minimum of two groups" as stated in the comment above. A method definition always needs the return type in front of the method name. But, type names can contain spaces (in the case of generics) where most of the complexity with this regex comes from, and where the previous regex failed miserably.

I will try to elaborate in the comments, but to a large degree these patterns are trial and error, and also fighting the weird limited subset of regex features available in this parser. The tests are really what should drive this type of code.

> >  	 /* Properties */
> > -	 "^[ \t]*(((static|public|internal|private|protected|new|virtual|sealed|override|unsafe)[ \t]+)*[][<>@.~_[:alnum:]]+[ \t]+[@._[:alnum:]]+)[ \t]*$\n"
> > +	 "^[ \t]*((([][[:alnum:]@_<>.,]+)[ \t]+[][[:alnum:]@_]*)+[^=:;,()]*)$\n"
> 
> OK. Since the second of the two words is optional and almost the same as
> the first, I have a gut feeling that the pattern can be expensive to
> match. But since I do not have data to back up my suspicion, let's leave
> it as it is.

I actually don't recall why I varied the second pattern. It seems to convey that the property name can't be generic (doesn't contain `<>` as per the first group), but because of the repeat pattern capturing both "words", that becomes mostly irrelevant. I will simplify the regex for now, as I did less testing on this either way, and no matching test cases to warrant the regex are available yet. The most important bit is really `[^=:;,()]` at the end.

j6t:
> > I am not 100% sure about the meaning of this expression. But would it
> > not match this line, which looks like part of a multi-line expression:
> > 
> > * func(x)
> 
> No. [`func(x)` doesn't match](https://regex101.com/r/Ob9Ppx/1), as it doesn't match "a minimum of two groups"

Sure, but `+ func(x)` does match. Do not mistake the `+` in my original message as a text formatting tool; I mean it literally.

Whathecode:
@j6t 

> Sure, but + func(x) does match. Do not mistake the + in my original message as a text formatting tool; I mean it literally.

Good catch! I'll try to address it tomorrow. In the meantime, I believe I addressed most of your other concerns.

j6t:
> * Excluded math operators (+-*/%) in method pattern to not catch multiline
>       operations, and tested for this in the `-skip-body` tests. Catching "-" only
>       worked when it was defined at the end of the exclusion block for some
>       reason. The regex matcher seems quite bugged.

Have you ever seen `[a-z]` in a regex? Now, what would then `[+-*]` mean? Of course, it's not a bugged matcher; the minus acts as itself only when it is first or last in the brackets.

When I wrote the example that looks like part of a multi-line expression, I meant it just as an example. By limiting the new regex to expression operators, you have missed my point. Can there not be other syntactical constructs (not expressions) that have the same problem?

When I look at my keyboard, I see many more symbols that can occur in expressions. Moreover

    ( func(x)

is not covered, either.

I have other suggestions to make. Please submit the patch, so I can comment the code inline.

Whathecode:
@j6t

> Have you ever seen [a-z] in a regex? Now, what would then [+-*] mean? Of course, it's not a bugged matcher; the minus acts as itself only when it is first or last in the brackets.

I escaped it, of course.

> When I wrote the example that looks like part of a multi-line expression, I meant it just as an example. By limiting the new regex to expression operators, you have missed my point. Can there not be other syntactical constructs (not expressions) that have the same problem?

I did get the point, but that arguably becomes more and more unlikely (also consider the exclusion rules at the end of the match). In the test I represented a common enough scenario to add it to exclusion. `{` was already excluded.

`(` I can try, but is likely to cause problems due to the fact this is a key characteristic of a method tested for. Probably if at the start of the line, I can exclude it, and I can make it more and more rigorous. But, there are diminishing returns here and tradeoffs to make. There will always be false positives due to the nature of the line by line matching, and missing regex features such a lookahead, etc.

Whathecode:
@j6t 

> Let me see if I make sense of this. The idea is to treat the ', '
> sequence as a single "character" so that the SP in this sequence does
> not count as the word separator that we otherwise have between two plain
> words.

Correct. This applies both the the definition and call sites. Generic types/methods ("templates" is maybe more familiar to you) can contain spaces between the `<` and `>`. And, in the case of call sites, this can be nested, since concrete types passed can be generic.

> I am unsure whether this is a workable solution. The set of symbols that
> can occur before a method definition overlaps too much with the symbols
> that can occur elsewhere.

False positives are definitely possible. But, consider my earlier argument that false positives are to be preferred over false negatives: a wrong method higher up in the file being marked, which currently happens _a lot_, giving a wrong contextual cue to the reader.

> `meth, func(x),`

Without running a test now (will look into it later), this would indeed match. The other line in the middle (`prop, y,`) wouldn't though, since it's not followed by an opening param.

I can try whether it's possible to detect whether the `,` lies within `< ... >`, but given the recursive nature, and the limitations of regex we can use (lookbehind would likely be useful here), I'm afraid we'll run into a wall. Maybe the test runners are no longer representative of regex capabilities out there in the wild and an upgrade is warranted?

However, if this fails, I don't see it as a deal-breaker. It's quite a common coding style to either place all parameters on one line (in which case the regex would work since it is preceded by `(`), or each separate parameter on a separate line (in which case the regex would also work since there aren't two groups). In addition, method calls often aren't inlined, but assigned to a variable first, often so that the method call fits on one line, and to have more self-documenting code.

It thus seems like a nice tradeoff for a false positive to have better matching everwhere else, to me.

> Or is the intent to find at least two words?

As documented: _"The strategy is to identify a minimum of two groups (any combination of keywords/type/name) before the opening parenthesis"_. So, I will give you alternate approach a try. Do note that it seems to be missing `[ \t]*` which is to allow for _"Optional space before parameters start"_ as documented. I.e., `void Method (int param)`.

> BTW that there is an | symbol in the potential space after the comma is
certainly an oversight, right?

Again, nice catch. Thank you for your very thorough review! 🙏 It must have been late and I was still thinking of "or" in the `( ... )` notation. I intended to write: `,[ \t]`. The tests didn't catch it because there were no generics without space. I'll add a case.

> Otherwise, in a non-matching case you send the matcher into an expensive
backtracking loop.

Backtracking is not needed in a lot of this regex. I tried using possessive quantifiers, but they don't seem supported, or at least not on all test runners.

dscho:
> I tried to respond before through the mailing list, but never having done this before, and with limited or arcane documentation, it seems more trouble trying to respond to comments than creating the original fix.

@Whathecode The problem is that your replies will not reach the intended recipients. I am sorry that GitGitGadget is not able to mirror your comments back to the Git mailing list. It's a feature I would have loved to implement, but due to a reorg at work I can no longer work on GitGitGadget on company time, nor on my main gig (Git for Windows), for that matter. So I won't be able to help here.

Whathecode:
@dscho Thank you for that perspective. And, thank you for Git for Windows! I use it daily, and it allows me to not have to resort to git CLI for quite a few of my common workflows (some additional rebase functionality would be useful, but I get by ;) ).

Know that your work is appreciated tremendously! Without it, I for one definitely wouldn't have bothered trying to contribute to this project. ❤️ I'm glad I was able to thanks to GitGitGadget.

At some point, I think it's also up to old open source communities to embrace new technologies. I don't quite see why you would want to conduct code reviews through mailing lists when you have GitHub PR reviews. 🤷 A good starting point would be accepting both.

PR #1683, opened 2024-03-02
Fix  git-p4 decode_path() missing an assignment

SaNeOr:
When using git-p4 in the python2 environment, when running to decode_path, 'bytes' is equal to 'str', decode( ) missing an assignment.


SaNeOr:
This CI error seems to have nothing to do with the commit?

marseko (commented):
git-p4.py

PR #1688, opened 2024-03-14
t5300: fix test_with_bad_commit()

john-cai:
0f8edf7317 (index-pack: --fsck-objects to take an optional argument for fsck msgs, 2024-02-01) added a test function test_with_bad_commit() that contained two bugs. test_expect_fail was used instead of test_must_fail, and a && was not included at the end of the line.

Fix these two issues in the test.

CC: szeder.dev@gmail.com


chaiyen123:
help

PR #1703, opened 2024-04-14
submodule: allow gitsubmodules to be pointed at tags @dark0dave

dark0dave:
I want to enable users freezing their submodules to a tag.

Example:

# Add
git submodule add changes

```sh
git submodule add --tag v34.3 https://github.com/Gibberlings3/SwordCoastStratagems.git
cat .gitmodules
[submodule "SwordCoastStratagems"]
  path = SwordCoastStratagems
  url = https://github.com/Gibberlings3/SwordCoastStratagems.git
  tag = v34.3
```

# Update

git submodule update changes

## Single tag

```sh
git submodule update  --remote -f --checkout
$
```
No changes as we are fixed to the tag

## No Single tag 

```sh
git submodule update --remote -f --checkout --no-single-tag
Submodule path 'SwordCoastStratagems': checked out '77426a2e14dbe584af811f922a6da3a48d74c074'
````

Checks out to HEAD of remote

CC: Revi Ewer <revi.ewer@example.com>, Ill Takalook <ill.takalook@example.net>
cc: Sean Allred <allred.sean@gmail.com>

### DIFF
--- a/builtin/submodule--helper.c
+++ b/builtin/submodule--helper.c
@@ -1541,9 +1541,11 @@ struct module_clone_data {
 	unsigned int dissociate: 1;
 	unsigned int require_init: 1;
 	int single_branch;
+	int single_tag;
 };
 #define MODULE_CLONE_DATA_INIT { \
 	.single_branch = -1, \
+	.single_tag = -1, \
 }
 
 struct submodule_alternate_setup {
@@ -1752,6 +1754,10 @@ static int clone_submodule(const struct module_clone_data *clone_data,
 					     clone_data->filter_options));
 		if (clone_data->single_branch >= 0)
 			strvec_push(&cp.args, clone_data->single_branch ?
+				    "--single-branch" :
+				    "--no-single-branch");
+		if (clone_data->single_tag >= 0)
+			strvec_push(&cp.args, clone_data->single_tag ?
 				    "--single-branch" :
 				    "--no-single-branch");
 
@@ -1981,6 +1987,7 @@ struct update_data {
 	int depth;
 	int max_jobs;
 	int single_branch;
+	int single_tag;
 	int recommend_shallow;
 	unsigned int require_init;
 	unsigned int force;
@@ -2004,6 +2011,7 @@ struct update_data {
 	.recommend_shallow = -1, \
 	.references = STRING_LIST_INIT_DUP, \
 	.single_branch = -1, \
+	.single_tag = -1, \
 	.max_jobs = 1, \
 }
 
@@ -2151,6 +2159,10 @@ static int prepare_to_clone_next_submodule(const struct cache_entry *ce,
 		strvec_push(&child->args, suc->update_data->single_branch ?
 					      "--single-branch" :
 					      "--no-single-branch");
+	if (suc->update_data->single_tag >= 0)
+		strvec_push(&child->args, suc->update_data->single_tag ?
+					      "--single-tag" :
+					      "--no-single-tag");
 
 cleanup:
 	free(displaypath);
@@ -2297,6 +2309,7 @@ static int fetch_in_submodule(const char *module_path, int depth, int quiet,
 	cp.dir = module_path;
 
 	strvec_push(&cp.args, "fetch");
+
 	if (quiet)
 		strvec_push(&cp.args, "--quiet");
 	if (depth)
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by dark0dave** [comment]
Thank you @dscho for enabling ci, this is now ready for review!

**Review by dscho — COMMENTED** [review]
Please note that the process required by the Git maintainer involves sending patches to the Git mailing list, see [here](https://github.com/git/git/pull/1703#issuecomment-2054145480) for more details.

As a "pre-review", I want to suggest avoiding the `feat(...)` notation: Reviewers on the Git mailing list do not follow such practices and are prone to be biased against contributors who use those.

I'd suggest to imitate 6f65f847664445b347810bece6eee84c2f39f4ef by using the `submodule:` prefix and then continuing in lower-case, in the imperative voice. Something like: "submodule: allow submodules to point to tags".

The commit message should not stopt there, though. You will note that commit message like the ones of 6f65f847664445b347810bece6eee84c2f39f4ef or 34b1a0d320e3a1531997d6681bacffbe9da7a046 go to great lengths to explain the motivation and add considerations that aren't easily deduced from reading the diff. You will want to follow the guidance in https://github.blog/2022-06-30-write-better-commits-build-better-projects/ to elaborate a lot more in the commit message, in particular with a strong focus on this part:

-- | -- | --

**Comment by dark0dave** [comment]
> Please note that the process required by the Git maintainer involves sending patches to the Git mailing list, see [here](https://github.com/git/git/pull/1703#issuecomment-2054145480) for more details.
> 
> As a "pre-review", I want to suggest avoiding the `feat(...)` notation: Reviewers on the Git mailing list do not follow such practices and are prone to be biased against contributors who use those.
> 
> I'd suggest to imitate [6f65f84](https://github.com/git/git/commit/6f65f847664445b347810bece6eee84c2f39f4ef) by using the `submodule:` prefix and then continuing in lower-case, in the imperative voice. Something like: "submodule: allow submodules to point to tags".
> 
> The commit message should not stopt there, though. You will note that commit message like the ones of [6f65f84](https://github.com/git/git/commit/6f65f847664445b347810bece6eee84c2f39f4ef) or [34b1a0d](https://github.com/git/git/commit/34b1a0d320e3a1531997d6681bacffbe9da7a046) go to great lengths to explain the motivation and add considerations that aren't easily deduced from reading the diff. You will want to follow the guidance in https://github.blog/2022-06-30-write-better-commits-build-better-projects/ to elaborate a lot more in the commit message, in particular with a strong focus on this part:
>   	What you’re doing 	Why you’re doing it
> High-level (strategic) 	**Intent** (what does this accomplish?) 	**Context** (why does the code do what it does now?)
> Low-level (tactical) 	**Implementation** (what did you do to accomplish your goal?) 	**Justification** (why is this change being made?)

Hi @dscho information added to commit. Please take annother look! Thank you.

**Comment by dscho** [comment]
> Lines in the body of the commit messages should be wrapped between 60 and 76 characters.
> Indented lines, and lines without whitespace, are exempt

Hey @dark0dave, much better! However, I think you may want to rework the commit message a little bit, not only to wrap the text at 76 columns per line, and not only avoiding Markdown (which the Git maintainer does not prefer to AsciiDoctor for some reason) by using two-space indent instead (in which case the lines are allowed to be longer), but also to convert it into a slightly more formal language that avoids the first person singular. Take a look around at Git commits that made it into published versions and you will likely understand what I'm trying to say here.

**Comment by dark0dave** [comment]
@dscho noted and thanks. Should be updated now.

**Comment by dscho** [comment]
> @dscho noted and thanks. Should be updated now.

Looks good to me!

**Comment by dark0dave** [comment]
@dscho should I submit now?

**Comment by dark0dave** [comment]
@dscho your input has been very useful thanks!

**Comment by paulsohn** [comment]
@dark0dave Allow me to necrobump.
I had a similar usecase to freeze/pin the submodule version, and I figured out that I can specify `update = none` inside `.gitmodules`, and then `git submodule update --remote` skips that submoudle.

```ini
[submodule "my_submodule_name"]
    path = path/to/my/submodule
    url = https://link.to.remote.repo
    update = none
```
It is somewhat inconvenient because you can't read the tag name directly, but you can always use `git submodule status` or `git describe`.

paulsohn:
At least I think, it would be handy to have an cli command to pin all submodules at once. One implementation idea would be setting `update = none` for all or selected submodules at once.

KrohnusMelavea:
This would solve a big problem we're having at our company at the minute. Necrobumping

johanbothad2d:
This is a huge need and submodules are subpar without it.

david0:
I also would like to see that feature.

When having merge conflicts with submodules its a mess to see which SHA is which version, so it would be way easier if a tag name is used instead of the commit OID.

> I'm assuming that the superproject tree itself still gets the subproject commit OID, so the commit is still stable when tags in the subproject.

I would really prefer if the human-readable tag name to be shown in diffs and merge conflicts. I do not care too much if the commit is stable or not (see below).

> But doesn't it seem odd -- particularly for the concept of tags -- that the behavior of updating that tree could change depending on when you do it?

It is a risk, but for me its acceptable and the lesser of two evils. In the rare case I retag, I need to be careful and consider that the new tag should not have the risk of any breaking change compared to the previous tag. 
Without tags, its just a PITA everyday, especially because if you get merge conflicts with submodules, its even hard to see which SHA is newer. 

> If you're ok with that behavior, then does it make sense to generalize it to the concept of a ref instead of just a tag name?
To me generalizing it to `ref` makes sense, but maybe its just a personal preference.

PR #1714, opened 2024-04-24
maintenance: continue on error

schiele:
When debugging a system where git maintenance was not handling all repositories registered for maintenance I found that there was one repository registered that failed maintenance operations since the upstream repository disappeared making the prefetch operation fail. Unfortunately this effectively also disabled all other repositories for maintenance following the problematic one in the list.

Obviously, I want other repositories still to be automatically maintained despite one particular one having a problem.

As such I implemented two parts:
1. I added a new option --continue to for-each-repo that makes the command continue despite errors.
2. I made maintenance use this option whenever calling for-each-repo

CC: Derrick Stolee <stolee@gmail.com>, Ævar Arnfjörð Bjarmason <avarab@gmail.com>, Lénaïc Huard <lenaic@lhuard.fr>

### DIFF
--- a/Documentation/git-for-each-repo.txt
+++ b/Documentation/git-for-each-repo.txt
@@ -9,7 +9,7 @@ git-for-each-repo - Run a Git command on a list of repositories
 SYNOPSIS
 --------
 [verse]
-'git for-each-repo' --config=<config> [--] <arguments>
+'git for-each-repo' --config=<config> [--continue] [--] <arguments>
 
 
 DESCRIPTION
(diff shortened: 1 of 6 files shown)


dscho:
Isn't this trying to address the same as gitgitgadget#1719, and hence competing with that [already-submitted patch series](https://lore.kernel.org/git/pull.1719.git.1713342535.gitgitgadget@gmail.com/T/#m20f3c5033fd2dd169e8ff13840755e40a0b137f4) @schiele?

schiele:
Indeed, this is essentially the same observation and it turns out we came to almost the same solution, mostly with different naming of the option. I was not aware of this submission. As such I will just close this PR since this is already handled in the other one.

dscho:
> Indeed, this is essentially the same observation and it turns out we came to almost the same solution, mostly with different naming of the option. I was not aware of this submission. As such I will just close this PR since this is already handled in the other one.

Excellent! You prodded me into submitting the final touch-ups for that patch series, I think it might even slip into v2.45.0, if not, then the next version.

PR #1723, opened 2024-05-20
dir: fix treat_leading_path() to return false on non-directories

ivantsepp:
Hi there!

I wanted to provide some extra context on the behavior I'm trying to fix. I noticed in one of my git hook scripts that it was showing a warning error message when running `git status` on a deleted directory. You can reproduce with the following on the git repo:

```
$ git s t/non_existent_directory/
warning: could not open directory 't/non_existent_directory/': No such file or directory
```

However, this warning doesn't show if you give a non-directory from the root of the repo:

```
$ git s non_existent_directory/
```

I found it strange that **sometimes** git is able to detect non-directories without warnings. Even stranger, an older version of git didn't show this warning in the first example. 

After running `git bisect`, I was able to track down the commit that introduced this behavior: b9670c1 ("dir: fix checks on common prefix directory", 2019-12-19).

Before, that change, `!is_directory(sb.buf)` would break out of the loop and return 0 since the branch of code that changes `rc` to 1 hasn't been reached yet. The change introduces the conditional `state == path_recurse` as the new return statement. This `state` variable could be from a previous iteration of the loop which is causing this new behavior. The fix is to ensure the directory conditional results in the method returning false.

I'm not sure if this is considered a bug or not since it's just a warning message. If this isn't noteworthy, then feel free to ignore this pull request!


ivantsepp:
> Before you submit, you may want to replace your `git s ...` commands to the full `git status ...` commands instead of using what I assume is a custom alias.

Thank you! Great catch! Updated the PR description to use the full command instead of my alias.

PR #1724, opened 2024-05-25
rebase: add `--update-refs=interactive`

intelfx:
In rebase-heavy workflows involving multiple interdependent feature branches, typing out `--update-refs` quickly becomes tiring, which can be mitigated with setting the `rebase.updateRefs` git-config option to perform update-refs by default.

However, the utility of `rebase.updateRefs` is somewhat limited because you rarely want it in a non-interactive rebase (as it does not give you the chance to review the update-refs candidates, likely leading to updating refs that you didn't want updated -- I made quite an amount of mess by setting this option and subsequently forgetting about it).

Try to find a middle ground by introducing a third value, `--update-refs=interactive` (and `rebase.updateRefs=interactive`) which means `--update-refs` when starting an interactive rebase and `--no-update-refs` otherwise. This option is primarily intended to be used in the gitconfig, but is also accepted on the command line for completeness.

### DIFF
--- a/Documentation/config/rebase.adoc
+++ b/Documentation/config/rebase.adoc
@@ -24,7 +24,12 @@ rebase.autoStash::
 	Defaults to false.
 
 rebase.updateRefs::
-	If set to true enable `--update-refs` option by default.
+	If set to true, enable the `--update-refs` option of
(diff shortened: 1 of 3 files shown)


LemmingAvalanche:
Why is this still a Draft PR? It looks ready?

PR #1726, opened 2024-05-27
fix: prevent date underflow when using positive timezone offset

dxrcy:
cc: darcy <acednes@gmail.com>
cc: Jeff King <peff@peff.net>
cc: <rsbecker@nexbridge.com>
cc: Karthik Nayak <karthik.188@gmail.com>

### DIFF
--- a/date.c
+++ b/date.c
@@ -868,6 +868,10 @@ static int match_object_header_date(const char *date, timestamp_t *timestamp, in
 	return 0;
 }
 
+
+/* timestamp of 2099-12-31T23:59:59Z, including 32 leap days */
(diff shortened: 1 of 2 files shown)


dxrcy:
Hey @dscho, sorry to bother you, but how should I best add the `Signed-off-by` trailer to the commit ? I don't want to force anything or rebase, if it will mess with the GitGadget process :-)

dxrcy:
> The best way is to make sure that your `user.name` and `user.email` are configured correctly, then call `git commit --amend -s` and force-push.

Thanks for the quick reply :-) should be good now.

dscho:
> > The best way is to make sure that your `user.name` and `user.email` are configured correctly, then call `git commit --amend -s` and force-push.
> 
> Thanks for the quick reply :-) should be good now.

Yup, that worked.

Lizzyrobin4:
Helo

PR #1731, opened 2024-06-12
attr: fix msan issue in read_attr_from_index

spectral54:


spectral54:
Wrong repo. Meant this for gitgitgadget, not git.

PR #1756, opened 2024-08-01
Small fixes for issues detected during internal CI runs

spectral54:
I'm attempting to get the git test suite running automatically during our weekly import. I have this mostly working, including with Address Sanitizer and Memory Sanitizer, but ran into a few issues:

- several tests were failing due to `strbuf_getcwd` not clearing `errno` on success after it internally looped due to the path being >128 bytes. This is resolved in depth; though either one of the commits alone would resolve our issues:
    - modify locations that call `strtoX` and check for `ERANGE` to set `errno = 0;` prior to calling the conversion function. This is the typical way that these functions are invoked, and may indicate that we want compatibility helpers in `git-compat-util.h` to ensure that this happens correctly (and add these functions to the banned list).
    - have `strbuf_getcwd` set `errno = 0;` prior to a successful exit. This isn't very common for most functions in the codebase, but some other examples of this were found.
- t6421-merge-partial-clone.sh had >10% flakiness. This is due to our build system using paths that contain a 64-hex-char hash, which had a 12.5% chance of containing the substring `d0`.

### DIFF
--- a/builtin/get-tar-commit-id.c
+++ b/builtin/get-tar-commit-id.c
@@ -35,6 +35,7 @@ int cmd_get_tar_commit_id(int argc, const char **argv UNUSED, const char *prefix
 	if (header->typeflag[0] != TYPEFLAG_GLOBAL_HEADER)
 		return 1;
 
+	errno = 0;
 	len = strtol(content, &end, 10);
 	if (errno == ERANGE || end == content || len < 0)
 		return 1;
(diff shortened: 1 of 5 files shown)


spectral54:
Removed second commit (set errno to 0 in strbuf_getcwd) per discussion. patch 1 and 3 untouched.

/submit

PR #1758, opened 2024-08-05
repository: prevent memory leak when releasing ref stores

csware:
CC: Patrick Steinhardt <ps@pks.im>


csware:
After another check this does not seem to be necessary. The ref stores are already free'd in `strmap_clear`.

PR #1766, opened 2024-08-12
git-svn: mention `svn:global-ignores` in help+docs

av-gal:


dscho:
@av-gal please avoid closing a PR only to open another, almost identical one (like [this one](https://github.com/git/git/pull/1765)). Please prefer to force-push and continue with the same PR.

PR #1761, opened 2024-08-07
explain the order of output in cat-file batch operations since it is not explicit

ahmedakef:
cc: Patrick Steinhardt <ps@pks.im>

PR #1768, opened 2024-08-21
docs: explain the order of output in The batched mode of git-cat-file(1)

ahmedakef:
this is the same change as https://github.com/git/git/pull/1761 but due to missteps, the PR got closed and I couldn't fix it,
also applied the review comments from @pks-t  
cc: Patrick Steinhardt [ps@pks.im](mailto:ps@pks.im)

### DIFF
--- a/Documentation/git-cat-file.txt
+++ b/Documentation/git-cat-file.txt
@@ -270,9 +270,9 @@ BATCH OUTPUT
 ------------
 
 If `--batch` or `--batch-check` is given, `cat-file` will read objects
-from stdin, one per line, and print information about them. By default,
-the whole line is considered as an object, as if it were fed to
-linkgit:git-rev-parse[1].
+from stdin, one per line, and print information about them in the same
+order as they have been read. By default, the whole line is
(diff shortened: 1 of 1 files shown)


dscho:
> /submit

That may have been a bit premature. The [initial comment](https://github.com/git/git/pull/1768#issue-2477472302) (which is used as "cover letter" between the commit message and the diff) neither contains the now-missing information about the previous iteration (including the range-diff and a detailed list of changes you made since sending v1, along with an explanation why this is marked as another v1 instead of a v2), plus: it still contains the boilerplate template starting with "Thanks for taking the time to contribute to Git!" that you most likely wanted to avoid sending in your contribution.

ahmedakef:
@dscho updated, thank you

dscho:
> /submit

Okay, hold your horses. You don't want to send more than _one_ iteration of the patch in the same day, preferably even in a couple of days. This is now the third iteration in quick succession, which would make for a fine way to upset reviewers, if that was your goal.

Instead, reply on-list to your patch if you have additional things to say, and do _slow down_ with the frequency of sending new iterations.

PR #1772, opened 2024-08-24
Clang tidy project

sondredl:


.clang-format
```diff
@@ -1,221 +1,64 @@
...
-ContinuationIndentWidth: 8
-ColumnLimit: 80
-
-# C Language specifics
+# 0 "../cppMakeProject/.clang-format"
+# 0 "<built-in>"
+# 0 "<command-line>"
+# 1 "/usr/include/stdc-predef.h" 1 3 4
+# 0 "<command-line>" 2
+# 1 "../cppMakeProject/.clang-format"
```
dscho, line 6: It's probably an idea in need of improvement to change the `.clang-format` file in a commit that claims to reformat the header files...

CMakeLists.txt
```diff
@@ -0,0 +1,1141 @@
+#
```
dscho, line 1: I suspect that there will be very few fans, if any, on the Git mailing list, of copying the CMake definition. Not only does this willfully risk diverging `CMakeLists.txt` files, it is also highly unclear why this is needed if the stated goal is to apply clang-tidy to the Git project.

Speaking of that goal: Patch series that touch so many files at once are likely to cause a lot of conflicts with other patch series that are in flight, and such conflicts rarely remain code conflicts and evolve into full-blown social conflicts. You may want to use a much more incremental approach, gaging what files in Git's code base are currently untouched by actively-developed patch series and handling only those in a first step.

dscho (commented):
This is a very, very large patch series that looks very, very different from other commits in Git's commit history. As such, I expect large changes to be required before it finds the Git reviewers' favor.

One thing to mention in particular is Git's focus on clear, informative commit messages. Have a look around in the revision history of Git's default branch, and you probably won't find a single one-liner commit message for a very, very long time. You will most likely be interested in following the guidance in https://github.blog/2022-06-30-write-better-commits-build-better-projects/ to improve the commit messages, paying close attention in particular to this part:

-- | -- | --

dscho:
@sondredl since you have submitted these patches manually, can this here PR be closed?

PR #1788, opened 2024-09-23
Remove the_repository global for am, annotate, apply, archive builtins

john-cai:
Remove the_repository global variable for the annotate, apply, and archive bulitins.

Changes since V3:

- Improve commit message in patch 2

Changes since V1:

- in patch 1, only pass in repo to the bulitin if the repo exists

Changes since V2:

- drop patch 3, which is a bit more involved to dis-entangle the_repository
- use a single variable in run_builtin() to keep track of whether or not we are operating in a repository

cc: shejialuo <shejialuo@gmail.com>

### DIFF
--- a/builtin/add.c
+++ b/builtin/add.c
@@ -385,7 +385,8 @@ int cmd_add(int argc,
 	char *ps_matched = NULL;
 	struct lock_file lock_file = LOCK_INIT;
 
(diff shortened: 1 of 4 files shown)


QQn00bs:
> Preview email sent as pull.1788.git.git.1727123126.gitgitgadget@gmail.com

PR #1803, opened 2024-10-06
[Outreachy][Patch v1] t3404: avoid losing exit status to pipes

Unique-Usman:
The exit code of the preceding command in a pipe is disregarded. So if that preceding command is a Git command that fails, the test would not fail. Instead, by saving the output of that Git command to a file, and removing the pipe, we make sure the test will fail if that Git command fails.


Unique-Usman:
cc: Patrick Steinhardt <ps@pks.im>, Christian Couder <christian.couder@gmail.com>

PR #1819, opened 2024-10-22
compat/mingw: implement POSIX-style atomic renames over open files

pks-t:
This PR implements POSIX-style atomic renames when the target path has open file handles.

### DIFF
--- a/compat/mingw.c
+++ b/compat/mingw.c
@@ -2217,10 +2217,16 @@ int mingw_accept(int sockfd1, struct sockaddr *sa, socklen_t *sz)
 #undef rename
 int mingw_rename(const char *pold, const char *pnew)
 {
+	static int supports_file_rename_info_ex = 1;
 	DWORD attrs, gle;
 	int tries = 0;
 	wchar_t wpold[MAX_PATH], wpnew[MAX_PATH];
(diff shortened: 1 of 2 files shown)


pks-t:
FYI @dscho

pks-t:
This works now. I had to drop `GENERIC_WRITE` permissions when opening the file-to-be-renamed as it could be a read-only file. We only require `DELETE` permissions anyway.

compat/mingw.c
```diff
@@ -2165,11 +2226,86 @@ int mingw_rename(const char *pold, const char *pnew)
...
+			return 0;
+
+		/*
+		 * When we see ERROR_BAD_ARGUMENTS we can assume that the
+		 * current system doesn't support FileRenameInfoEx. Keep us
+		 * from using it in future calls and retry.
+		 *
+		 * TODO: Test on systems that do not have FileRenameInfoEx.
+		 */
+		if (gle == ERROR_BAD_ARGUMENTS) {
```
dscho, line 2290: @pks-t I just tested this on Windows 8.1, and indeed, it needs to be `ERROR_INVALID_PARAMETER` instead of `ERROR_BAD_ARGUMENTS`, then it works.

pks-t: @dscho Thanks, adapted :+1:

pks-t:
Closing, this has been merged.

PR #1826, opened 2024-11-04
Put is_bare_repository_cfg global into repository struct

john-cai:
This patch series removes the global state introduced by the is_bare_repository_cfg variable by moving it into the repository struct. Patch 2 adds a safety measure by BUG()ing when the variable has not been properly initialized.


dscho:
Aargh. This is another instance of https://github.com/gitgitgadget/gitgitgadget/issues/1747. I'll investigate immediately.

a0983627353:
發表評論

PR #1828, opened 2024-11-11
diff: update conflict handling for whitespace to issue a warning

Unique-Usman:
Changes from V1
- Disable both WS_TAB_IN_INDENT and WS_INDENT_WITH_NON_TAB when both are set.


a0983627353:
[]()

PR #1832, opened 2024-11-24
fast-import: disallow "." and ".." path components

newren:


newren:
Why are my pull requests opening in git/git when I requested gitgitgadget/git?  Let's try again...

PR #1843, opened 2024-12-15
git: use signed loop counters when comparing with signed vars

AZero13:
The compiler (at least LLVM) treats them better anyway because of the C standard not defining signed overflow.

### DIFF
--- a/add-interactive.c
+++ b/add-interactive.c
@@ -848,7 +848,7 @@ static int get_untracked_files(struct repository *r,
 			       const struct pathspec *ps)
 {
 	struct dir_struct dir = { 0 };
-	size_t i;
+	int i;
 	struct strbuf buf = STRBUF_INIT;
 
 	if (repo_read_index(r) < 0)
--- a/bloom.c
+++ b/bloom.c
@@ -314,7 +314,7 @@ static int has_entries_with_high_bit(struct repository *r, struct tree *t)
 
 		init_tree_desc(&desc, &t->object.oid, t->buffer, t->size);
 		while (tree_entry(&desc, &entry)) {
-			size_t i;
+			int i;
 			for (i = 0; i < entry.pathlen; i++) {
 				if (entry.path[i] & 0x80) {
 					t->object.flags |= HIGH_BITS;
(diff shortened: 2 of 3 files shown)


add-interactive.c
```diff
@@ -848,7 +848,7 @@ static int get_untracked_files(struct repository *r,
 			       const struct pathspec *ps)
 {
 	struct dir_struct dir = { 0 };
-	size_t i;
+	int i;
```
dscho, line 851: How about `ssize_t` instead? `int` is not guaranteed to have the same amount of bits as `size_t`.

At the same time you may want to look more closely at the comparisons and consider overflow-proof variants similar to https://github.com/git/git/blob/v2.47.1/git-compat-util.h#L133

dscho:
@AreaZR I also edited out the welcome message from [the PR description](https://github.com/git/git/pull/1843#issue-2740259205): this description is sent as cover letter, and leaving in this message would not be well-received on the Git mailing list.

dscho:
> User AreaZR is now allowed to use GitGitGadget.
> 
> WARNING: AreaZR has no public email address set on GitHub; GitGitGadget needs an email address to Cc: you on your contribution, so that you receive any feedback on the Git mailing list. Go to https://github.com/settings/profile to make your preferred email public to let GitGitGadget know which email address to use.

@AZero13 just _how_ often do you change your GitHub handle?!?

ChrisIdema:
I understand using `int` for the loop `variable` that compares with `argc`. Since `argc` is always of type `int`.

But I disagree with using `int` for loop variable when comparing with `pathlen`.
Struct member `pathlen` of struct `name_entry` is of type `int`. As far as I know `pathlen` doesn't go negative (I used github search to search). So for me it would make more sense to change `pathlen` to `size_t`. That way `i` can remain of type `size_t`. But I don't know if this causes issues elsewhere. In at least 2 places `pathlen` is passed to a function argument of type `size_t`: `strbuf_add` and `strbuf_make_traverse_path`. So to me it seems `pathlen` should be `size_t`.

PR #1851, opened 2024-12-19

Unique-Usman:
This is the fourth iteration of adding os-version capability to Git v2. You can check the initial iterations on my Git fork https://github.com/Unique-Usman/git/ by going to different branches.

### DIFF
--- a/t/t5555-http-smart-common.sh
+++ b/t/t5555-http-smart-common.sh
@@ -123,9 +123,19 @@ test_expect_success 'git receive-pack --advertise-refs: v1' '
 '
 
 test_expect_success 'git upload-pack --advertise-refs: v2' '
+	printf "agent=FAKE" >agent_os_name &&
+	if test_have_prereq WINDOWS
+	then
+		test_config transfer.advertiseOSVersion false
+	else
+		# Octal intervals \001-\040 and \177-\377
+		# corresponds to decimal intervals 1-32 and 127-255
+		printf "\nos-version=%s\n" $(uname -s | tr -d "\n" | tr "[\001-\040][\177-\377]" ".") >>agent_os_name
+	fi &&
+
 	cat >expect <<-EOF &&
 	version 2
-	agent=FAKE
+	$(cat agent_os_name)
 	ls-refs=unborn
 	fetch=shallow wait-for-done
 	server-option
--- a/t/t5701-git-serve.sh
+++ b/t/t5701-git-serve.sh
@@ -8,13 +8,23 @@ export GIT_TEST_DEFAULT_INITIAL_BRANCH_NAME
 . ./test-lib.sh
 
 test_expect_success 'test capability advertisement' '
+	printf "agent=git/$(git version | cut -d" " -f3)" >agent_os_name &&
+	if test_have_prereq WINDOWS
+	then
+		test_config transfer.advertiseOSVersion false
+	else
+		# Octal intervals \001-\040 and \177-\377
+		# corresponds to decimal intervals 1-32 and 127-255
+		printf "\nos-version=%s\n" $(uname -s | tr -d "\n" | tr "[\001-\040][\177-\377]" ".") >>agent_os_name
+	fi &&
+
 	test_oid_cache <<-EOF &&
 	wrong_algo sha1:sha256
 	wrong_algo sha256:sha1
 	EOF
 	cat >expect.base <<-EOF &&
 	version 2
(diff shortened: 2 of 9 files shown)


version.c
```diff
@@ -7,7 +7,11 @@
 const char git_version_string[] = GIT_VERSION;
 const char git_built_from_commit_string[] = GIT_BUILT_FROM_COMMIT;
 
-static void strbuf_sanitize(struct strbuf *buf)
+/*
+ * Trim and replace each character with ascii code below 32 or above
+ * 127 (included) using a dot '.' character.
```
chriscool, line 12: It might be good idea to add something like:

TODO: ensure consecutive non-printable characters are only replaced once

chriscool: I think the TODO should be at the end of the comment, not at the beginning.

version.c
```diff
@@ -5,6 +5,20 @@
 const char git_version_string[] = GIT_VERSION;
 const char git_built_from_commit_string[] = GIT_BUILT_FROM_COMMIT;
 
+/*
+ * TODO: ensure consecutive non-printable characters are only replaced once;
+ * Trim and replace each character with ascii code below 32 or above
+ * 127 (included) using a dot '.' character.
+*/
+static void redact_non_printables(struct strbuf *buf)
```
chriscool, line 15: As the name of the function is now redact_non_printables(), the subject of the commit message should use that name instead of strbuf_sanitize(). So it should be something like "version: refactor redact_non_printables()".

version.c
Unique-Usman: Ohh, I saw the TODO being placed at the begining of the comment in the codebase. I will put it at the bottom.

t/t5555-http-smart-common.sh
```diff
@@ -123,9 +123,19 @@ test_expect_success 'git receive-pack --advertise-refs: v1' '
 '
 
 test_expect_success 'git upload-pack --advertise-refs: v2' '
+	printf "agent=FAKE" >agent_os_name &&
+	if test_have_prereq WINDOWS
+	then
+		test_config transfer.advertiseOSVersion false
```
chriscool, line 129: After thinking again about it, I changed my mind and my opinion now is that it's probably better to always use `git config` instead of `test_config` when disabling transfer.advertiseOSVersion on Windows. Otherwise the following tests cannot easily reuse the files that are created with `os-version=...` as we saw with t5701 below. They would all have to disable transfer.advertiseOSVersion on Windows first which would be tedious if there are multiple tests that want to reuse the files.

It might be worth telling in a comment that we use `git config` instead of `test_config` so that other tests below can reuse the 'expect' file though.

t/t5701-git-serve.sh
```diff
@@ -8,13 +8,23 @@ export GIT_TEST_DEFAULT_INITIAL_BRANCH_NAME
 . ./test-lib.sh
 
 test_expect_success 'test capability advertisement' '
+	printf "agent=git/$(git version | cut -d" " -f3)" >agent_os_name &&
+	if test_have_prereq WINDOWS
+	then
+		test_config transfer.advertiseOSVersion false
```
chriscool, line 14: Here also I think we should use `git config` instead of `test_config` to disabling transfer.advertiseOSVersion on Windows, for the same reason as in my comment above.

chriscool: And yeah, here also it might be worth telling in a comment that we use `git config` instead of `test_config` so that other tests below can reuse the 'expect.base' file.

t/t5555-http-smart-common.sh
Unique-Usman: Yeah, this is well thought out. It is very true. Using test_config will lead to repetition. I will change it to git config.

t/t5701-git-serve.sh
Unique-Usman: Yeah, but, will git config be enough to test the transfer.advertiseOSVersion or we will need to test in another way ?

version.c
Unique-Usman: Yeah, this is true. I will change it.

t/t5701-git-serve.sh
chriscool: Yeah, I think we might need to add other tests to test `transfer.advertiseOSVersion` but we can do that later. I think it will become clearer how we should test everything when we have `osversion.command` and `osversion.format`.

Unique-Usman: Ok, noted, thank you.

PR #1854, opened 2024-12-24

Unique-Usman:
Changes from previous version(transfer.advertise_5).
 - Change the commit message of the last commit. 
 - Fixed some typo
 - change the order of the Mentored-by and Signed-off-by
 - Rename the agent_os_name file to agent_and_os_name

### DIFF
--- a/Documentation/config/transfer.txt
+++ b/Documentation/config/transfer.txt
@@ -125,3 +125,10 @@ transfer.bundleURI::
 transfer.advertiseObjectInfo::
 	When `true`, the `object-info` capability is advertised by
 	servers. Defaults to false.
+
+transfer.advertiseOSVersion::
+	When `true`, the `os-version` capability is advertised by clients and
+	servers. It makes clients and servers send to each other a string
+	representing the operating system name, like "Linux" or "Windows".
+	This string is retrieved from the 'sysname' field of the struct returned
+	by the uname(2) system call. Defaults to true.
--- a/Documentation/gitprotocol-v2.txt
+++ b/Documentation/gitprotocol-v2.txt
@@ -190,6 +190,26 @@ printable ASCII characters except space (i.e., the byte range 32 < x <
 and debugging purposes, and MUST NOT be used to programmatically assume
 the presence or absence of particular features.
 
+os-version
+~~~~~~~~~~
+
+In the same way as the `agent` capability above, the server can
+advertise the `os-version` capability with a value `X` (in the form
+`os-version=X`) to notify the client that the server is running an
+operating system that can be identified by `X`. The client may
+optionally send its own `os-version` string by including the
+`os-version` capability with a value `Y` (in the form `os-version=Y`)
+in its request to the server (but it MUST NOT do so if the server did
+not advertise the os-version capability). The `X` and `Y` strings may
+contain any printable ASCII characters except space (i.e., the byte
+range 32 < x < 127), and are typically made from the result of
+`uname -s`(OS name e.g Linux). The os-version capability can be disabled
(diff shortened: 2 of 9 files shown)


chriscool:
About "This commit introduces" in the commit message of the third patch, we try to use either a verb in the imperative form like "Introduce ..." or something like "Let's introduce ..." instead. Please take a look at 'Documentation/SubmittingPatches' which contains an "imperative-mood" section.

chriscool:
Also in the same commit message:
 
- s/is added for privacy concern issue/is added to address privacy concerns/
- s/It is default/It defaults/

chriscool:
About the following paragraph in the same commit message:

"To address these inconsistencies, ongoing discussions are being held with my mentor to determine the best approach. Until resolved, the `transfer.advertiseOSVersion` is set to `false` on Windows during testing to avoid confusion."

I don't think talking about discussions between us is a good thing. When you send this patch series, discussions between us should be over and what matters then are the discussions with the mailing list and with Junio. So I think it's better to just say something like:

"Until a good way to test the feature on Windows is found, the `transfer.advertiseOSVersion` is set to `false` on Windows during testing."

Unique-Usman:
Thanks for the review. I made the necessary changes in #1856

dscho:
> Thanks for the review. I made the necessary changes in #1856

Please just force-push to the same branch. This whole "versioning in the branch name" business appears quite unlike Git.

Unique-Usman:
> > Thanks for the review. I made the necessary changes in #1856
> 
> Please just force-push to the same branch. This whole "versioning in the branch name" business appears quite unlike Git.

Hi @dscho, there is couples of reason why there is couple of branches.  This is a quote from https://git.github.io/Mentoring-Program-Guide/ in the *Draft patch series and branches* section. 

"For example, if you are working on git bisect you could make some initial changes on a branch named bisect1, then, after receiving some comments on it, make further changes on bisect2, and so on. This way your mentors and others could check what you did using a diff, or a range-diff, between the branches."

Having multiple branches also makes it easier for future interns to quickly review the progress and contributions of previous projects, especially when their role involves continuing or building upon the existing work.

That is the initial reason why I was not opening any PR on the main git/git.

dscho:
Please use tags where tags are due, not branches. (And plese only tag revisions that are worth having, as opposed to every iteration including known-incorrect ones.)

And I fear that you over-index on the intention to teach future readers by diligently keeping a record of every step and mis-step along the way: In my experience, nobody looks at such records ever again, and I doubt that you will, either. It is _vastly_ better use of your time and focus to aim for recording the _relevant_ parts of the patches' history in their commit messages. For example, if you tried a variation of the patch and found out that it does not work, describe that approach and the reasoning in the commit message. This will make it substantially easier to learn from your iterations than throwing dozens of suffix-versioned branches at an interested future reader. See also https://github.blog/developer-skills/github/write-better-commits-build-better-projects/. 

Besides, having multiple revisions is pretty useless without the context provided by the conversations that led to the latter conversations.

Force-pushing to the same PR keeps the context of those conversations, and _also_ records the history of the iterations (and in a much less error-prone way than "suffix-versioned branches").

Using branch name suffixes for versioning, and opening new PRs left and right (so that inevitably, you start commenting on an outdated PR because you successfully confused _yourself_), does a much worse job at keeping the context. Sure, it's all there, just scattered and much less useful than if you used best practice (which is: keep iterating within the same PR, keep all conversations there, too, and coalesce important learnings in the final commit messages).

tl;dr if there is an easy, paved path that _also_ delivers superior results, why not go ahead and use it?

firoj0 (commented):
Hello world

PR #1856, opened 2024-12-25

Unique-Usman:
Changes from previous version(transfer.advertise_6).
- Fix typos
- Use imperative-mood in the commit message


dscho:
> Changes from previous version(transfer.advertise_6).
> - Fix typos
> - Use imperative-mood in the commit message
> 

Why not accompany this change log with a range-diff? This would also make it easier to review the changes, and to figure out much quicker what the corresponding commit OIDs are.

PR #1866, opened 2025-01-03
Convert unittest 2 clar

Seyi007:


dscho:
@Seyi007 please edit [the PR description](https://github.com/git/git/pull/1866#issue-2767745667); It will be sent as cover letter (and it would be better if it was something different than the boiler-plate text).

t/unit-tests/generate-clar-decls.sh
```diff
@@ -14,6 +14,6 @@ do
 	suite_name=$(basename "$suite")
```
pks-t, line 14: Regarding the commit message: we don't typically say "This change", but use more imperative style, as if instructing the code to change. So you'd rather say "Change the function signatures...". Note though that it's not the function signatures that you're changing -- you are changing the regular expression that is used to _extract_ function signatures.

It would also be helpful to give a bit of context what "generate-clar-decls.sh" does and what the issue is that you've got with it that you're trying to fix.

Makefile
```diff
@@ -1338,14 +1338,14 @@ THIRD_PARTY_SOURCES += $(UNIT_TEST_DIR)/clar/%
 THIRD_PARTY_SOURCES += $(UNIT_TEST_DIR)/clar/clar/%
```
pks-t, line 1338: Same comment here regarding the commit message, it should use imperative style. This is also documented in `Documentation/SubmittingPatches`, anchor "imperative-mood".

I'd also adapt the subject a bit to indicate that you're converting "hash tests". Right now it reads as if you were changing the hash itself. For example: "t/unit-tests: convert hash tests to use clar".

t/meson.build
```diff
@@ -41,7 +42,6 @@ test('unit-tests', clar_unit_tests)
 
 unit_test_programs = [
   'unit-tests/t-example-decorate.c',
-  'unit-tests/t-hash.c',
```
pks-t, line 44: This whole commit should be merged into the preceding commit as it only contains fixups for it.

t/unit-tests/u-hash.c
```diff
@@ -0,0 +1,93 @@
...
+/* Only works with a literal string, useful when it contains a NUL character. */
+#define TEST_HASH_LITERAL(literal, expected_sha1, expected_sha256) { \
+	const char *expected_hashes[] = { expected_sha1, expected_sha256 }; \
+	check_hash_data(literal, (sizeof(literal) - 1), expected_hashes); \
+}
+
+/*  */
+void test_hash__empty_string(void) TEST_HASH_STR("",
+	"da39a3ee5e6b4b0d3255bfef95601890afd80709",
+	"e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855")
```
pks-t, line 37: Hm, unfortunately this macro breaks syntax parsers when not part of of a proper function body. Might be that my idea wasn't that great after all. We might consider instead converting all of these to look like this:

```
void test_hash__empty_string(void)
{
	TEST_HASH_STR("", "da39a3ee5e6b4b0d3255bfef95601890afd80709",
		"e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855");
}
```

Which is a bit more boilerplate, but not all that bad.

Feel free to either do that now or to use the current version and wait for upstream feedback first.

pks-t:
I'd also recommend to slim down the cover letter a bit. It contains too much technical information that isn't really of much interest as a summary for the patch series. You should rather give a broad overview of the intent of your patch series without going into the details of how exactly you do them.

t/unit-tests/generate-clar-decls.sh
Seyi007: Right, got it!

Makefile
Seyi007: Okay

t/unit-tests/u-hash.c
Seyi007: Yeah, I think we should test the waters and see the feedback from upstream.

t/unit-tests/u-hash.c
```diff
@@ -19,66 +16,94 @@ static void check_hash_data(const void *data, size_t data_length,
...
 	}
 }
 
 /* Works with a NUL terminated string. Doesn't work if it should contain a NUL character. */
-#define TEST_HASH_STR(data, expected_sha1, expected_sha256) do { \
-		const char *expected_hashes[] = { expected_sha1, expected_sha256 }; \
-		TEST(check_hash_data(data, strlen(data), expected_hashes), \
-		     "SHA1 and SHA256 (%s) works", #data); \
-	} while (0)
+#define TEST_HASH_STR(data, expected_sha1, expected_sha256) { \
```
pks-t, line 24: Let's add the `do` and `while (0)` back in. The effect of this is that users of the macro must finish it with a semicolon, which makes code parsers happy.

t/unit-tests/u-hash.c
```diff
@@ -19,66 +16,94 @@ static void check_hash_data(const void *data, size_t data_length,
...
+	const char *expected_hashes[] = { expected_sha1, expected_sha256 }; \
+	check_hash_data(data, strlen(data), expected_hashes); \
+}
 
 /* Only works with a literal string, useful when it contains a NUL character. */
-#define TEST_HASH_LITERAL(literal, expected_sha1, expected_sha256) do { \
```
pks-t, line 35: Here, as well.

PR #1873, opened 2025-01-13
t/unit-tests: convert mem-pool to use clar test framework

Seyi007:
Adapt the mem-pool test functions to clar framework by using clar assertions where necessary. Following the consensus to convert the unit-tests scripts found in the t/unit-tests folder to clar driven by Patrick Steinhardt.

Signed-off-by: Seyi Kuforiji <kuforiji98@gmail.com>


t/unit-tests/generate-clar-decls.sh
```diff
@@ -14,6 +14,7 @@ do
 	suite_name=$(basename "$suite")
 	suite_name=${suite_name%.c}
 	suite_name=${suite_name#u-}
+	suite_name=$(echo "$suite_name" | sed 's/-/_/g')
```
pks-t, line 17: This is fine as-is, but I think it's more customary in our codebase to use tr(1) instead. E.g. `echo "$suite_name" | tr '-' '_')`.

t/unit-tests/generate-clar-decls.sh
```diff
@@ -14,6 +14,7 @@ do
 	suite_name=$(basename "$suite")
```
pks-t, line 14: This commit should be ordered before the first one as it is a mandatory prerequisite. The Git project aims for "bisectability", which means that every single commit in your patch series should always compile and test alright. With the current ordering that is not the case.

t/unit-tests/u-mem-pool.c
```diff
@@ -0,0 +1,28 @@
+#include "unit-test.h"
+#include "mem-pool.h"
+
+static void t_calloc_100(struct mem_pool *pool)
```
pks-t, line 4: None of the testcases do anything with their local pool. So how about we make this take as parameter the `block_alloc` size and have the function here handle the setup and teardown of the pool?

t/unit-tests/u-reftable-tree.c
```diff
@@ -64,23 +64,25 @@ static void t_infix_walk(void)
...
+	cl_assert_equal_i(c.len, count);
 	tree_free(root);
 }
 
-int cmd_main(int argc UNUSED, const char *argv[] UNUSED)
+void test_reftable_tree__tree_search(void)
 {
-	TEST(t_tree_search(), "tree_search works");
-	TEST(t_infix_walk(), "infix_walk works");
+	t_tree_search();
```
pks-t, line 82: This is a bit curious. I would've expected you to inline `t_tree_search()` here.

t/unit-tests/u-reftable-tree.c
```diff
@@ -64,23 +64,25 @@ static void t_infix_walk(void)
...
+}
 
-	return test_done();
+void test_reftable_tree__infix_walk(void)
+{
+	t_infix_walk();
```
pks-t, line 87: Same here, I think this should be inlined.

PR #1879, opened 2025-01-22
t/unit-tests: convert hashmap test to use clar test framework Adapts

Seyi007:
hashmap test script to clar framework by using clar assertions where necessary. Test functions are created as both standalone and inline to test different test cases.


t/unit-tests/u-hashmap.c
```diff
@@ -165,39 +163,26 @@ static void t_add(struct hashmap *map, unsigned int ignore_case)
...
-						 entry->key, get_value(entry));
-					break;
-				}
+			int ret = key_val_contains(key_val, seen,
+							ARRAY_SIZE(key_val), entry);
+			cl_assert(ret == 0);
+			if (ret == 1) {
+				cl_check_fail("found entry was not given in the input\n");
+			} else if (ret == 2) {
+				cl_check_fail("duplicate entry detected\n");
```
pks-t, line 172: This if/else-cascade is impossible to reach: we already use `cl_assert(ret == 0)`, which would cause us to stop executing the testcase when the assert is violated. We can probably just drop it.

t/unit-tests/u-hashmap.c
```diff
@@ -165,39 +163,26 @@ static void t_add(struct hashmap *map, unsigned int ignore_case)
...
+		cl_assert_equal_i(count, 2);
 	}
 
 	for (size_t i = 0; i < ARRAY_SIZE(seen); i++) {
-		if (!check_int(seen[i], ==, 1))
-			test_msg("following key-val pair was not iterated over:\n"
-				 "    key: %s\n  value: %s",
-				 key_val[i][0], key_val[i][1]);
+		cl_assert_equal_i(seen[i], 1);
 	}
```
pks-t, line 182: You can also drop the braces around the for loop here -- we don't use them for single-line statements.

t/unit-tests/u-hashmap.c
```diff
@@ -165,39 +163,26 @@ static void t_add(struct hashmap *map, unsigned int ignore_case)
...
-						 entry->key, get_value(entry));
-					break;
-				case 2:
-					test_msg("duplicate entry detected\n"
-						 "    key: %s\n  value: %s",
-						 entry->key, get_value(entry));
-					break;
-				}
+			int ret = key_val_contains(key_val, seen,
+							ARRAY_SIZE(key_val), entry);
```
pks-t, line 167: I think indentation is off here. The line should start in the same column as `key_val`.

t/unit-tests/u-hashmap.c
```diff
@@ -211,24 +196,25 @@ static void t_remove(struct hashmap *map, unsigned int ignore_case)
...
 	entry = alloc_test_entry("notInMap", "", ignore_case);
-	check_pointer_eq(hashmap_remove_entry(map, entry, ent, "notInMap"), NULL);
+	cl_assert_equal_p(hashmap_remove_entry(map, entry, ent, "notInMap"), NULL);
 	free(entry);
 
-	check_int(map->tablesize, ==, 64);
-	check_int(hashmap_get_size(map), ==, ARRAY_SIZE(key_val) - ARRAY_SIZE(remove));
+	cl_assert_equal_i(map->tablesize, 64);
+	cl_assert_equal_i(hashmap_get_size(map),
+				ARRAY_SIZE(key_val) - ARRAY_SIZE(remove));
```
pks-t, line 217: Indentation is off here, too.

t/unit-tests/u-hashmap.c
```diff
@@ -242,38 +228,28 @@ static void t_iterate(struct hashmap *map, unsigned int ignore_case)
...
-			}
+		int ret = key_val_contains(key_val, seen,
+						ARRAY_SIZE(key_val),
+						entry);
+		cl_assert(ret == 0);
+		switch (ret) {
+		case 1:
+			cl_check_fail("found entry was not given in the input\n");
+		case 2:
+			cl_check_fail("duplicate entry detected\n");
```
pks-t, line 244: Same comment here regarding reachability of this block.

t/unit-tests/u-hashmap.c
```diff
@@ -242,38 +228,28 @@ static void t_iterate(struct hashmap *map, unsigned int ignore_case)
...
 		}
 	}
 
 	for (size_t i = 0; i < ARRAY_SIZE(seen); i++) {
-		if (!check_int(seen[i], ==, 1))
-			test_msg("following key-val pair was not iterated over:\n"
-				 "    key: %s\n  value: %s",
-				 key_val[i][0], key_val[i][1]);
+		cl_assert_equal_i(seen[i], 1);
 	}
```
pks-t, line 250: We can drop the braces.

t/unit-tests/u-hashmap.c
```diff
@@ -284,78 +260,91 @@ static void t_alloc(struct hashmap *map, unsigned int ignore_case)
...
-			test_msg("strintern(%s) returns %s\n", values[i], i1);
-		else if (!check(i1 != values[i]))
-			test_msg("strintern(%s) returns input pointer\n",
-				 values[i]);
-		else if (!check_pointer_eq(i1, i2))
-			test_msg("address('%s') != address('%s'), so strintern('%s') != strintern('%s')",
-				 i1, i2, values[i], values[i]);
-		else
-			check_str(i1, values[i]);
+		cl_assert_equal_s(strcmp(i1, values[i]), 0);
```
pks-t, line 309: This is wrong: `cl_assert_equal_s()` expects two strings, but you pass it two integers. You can pass `i1` and `values[i]` directly.

t/unit-tests/u-hashmap.c
```diff
@@ -284,78 +260,91 @@ static void t_alloc(struct hashmap *map, unsigned int ignore_case)
...
+		cl_assert_equal_p(i1, i2);
+		cl_assert_equal_s(i1, values[i]);
 	}
 }
 
-int cmd_main(int argc UNUSED, const char **argv UNUSED)
+void test_hashmap__replace(void)
+{
+	setup(t_replace, 0);
+	setup(t_replace, 1);
```
pks-t, line 319: I wonder whether we want to split up all testcases into two: one for case-sensitive, one for case-insensitive.

Seyi007: That is a function for each test case which means we have more standalone functions right?

t/unit-tests/u-example-decorate.c
```diff
@@ -0,0 +1,104 @@
...
+}
+
+void test_example_decorate__readd(void)
+{
+	struct object_id one_oid = { { 1 } }, two_oid = { { 2 } }, three_oid = { { 3 } };
+	struct test_vars vars = { 0 };
+
+	vars.one = lookup_unknown_object(the_repository, &one_oid);
+	vars.two = lookup_unknown_object(the_repository, &two_oid);
+	vars.three = lookup_unknown_object(the_repository, &three_oid);
```
pks-t, line 74: This setup code is repeated across all tests. You could instead introduce a function `void test_exmaple_decorate__init(void)` that initializes a global `vars` variable.

t/unit-tests/u-example-decorate.c
```diff
@@ -0,0 +1,104 @@
...
+
+	t_lookup(&vars);
+	clear_decoration(&vars.n, NULL);
```
pks-t, line 90: `clear_decoration()` could similarly be executed via a `test_example_decorate__cleanup()` function. That would only leave `t_lookup(&vars)`, which could then be inlined.

t/unit-tests/u-example-decorate.c
```diff
@@ -11,64 +11,84 @@ struct test_vars {
...
+}
+
+void test_example_decorate__add(void)
+{
+	t_add(&vars);
 
-	return test_done();
+}
+
+void test_example_decorate__readd(void)
```
pks-t, line 81: Adding these new functions is now unnecessary. I'd rather rename e.g. `t_readd()` to `test_example_decorate__readd()`.

t/unit-tests/u-example-decorate.c
```diff
@@ -11,64 +11,84 @@ struct test_vars {
...
 {
-	void *ret = add_decoration(&vars->n, vars->one, NULL);
+	void *ret = add_decoration(&vars->n, vars->one, &vars->decoration_a);
+	ret = add_decoration(&vars->n, vars->one, NULL);
 
-	check(ret == &vars->decoration_a);
+	cl_assert(ret == &vars->decoration_a);
 	ret = add_decoration(&vars->n, vars->two, &vars->decoration_b);
-	check(ret == NULL);
+	cl_assert(ret == NULL);
```
pks-t, line 32: Handling of `vars->one` seems true to the original, but `vars->two` doesn't. In the original, we first had a call to `add_decoration(&vars->n, vars->two, NULL)` before calling `add_decoration(&vars->n, vars->two, &vars->decoration_b)`. This would assert that we do see the `NULL` value as expected.

t/unit-tests/u-example-decorate.c
```diff
@@ -11,64 +11,84 @@ struct test_vars {
...
 }
 
 static void t_lookup(struct test_vars *vars)
 {
-	void *ret = lookup_decoration(&vars->n, vars->one);
+	void *ret;
 
-	check(ret == NULL);
+	add_decoration(&vars->n, vars->two, &vars->decoration_b);
```
pks-t, line 39: This looks mostly as expected, but you'd also have to `add_decoration(&vars->n, vars->one, NULL)` to remain true to the original. Again, this asserts that we can look up `NULL` values.

t/unit-tests/u-example-decorate.c
```diff
@@ -11,64 +11,84 @@ struct test_vars {
...
 	int objects_noticed = 0;
 
+	add_decoration(&vars->n, vars->one, &vars->decoration_a);
+	add_decoration(&vars->n, vars->two, &vars->decoration_b);
 	for (size_t i = 0; i < vars->n.size; i++) {
 		if (vars->n.entries[i].base)
 			objects_noticed++;
 	}
-	check_int(objects_noticed, ==, 2);
+	cl_assert_equal_i(objects_noticed, 2);
```
pks-t, line 58: Yup, this looks as expected.

t/unit-tests/u-example-decorate.c
```diff
@@ -0,0 +1,73 @@
...
+	struct decoration n;
+	int decoration_a, decoration_b;
+};
+
+static struct test_vars vars;
+
+void test_example_decorate__add(void)
+{
+	void *ret = add_decoration(&vars.n, vars.one, &vars.decoration_a);
+
```
pks-t, line 19: Nit: I'd remove the empty newline here, as it doesn't really help to clarify structure. Also applies to the other tests, except when declaring `void *ret;` without assigning it a value as you do in `__lookup()`.

t/unit-tests/u-example-decorate.c
```diff
@@ -0,0 +1,73 @@
...
+	cl_assert(ret == NULL);
+}
+
+void test_example_decorate__readd(void)
+{
+	void *ret = add_decoration(&vars.n, vars.one, &vars.decoration_a);
+
+	ret = add_decoration(&vars.n, vars.one, NULL);
+	cl_assert(ret == &vars.decoration_a);
+	add_decoration(&vars.n, vars.two, NULL);
```
pks-t, line 31: You should probably assert the return value here, as well.

t/unit-tests/u-example-decorate.c
```diff
@@ -0,0 +1,74 @@
...
+
+void test_example_decorate__readd(void)
+{
+	void *ret = add_decoration(&vars.n, vars.one, &vars.decoration_a);
+	ret = add_decoration(&vars.n, vars.one, NULL);
+	cl_assert(ret == &vars.decoration_a);
+	ret = add_decoration(&vars.n, vars.two, NULL);
+	cl_assert(ret == NULL);
+	ret = add_decoration(&vars.n, vars.two, &vars.decoration_b);
+	cl_assert(ret == NULL);
```
pks-t, line 32: This test here still intermingles the setup with the actual logic you want to test. Something like this would help:

```

void test_example_decorate__readd(void)
{
	void *ret

	cl_assert(add_decoration(&vars.n, vars.one, &vars.decoration_a) == NULL);
	cl_assert(add_decoration(&vars.n, vars.two, NULL) == NULL);

	ret = add_decoration(&vars.n, vars.one, NULL);
	cl_assert(ret == &vars.decoration_a);
	ret = add_decoration(&vars.n, vars.two, &vars.decoration_b);
	cl_assert(ret == NULL);
}
```

t/unit-tests/u-strbuf.c
```diff
@@ -0,0 +1,122 @@
...
+	/* Initialized strbufs should always have a non-NULL buffer */
+	cl_assert(buf->buf != NULL);
+	/* Buffers should always be NUL-terminated */
+	cl_assert(buf->buf[buf->len] == '\0');
+	/*
+	 * Freshly-initialized strbufs may not have a dynamically allocated
+	 * buffer
+	 */
+	if (buf->len == 0 && buf->alloc == 0)
+		return 1;
```
pks-t, line 41: There is no need for return values in this function as the `cl_assert()` functions would immediately abort execution anyway. So the function can be converted to return `void`, only.

t/unit-tests/u-strcmp-offset.c
```diff
@@ -15,21 +15,32 @@ static void check_strcmp_offset(const char *string1, const char *string2,
...
+	cl_assert_equal_i(result, expect_result);
+	cl_assert_equal_i((uintmax_t)offset, expect_offset);
 }
 
 #define TEST_STRCMP_OFFSET(string1, string2, expect_result, expect_offset) \
-	TEST(check_strcmp_offset(string1, string2, expect_result,          \
-				 expect_offset),                           \
-	     "strcmp_offset(%s, %s) works", #string1, #string2)
+	check_strcmp_offset(string1, string2, expect_result,          \
+				 expect_offset)
```
pks-t, line 24: I don't really see any benefit in retaining this macro. Previously we were doing some extra stuff, but now it merely forwards the four arguments to `check_strcmp_offset()`. So I'd drop the macro and instead call the function directly.

t/unit-tests/u-strcmp-offset.c
```diff
@@ -15,21 +15,32 @@ static void check_strcmp_offset(const char *string1, const char *string2,
...
+
+void test_strcmp_offset__different(void)
+{
 	TEST_STRCMP_OFFSET("abc", "def", -1, 0);
+}
+
+void test_strcmp_offset__different_length(void)
+{
 	TEST_STRCMP_OFFSET("abc", "abz", -1, 2);
 	TEST_STRCMP_OFFSET("abc", "abcdef", -1, 3);
```
pks-t, line 45: Should these be two different tests?

Seyi007: They do the same thing which checks the difference between the two string. Do you think they should be separated?

t/unit-tests/u-strbuf.c
angelass921: Hello said I had been framed know we've gotten deadline for finishing all our codes til February 23rd

PR #1892, opened 2025-02-12
t/unit-tests: convert oid-array test to use clar test framework

Seyi007:
Adapt oid-array test script to clar framework by using clar assertions where necessary.

Signed-off-by: Seyi Kuforiji <kuforiji98@gmail.com>


t/unit-tests/u-oid-array.c
```diff
@@ -34,30 +31,22 @@ static void t_enumeration(const char **input_args, size_t input_sz,
...
+
+	for (i = 0; i < actual.nr; i++)
+		cl_assert(oideq(&actual.oid[i], &expect.oid[i]));
 
 	oid_array_clear(&actual);
 	oid_array_clear(&input);
 	oid_array_clear(&expect);
 }
 
 #define TEST_ENUMERATION(input, expect, desc)                                     \
```
pks-t, line 48: The description isn't used, so it can be dropped.

t/unit-tests/unit-test.c
```diff
@@ -2,6 +2,9 @@
 #include "parse-options.h"
 #include "string-list.h"
 #include "strvec.h"
+#include "strbuf.h"
+#include "hex.h"
+#include "unit-test.h"
```
pks-t, line 7: Let's keep headers sorted alphabetically (except for "unit-test.h", which is expected to come first. You also got that one duplicated now.

t/unit-tests/unit-test.c
```diff
@@ -62,3 +65,44 @@ int cmd_main(int argc, const char **argv)
...
+	if (algo < 0) {
+		const char *algo_name = getenv("GIT_TEST_DEFAULT_HASH");
+		algo = algo_name ? hash_algo_by_name(algo_name) : GIT_HASH_SHA1;
+
+		cl_assert(algo != GIT_HASH_UNKNOWN);
+	}
+	return algo;
+}
+
+static int get_oid_arbitrary_hex_algop(const char *hex, struct object_id *oid,
```
pks-t, line 82: I wonder whether we want to give these functions names which give a bit of a better hint that they are part of the test suite. E.g. `cl_parse_oid()`.

t/unit-tests/unit-test.c
```diff
@@ -62,3 +65,44 @@ int cmd_main(int argc, const char **argv)
...
+	cl_assert(sz <= algop->hexsz);
+
+	strbuf_add(&buf, hex, sz);
+	strbuf_addchars(&buf, '0', algop->hexsz - sz);
+
+	ret = get_oid_hex_algop(buf.buf, oid, algop);
+	cl_assert_equal_i(ret, 0);
+
+	strbuf_release(&buf);
+	return ret;
```
pks-t, line 98: The return value can be removed as you already assert that it's always 0.

t/unit-tests/unit-test.c
```diff
@@ -62,3 +65,44 @@ int cmd_main(int argc, const char **argv)
...
+}
+
+
+int get_oid_arbitrary_hex(const char *hex, struct object_id *oid)
```
pks-t, line 102: Same here. How about `cl_parse_any_oid()`?

t/unit-tests/unit-test.c
```diff
@@ -62,3 +65,44 @@ int cmd_main(int argc, const char **argv)
...
+{
+	int hash_algo = init_hash_algo();
+
+	cl_assert(hash_algo != GIT_HASH_UNKNOWN);
+	return get_oid_arbitrary_hex_algop(hex, oid, &hash_algos[hash_algo]);
```
pks-t, line 107: Do we want to also assert that the return value is 0? I guess that should be a safe assumption given that this is a helper. We can then drop the return value again.

t/unit-tests/u-oidmap.c
```diff
@@ -0,0 +1,152 @@
+#include "unit-test.h"
+#include "lib-oid.h"
```
pks-t, line 2: This header does not need to be included as it only contains the "old" assertions used by the other unit testing framework, right?

t/unit-tests/u-oidmap.c
```diff
@@ -0,0 +1,152 @@
...
+{
+	struct oidmap map = OIDMAP_INIT;
+
+	for (size_t i = 0; i < ARRAY_SIZE(key_val); i++){
+		struct test_entry *entry;
+
+		FLEX_ALLOC_STR(entry, name, key_val[i][1]);
+		cl_parse_any_oid(key_val[i][0], &entry->entry.oid);
+		entry = oidmap_put(&map, entry);
+		cl_assert(entry == NULL);
```
pks-t, line 31: You can probably drop the `entry` variable and directly `cl_assert_equal_p(oidmap_put(&map, entry), NULL)`. The same is also true in other cases.

t/unit-tests/u-oidmap.c
```diff
@@ -0,0 +1,152 @@
...
+struct test_entry {
+	struct oidmap_entry entry;
+	char name[FLEX_ARRAY];
+};
+
+static const char *const key_val[][2] = { { "11", "one" },
+					  { "22", "two" },
+					  { "33", "three" } };
+
+static void setup(void (*f)(struct oidmap *map))
```
pks-t, line 21: Instead of using a `setup()` function, let's use a `test_oidmap__initialize()` function that gets executed automatically and initializes a global variable, as well as a `test_oidmap__cleanup()` function that cleans up state. When we have that we can get rid of `setup()` and also of functions like `test_oidmap__replace()`, where `t_replace_()` itself would be renamed.

t/unit-tests/u-oidmap.c
```diff
@@ -0,0 +1,152 @@
...
+{
+	struct oidmap_iter iter;
+	struct test_entry *entry;
+	char seen[ARRAY_SIZE(key_val)] = { 0 };
+	int count = 0;
+
+	oidmap_iter_init(map, &iter);
+	while ((entry = oidmap_iter_next(&iter))) {
+		int ret;
+		cl_assert_equal_i(ret = key_val_contains(entry, seen), 0);
```
pks-t, line 127: `ret` is unneeded here, so we can drop it.

ssss1524s (commented):
معالجة جمبع الاخطا

ssss1524s:
معالجة السطر 732

PR #1909, opened 2025-03-06
ci: use Visual Studio for win+meson job on GitHub Workflows

pks-t:
In 7304bd2bc39 (ci: wire up Visual Studio build with Meson, 2025-01-22)
we have wired up a new CI job that builds and tests Git with Meson on a
Windows machine. The expectation here was that this build uses the
Visual Studio toolchain to do so, and that is true on GitLab CI. But on
GitHub Workflows it is not the cause because we've got GCC in our PATH,
and thus Meson favors that compiler toolchain over Visual Studio.

Fix this by explicitly asking Meson to use the Visual Studio toolchain.
While this is only really required for GitHub Workflows, let's also pass
the flag in GitLab CI so that we don't implicitly assume the toolchain
that Meson is going to pick.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -265,7 +265,7 @@ jobs:
       run: pip install meson ninja
     - name: Setup
       shell: pwsh
-      run: meson setup build -Dperl=disabled -Dcredential_helpers=wincred
+      run: meson setup build --vsenv -Dperl=disabled -Dcredential_helpers=wincred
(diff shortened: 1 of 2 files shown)


dscho:
Please don't invest more energy and time (as I do not want to be forced to spend mine) in making Meson work with Visual Studio. No Visual Studio-based contributor will ever use this, and https://xkcd.com/1205/ is real.

PR #1937, opened 2025-03-31
column: exit early when indent length is larger than width

lavendarlatte:
The code exits with "fatal size_t overflow" when indent length is larger than width. This is because when calculating cols of struct column_data, unsigned underflow happens and cols is set to negative value, then converted to size_t when calling REALLOC_ARRAY in shrink_columns() function. This can lead to allocating extremely large chunk of memory when succeeds, or crash when fails.

The change exits code early with failure reason to avoid underflow and clarify argument limitations. This change ensures that cols is always positive, making the code clearer. It also eliminates the need for warning suppression related to signed-unsigned comparisons, as cols can be safely converted to size_t.

/allow


dscho:
@lavendarlatte I took the liberty of deleting the boiler plate text from the PR description, as it would have been sent as part of the cover letter (which is clearly undesirable).

PR #1947, opened 2025-04-17
blame: ignore nonexistent ignore files

lf-:


t/t8013-blame-ignore-revs.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/nix/store/cvlbhhrvzfkjl2hrrzhq3vr5gzan1r60-bash-interactive-5.2p37/bin/sh
```
derrickstolee, line 1: This looks like a bad artifact.

SURAK605393: version: 2
updates:
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "monthly"

PR #1949, opened 2025-04-22
Add support for Oauth2 and fix message-id bug in outlook

AdityaGarg8:
This pull request introduces 2 changes:

1. It adds support for Oauth2 authentication, which is now compulsory my Microsoft. This patch has been rebased to the latest version from the original version at https://lore.kernel.org/git/20250125190131.48717-1-julian@swagemakers.org/

2. The second patch makes the script reply to the message id set by the outlook, since outlook has its own proprietary way to handle message ids, and does not allow user to set their own. As a result, threads were breaking.

Cc: M Hickford <mirth.hickford@gmail.com>, Julian Swagemakers <julian@swagemakers.org>, sandals@crustytoothpaste.net, Shengyu Qu <wiagn233@outlook.com>

### DIFF
--- a/git-send-email.perl
+++ b/git-send-email.perl
@@ -1799,6 +1799,17 @@ EOF
 			$smtp->datasend("$line") or die $smtp->message;
 		}
 		$smtp->dataend() or die $smtp->message;
+
+		# Retrieve the Message-ID from the server response in case of Outlook
+		if ($smtp_server eq 'smtp.office365.com' || $smtp_server eq 'smtp-mail.outlook.com') {
+			if ($smtp->message =~ /<([^>]+)>/) {
+				$message_id = "<$1>";
+				printf __("Outlook: Retrieved Message-ID: %s\n"), $message_id;
+			} else {
+				warn __("Warning: Could not retrieve Message-ID from server response.\n");
+			}
+		}
+
 		$smtp->code =~ /250|200/ or die sprintf(__("Failed to send %s\n"), $subject).$smtp->message;
 	}
 	if ($quiet) {
--- a/git-send-email.perl
+++ b/git-send-email.perl
@@ -1398,6 +1398,63 @@ sub smtp_host_string {
 	}
 }
 
+sub generate_oauthbearer_string {
+	# This will generate the oauthbearer string used for authentication.
+	#
+	# "n,a=" {User} ",^Ahost=" {Host} "^Aport=" {Port} "^Aauth=Bearer " {Access Token} "^A^A
+	#
+	# The first part `n,a=" {User} ",` is the gs2 header described in RFC5801.
+	# * gs2-cb-flag `n` -> client does not support CB
+	# * gs2-authzid `a=" {User} "`
+	#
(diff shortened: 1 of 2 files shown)


git-send-email.perl
```diff
@@ -1799,6 +1799,17 @@ sub send_message {
 			$smtp->datasend("$line") or die $smtp->message;
 		}
 		$smtp->dataend() or die $smtp->message;
+
+		# Retrieve the Message-ID from the server response in case of Outlook
+		if ($smtp_server eq 'smtp.office365.com' || $smtp_server eq 'smtp-mail.outlook.com') {
+			if ($smtp->message =~ /<([^>]+)>/) {
+				$message_id = "<$1>";
+				print __("Outlook: Retrieved Message-ID: $message_id\n");
```
dscho, line 1807: This [fails the build](https://github.com/git/git/actions/runs/14588452096/job/40918206044?pr=1949#step:4:1915). (I have no idea why the line number displayed in the error message does not match up with the line number in the diff, but am fairly confident that it is this line.)

@AdityaGarg8 I believe that this needs to be rewritten to a non-variable argument of `__()`, to allow for translating the message. Looking around in this file, I found [this existing instance](https://github.com/git/git/blob/00e1be73ab91db1c37447544ebcdc4f33432dbcb/git-send-email.perl#L362), which suggests that this should be written like this:

```perl
printf _("Outlook: Retrieved Message-ID: %s\n", $message_id);
```

AdityaGarg8: Thanks, I've applied this change

dscho: @AdityaGarg8 thank you! Unfortunately, the build failure [is still occurring](https://github.com/AdityaGarg8/git/actions/runs/14589216899/job/40920435996#step:4:1913).

I did notice that this is not even in the part where Coccinelle is used, but simply where the translations are built. In other words, you should be able to replicate the build failure locally by running `make .build/pot/po/git-send-email.perl.po`. Could you try that?

dscho: Since the win build failure is unrelated to your change, I suggest to ignore it.

AdityaGarg8: I've sent the v2 the traditional Linux kernel style. I thought the PRs were mergeable but it just turned out to be a helper bot to send to the mailing list :)

dscho: Hmm. Now you cannot use GitGitGadget anymore to send subsequent iterations. And your [v2](https://lore.kernel.org/git/PN3PR01MB959739CFCF4D1EFC14259F8BB8BB2@PN3PR01MB9597.INDPRD01.PROD.OUTLOOK.COM/T/#m6f3d15f7bb78993a8da6f19cab203dc2e18dd842) is not marked as a reply to v1, and lacks a range-diff. Those things would have been what GitGitGadget would have done for you.

But since you cannot use GitGitGadget to send those iterations anymore, you might just as well close this here PR.

AdityaGarg8: I'm sorry about that. I have never contributed to git and am used to linux kernel way or a proper github pull request.

PR #1962, opened 2025-05-12
pack-bitmap: fix memory leak if `load_bitmap` failed

brandb97:
Since it seems this patch has been inactive for some time, I have revised the comments according to Taylor's feedback and submitted a new version.

This patch prevents pack-bitmap.c:load_bitmap() from nulling bitmap_git->bitmap when loading failed. Thus eliminates memory leak. This patch also add a test case in t5310 which use clang leak sanitizer to detect whether leak happens when loading failed.

cc: Jeff King <peff@peff.net>


igs-raiderler:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

PR #1967, opened 2025-05-17
pack-bitmap: add loading corrupt bitmap_index test

brandb97:
This patch add a test function `test_bitmap_load_corrupt` in patch-bitmap.c , a `load corrupt bitmap` test case in t5310-pack-bitmaps.sh and a new command `load-corrupt` for `test-tool` in t/helper/test-bitmap.c.

To make sure we are loading a corrupt bitmap, we need enable bitmap table lookup so that `prepare_bitmap()` won't call `load_bitmap_entries_v1()`. So to test corrupt bitmap_index, we first `prepare_bitmap()` to set everything up but `bitmap_index->bitmaps` for us. Then we do any corruption we want to the bitmap_index. Finally we call `load_bitmap_entries_v1()` to test loading corrupt bitmap.

cc: Jeff King <peff@peff.net>

### DIFF
--- a/pack-bitmap.c
+++ b/pack-bitmap.c
@@ -3022,6 +3022,71 @@ cleanup:
 	return ret;
 }
 
+typedef void(corrupt_fn)(struct bitmap_index *);
+
+static int bitmap_corrupt_then_load(struct repository *r, corrupt_fn *do_corrupt)
+{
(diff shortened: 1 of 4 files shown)


ribflexir:
Huh?

PR #1985, opened 2025-06-03
bundle-uri: print_bundle_uri prints heuristic to stderr with fprintf rather than stdout with printf

mzr:
Small fix to bundle-uri convenience debug function.

cc: Derrick Stolee [stolee@gmail.com](mailto:stolee@gmail.com)


derrickstolee:
@mzr: Thanks for noticing this. It's clearly wrong.

In addition to your other things to fix, could you please explore why this wasn't being caught by the test suite? (Likely, it's due to tests that use test helpers and `fp` is `stdout` anyway.)

dscho:
> Commit not signed off

That's the [`Signed-off-by:` footers](https://git-scm.com/docs/SubmittingPatches#sign-off) you see elsewhere in the commit history.

mzr:
@derrickstolee yup, you're right. It's only being used with stdout. I sort of assumed it would be used with stderr somewhere. 

```
➜  rg print_bundle_list
bundle-uri.c
99:void print_bundle_list(FILE *fp, struct bundle_list *list)

bundle-uri.h
111:void print_bundle_list(FILE *fp, struct bundle_list *list);

t/helper/test-bundle-uri.c
66:     print_bundle_list(stdout, &list);
99:     print_bundle_list(stdout, transport->bundles);
```

PR #1994, opened 2025-06-12
git.c: remove the_repository dependence in run_builtin()

brandb97:
run_builtin() takes a repo parameter, so the use of the_repository is no longer necessary. Removed the usage of the_repository.

The comment before trace_repo_setup() advises not to use get_git_dir(), but this note is unrelated to trace_repo_setup() itself. Additionally, get_git_dir() has now been renamed to repo_get_git_dir(). Remove this comment line.


brandb97:
I know, that why I haven't submit. And now I am only use this pr to trigger ci test. Please allow me to keep doing testing on this pr.

PR #2004, opened 2025-06-27
Revert "sequencer: make it clearer that commit descriptions are just …

ZuBB:
Hi there,

Its a not a real PR, is more like a way to bring up a discussion about new hardcoded way of creating a log for rebase process.

in v2.50 when doing rebase user will see smth like next

```
pick 03cfb8fb7 # WIP
pick d5f152e94 # WIP
pick 4e672c9cf # WIP
...
```

Before 2.50 that `#` wasnt there. I completely understand reasoning for adding it, but I do not understand why everyone has to face it. New users may do a mistake mentioned in the original commit [e42667241d](https://github.com/git/git/commit/e42667241de12840ef58c0ba1c060b86c850bae0), but experienced users probably dont. Why everyone has to see that char in the log? I am not competent engouh to do a proper PR for it, but I would like to ask you to consider adding a config option for opting out of this.

PS: I see that reactions are allowed here, so I would like to ask ppl to support it through voting (👍🏻 / 👎🏻).

### DIFF
--- a/sequencer.c
+++ b/sequencer.c
@@ -5904,11 +5904,11 @@ static int make_script_with_merges(struct pretty_print_context *pp,
 
 		/* Create a label from the commit message */
 		strbuf_reset(&label_from_message);
-		if (skip_prefix(oneline.buf, "# Merge ", &p1) &&
+		if (skip_prefix(oneline.buf, "Merge ", &p1) &&
 		    (p1 = strchr(p1, '\'')) &&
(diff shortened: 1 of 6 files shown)


dscho:
Thank you for writing up a bug report. However, this is not the place the Git projects expects bug reports: please send it to the Git mailing list. See https://git-scm.com/community for more details.

PR #2008, opened 2025-07-02
docs: clarify FETCH_HEAD overwrite behavior in gittutorial

monteiz:
Adds a note explaining that FETCH_HEAD is overwritten on every fetch, to avoid misunderstandings when fetching from multiple sources.


monteiz:
Profile updated.

dscho:
@monteiz may I suggest trying a `/preview`?

PR #2012, opened 2025-07-16
Update t7450-bad-git-dotfiles.sh

cjhxmx:
When post-checkout is executed, the foo file is actually created in the "$PWD/bad-clone/sub" path.


dscho:
@cjhxmx you will want to fix the commit message by adding a [`Signed-off-by` line](https://git-scm.com/docs/SubmittingPatches#dco) and by following the guidance in https://github.blog/2022-06-30-write-better-commits-build-better-projects/ to improve it, in particular with a strong focus on this part:

-- | -- | --

PR #2040, opened 2025-08-26
alloc: fix dangling pointer in alloc_state cleanup

nouraellm:
cc: Torsten Bögershausen <tboegi@web.de>
cc: Jeff King <peff@peff.net>

### DIFF
--- a/alloc.c
+++ b/alloc.c
@@ -36,19 +36,25 @@ struct alloc_state {
 	int slab_nr, slab_alloc;
 };
 
-struct alloc_state *allocate_alloc_state(void)
+struct alloc_state *alloc_state_alloc(void)
 {
 	return xcalloc(1, sizeof(struct alloc_state));
 }
 
-void clear_alloc_state(struct alloc_state *s)
+void alloc_state_free_and_null(struct alloc_state **s_)
 {
+	struct alloc_state *s = *s_;
+
+	if (!s)
+	    return;
+
 	while (s->slab_nr > 0) {
 		s->slab_nr--;
 		free(s->slabs[s->slab_nr]);
 	}
 
 	FREE_AND_NULL(s->slabs);
+	FREE_AND_NULL(*s_);
 }
 
 static inline void *alloc_node(struct alloc_state *s, size_t node_size)
--- a/alloc.h
+++ b/alloc.h
@@ -14,7 +14,7 @@ void *alloc_commit_node(struct repository *r);
 void *alloc_tag_node(struct repository *r);
 void *alloc_object_node(struct repository *r);
 
-struct alloc_state *allocate_alloc_state(void);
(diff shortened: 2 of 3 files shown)


nouraellm:
True, my bad, didn't pay attention to it, sorry.
In C90, all variable declarations must appear at the top of a block
So we cannot place the NULL check before the declaration.

However, assigning a ptr by dereferencing another ptr before checking is an
undefined behavior.
To avoid this, we should check that s_ and *s_ are non-NULL before
dereferencing.

A safe C90-compatible approach is to:

diff --git a/alloc.c b/alloc.c
index c1065551c0..be6a9edf10 100644
--- a/alloc.c
+++ b/alloc.c
@@ -43,10 +43,12 @@ struct alloc_state *alloc_state_alloc(void)

 void alloc_state_free_and_null(struct alloc_state **s_)
 {
-       struct alloc_state *s = *s_;
+       struct alloc_state *s;

        if (!s_ || !*s_) return;

+       s = *s_;
+
        while (s->slab_nr > 0) {
                s->slab_nr--;
                free(s->slabs[s->slab_nr]);

If you confirm, I can then push the changes.

nouraellm:
/submit

Submitting again, since this doesn't seem to include newly added changes (if that's a wrong approach, please feel free to correct me)

> This patch series was integrated into seen via [be4e5af](https://github.com/git/git/commit/be4e5af6e78ee4f6700e053e68054b9ab8b3e856).

nouraellm:
The point of the check was to avoid subtle crashes
aligning with defensive programming requirements.

If you lean more towards strict contract enforcement
Just say the word.

dscho:
> The point of the check was to avoid subtle crashes aligning with defensive programming requirements. If you lean more towards strict contract enforcement Just say the word.
> […](#)
> On Thu, Sep 4, 2025 at 9:53 AM gitgitgadget-git[bot] < ***@***.***> wrote: *gitgitgadget-git[bot]* left a comment ([git/git#2040](https://github.com/git/git/pull/2040)) <[#2040 (comment)](https://github.com/git/git/pull/2040#issuecomment-3252375885)
> On the Git mailing list ***@***.***>, Junio C Hamano wrote (reply to this <https://gitgitgadget.github.io/reply-to-this>): "ノウラ | Flare via GitGitGadget" ***@***.***> writes:
> +void alloc_state_free_and_null(struct alloc_state **s_)> {
> + struct alloc_state *s;> +
> + if (!s_ || !*s_) return;
> [...]

@nouraellm while you replied via email, unfortunately you replied to the PR comment that mirrored the actual email. And therefore, the reply won't reach the intended recipient 😞 

Please follow the guidance in https://gitgitgadget.github.io/reply-to-this to reply to [Junio's email](https://lore.kernel.org/git/xmqqh5xibqvu.fsf@gitster.g).

PR #2049, opened 2025-09-11
Update Unix mailbox date in patch format

walter40:
The original Unix mailbox time for the patch format was chosen arbitrarily but there are cases when this time could have collided with the actual time of somebody making a patch. There is a date extremely close to the original which should be used instead because the chance is far less likely that anyone would have been making a patch at that time.


dscho:
@walter40 just for your information: The Git project refuses to accept Pull Requests. I created GitGitGadget to help bridge that gap, but you'll still have to `/submit` this PR, otherwise your contribution will be ignored by the Git maintainer.

PR #2053, opened 2025-09-12
lockfile: add case insensitive filesystem note

adacosta:
When running git fetch or git pull on a case insensitive filesystem (e.g., default macOS), if multiple case variants of the same remote ref exist (often after a case-only rename), both variant locks map to the same on-disk path. When a local update is required, Git creates a lock for the first variant and then attempts to lock the second, which collides with the same lock file, so an “existing lock” error is reported. The underlying issue is mixed-case refs; resolve it by consolidating the remote to a single-case variant and update local refs accordingly.

cc: Karthik Nayak <karthik.188@gmail.com>

### DIFF
--- a/lockfile.c
+++ b/lockfile.c
@@ -156,7 +156,10 @@ void unable_to_lock_message(const char *path, int err, struct strbuf *buf)
 		    "an editor opened by 'git commit'. Please make sure all processes\n"
 		    "are terminated then try again. If it still fails, a git process\n"
 		    "may have crashed in this repository earlier:\n"
-		    "remove the file manually to continue."),
+		    "remove the file manually to continue.\n\n"
+		    "On case insensitive client filesystems, multiple mixed-case refs will resolve\n"
+		    "to the same lock file, possibly causing this error. If so, ensure your remote\n"
+		    "refs have a single case variant."),
 			    absolute_path(path), strerror(err));
 	} else
 		strbuf_addf(buf, _("Unable to create '%s.lock': %s"),


dscho:
@adacosta the Git project is _very_ particular with commit messages. You will want to review the guidance in https://github.blog/2022-06-30-write-better-commits-build-better-projects/ to improve your commit message (`git commit --amend`), in particular with a strong focus on this part:

-- | -- | --

Also, could you please sign off on your patch? The Git project [insists on that](https://git-scm.com/docs/SubmittingPatches#dco). You can fix it by `git commit --amend --signoff`, and force-pushing.

I am sorry for all the red tape...

stratigos:
I think this PR will be helpful to developers facing this issue, as it eliminates the time it takes to understand the issue and fix it within one's repository or filesystem.

adacosta:
@dscho thanks for the feedback! Not included in my patch are the translations because I'm not sure how they're handled. Does my PR need to be flagged to translate? For an example of files, search "Another git process seems to be running in this repository".

dscho:
> Not included in my patch are the translations because I'm not sure how they're handled. Does my PR need to be flagged to translate?

The message you modify is enclosed in `_(...)`. That marks it for translation, and the excellent translators over at https://github.com/git-l10n/git-po typically do a very good job at adjusting the translations during the "-rc phase" short before the next `.0` version comes out.

dscho:
@adacosta did you mean to `/submit` this?

PR #2065, opened 2025-10-01
Makefile update libgit.a: Include xdiff and reftable in libgit.a

ezekielnewren:
Changes in v2:
  * Add xdiff and reftable objects directly to LIB_OBJS.
  * Explain why xdiff-objs is removed.

Original cover letter:
---
Add xdiff and reftable to the static library libgit.a that Makefile produces. Meson does not require any changes since it already includes those libraries. The motivation is to simplify Rust's job of linking against the C code by requiring it to 
only link against a single static library (libgit.a).

The Rust compiler only needs to know how to link against libgit.a in 2 cases that I can think of:
  * Rust unit tests
  * Rust defining the main function

Otherwise Rust can be compiled without linking, and then Makefile and Meson can use Cargo's produced static lib files to build Git.

Note: The flag -fPIE or -fPIC is required for Makefile to build libgit.a in a way that Cargo can use. It has been deliberately omitted from the Makefile, for now, since Rust isn't part of Git (yet).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -918,7 +918,6 @@ export PYTHON_PATH
 TEST_SHELL_PATH = $(SHELL_PATH)
 
 LIB_FILE = libgit.a
-REFTABLE_LIB = reftable/libreftable.a
 
 GENERATED_H += command-list.h
 GENERATED_H += config-list.h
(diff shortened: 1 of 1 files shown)


nadinenpendleton81:
Inputdata=("googprice"),"daily",2018,1,1")
Importdata=("googprice")"2018,1,1start query

PR #2148, opened 2025-12-30
fsmonitor: fix khash memory leak in do_handle_client

ptarjan:
The do_handle_client() function allocates a khash table to de-duplicate
pathnames when responding to client requests. However, kh_release_str()
was used instead of kh_destroy_str(). The release function only frees
internal arrays (flags, keys, vals) but not the struct itself, which is
allocated by kh_init_str() via xcalloc. This caused a 40-byte leak per
client request.

Fix by using kh_destroy_str() which properly frees both internal arrays
and the struct itself. Also move the cleanup to the cleanup section and
initialize shown to NULL so that kh_destroy_str() is safe to call on all
exit paths.

Changes since v1:
- Removed incorrect claim about interrupted threads reaching cleanup
- Simplified commit message to focus on the real bug (kh_release vs kh_destroy)

Signed-off-by: Paul Tarjan <github@paulisageek.com>

cc: René Scharfe <l.s.r@web.de>
cc: Paul Tarjan <paul@paultarjan.com>

### DIFF
--- a/builtin/fsmonitor--daemon.c
+++ b/builtin/fsmonitor--daemon.c
@@ -671,7 +671,7 @@ static int do_handle_client(struct fsmonitor_daemon_state *state,
 	const struct fsmonitor_batch *batch;
 	struct fsmonitor_batch *remainder = NULL;
 	intmax_t count = 0, duplicates = 0;
-	kh_str_t *shown;
+	kh_str_t *shown = NULL;
 	int hash_ret;
 	int do_trivial = 0;
 	int do_flush = 0;
@@ -909,8 +909,6 @@ static int do_handle_client(struct fsmonitor_daemon_state *state,
 		total_response_len += payload.len;
 	}
 
-	kh_release_str(shown);
-
 	pthread_mutex_lock(&state->main_lock);
 
 	if (token_data->client_ref_count > 0)
@@ -954,6 +952,7 @@ static int do_handle_client(struct fsmonitor_daemon_state *state,
 	trace2_data_intmax("fsmonitor", the_repository, "response/count/duplicates", duplicates);
 
 cleanup:
+	kh_destroy_str(shown);
 	strbuf_release(&response_token);
 	strbuf_release(&requested_token_id);
 	strbuf_release(&payload);


ptarjan:
> Really?  Would an interrupted thread even reach its cleanup section?

You’re right - an interrupted thread wouldn’t reach the cleanup section anyway. I’ll drop that claim from the commit message.
The real fix is simply using kh_destroy_str() instead of kh_release_str() to properly free the khash struct. Moving it to the cleanup section is just a minor code hygiene improvement, not a functional fix.
I’ll send a v2 with an updated commit message.

dscho:
> > Really?  Would an interrupted thread even reach its cleanup section?
> 
> You’re right - an interrupted thread wouldn’t reach the cleanup section anyway. I’ll drop that claim from the commit message. The real fix is simply using kh_destroy_str() instead of kh_release_str() to properly free the khash struct. Moving it to the cleanup section is just a minor code hygiene improvement, not a functional fix. I’ll send a v2 with an updated commit message.

@ptarjan please note that you really have to follow the guidance in the "Reply to this" link instead of commenting here on GitHub.

ptarjan:
> > > Really?  Would an interrupted thread even reach its cleanup section?
> 
> > 
> 
> > You’re right - an interrupted thread wouldn’t reach the cleanup section anyway. I’ll drop that claim from the commit message. The real fix is simply using kh_destroy_str() instead of kh_release_str() to properly free the khash struct. Moving it to the cleanup section is just a minor code hygiene improvement, not a functional fix. I’ll send a v2 with an updated commit message.
> 
> 
> 
> @ptarjan please note that you really have to follow the guidance in the "Reply to this" link instead of commenting here on GitHub.

Sorry. I'm new to this procedure. I'll find the text you're talking about and follow it.

ptarjan:
On Tue, Dec 30, 2025 at 10:58 PM gitgitgadget-git[bot]
***@***.***> wrote:
>
> gitgitgadget-git[bot] left a comment (git/git#2148)
>
> On the Git mailing list, René Scharfe wrote (reply to this):
>
> On 12/30/25 1:42 PM, Paul Tarjan via GitGitGadget wrote:
> > From: Paul Tarjan ***@***.***>
> >
> > The do_handle_client() function allocates a khash table to de-duplicate
> > pathnames when responding to client requests. Two issues existed:
> >
> > 1. kh_release_str() was used instead of kh_destroy_str(). The release
> >    function only frees internal arrays (flags, keys, vals) but not the
> >    struct itself (allocated by kh_init_str via xcalloc). This caused a
> >    40-byte leak per request.
> >
> > 2. The khash was freed mid-function rather than in the cleanup section,
> >    so if the worker thread was interrupted before reaching that point
> >    during daemon shutdown, the memory would leak.
> >
> > Fix both issues by:
> > - Initializing shown = NULL at declaration
> > - Using kh_destroy_str() which handles NULL and frees both internal
> >   arrays and the struct itself
> > - Moving the cleanup to the cleanup section so it runs on all exit paths
> >
> > Signed-off-by: Claude ***@***.***>
> > ---
> >     fsmonitor: fix khash memory leak in do_handle_client
> >
> >     The do_handle_client() function allocates a khash table to de-duplicate
> >     pathnames when responding to client requests. Two issues existed:
> >
> >      1. kh_release_str() was used instead of kh_destroy_str(). The release
> >         function only frees internal arrays (flags, keys, vals) but not the
> >         struct itself (allocated by kh_init_str via xcalloc). This caused a
> >         40-byte leak per request.
>
> Makes sense.
>
> >
> >      2. The khash was freed mid-function rather than in the cleanup section,
> >         so if the worker thread was interrupted before reaching that point
> >         during daemon shutdown, the memory would leak.
>
> Really?  Would an interrupted thread even reach its cleanup section?

You’re right - an interrupted thread wouldn’t reach the cleanup
section anyway. I’ll drop that claim from the commit message.
The real fix is simply using kh_destroy_str() instead of
kh_release_str() to properly free the khash struct. Moving it to the
cleanup section is just a minor code hygiene improvement, not a
functional fix.
I’ll send a v2 with an updated commit message.

PR #2194, opened 2026-02-10
rebase -i: introduce `pick -x` for adding a "cherry picked from ..." note

tgross35:


dscho:
@tgross35 are you aware that every time you push, you trigger a new GitHub workflow run? These don't come cheap, and just because somebody else pays for them should not necessarily mean that you should feel eager to spend those minutes. For example, [this run](https://github.com/git/git/actions/runs/28742395764/usage), although canceled because you pushed another update while the workflow was running, spent thirteen _hours_ of compute.

PR #2274, opened 2026-04-19
merge-ort: handle cached rename & trivial resolution interaction better

newren:
Long-standing bug discovered somewhat recently.


newren:
Wrong project; closing so I can reopen against gitgitgadget/git.

PR #2275, opened 2026-04-19
Duplicate entry hardening

newren:
We had some corrupt trees with duplicate entries in real world repositories, which triggered an assertion failure in merge-ort.  Further, the corrupt tree creation in the third party tool would have been avoided had verify_cache() correctly checked for D/F conflicts.  Provide fixes for both issues, including 3 preparatory changes for the merge-ort fix.


newren:
Wrong project; closing so I can reopen against gitgitgadget/git.

PR #2285, opened 2026-04-29
branch: delete-merged

HaraldNordgren:
Delete branches that have already been merged on upstream.

Changes in v25:
- Wrap long lines in tests.

Changes in v24:
- Map branches through the configured push and fetch refspecs before deciding whether pushing would update their upstream.
- Protect only the direct upstream of surviving work, delete deeper merged bases and clear stale upstream configuration.
- Clarify that `--delete-merged` takes upstream patterns in docs.

Changes in v23:
- Use a loop-local `size_t` iterator for `--forked` patterns.
- Simplify upstream-prefix stripping into an explicit fallback.

Changes in v22:
- Loop variable int when appropriate to avoid typecast.

Changes in v21:
- Change loop variables to prevent CI timeout.

Changes in v20:
- Protect branches transitively required by a surviving local upstream stack. Traverse upstream chains once and defer delete-set mutation until traversal completes.
- Make stacked-branch handling independent of ref iteration order and update the documentation accordingly.
- Clarify variable names with regards to branch names (short) to reduce confusion.

Changes in v19:
- Fix bug where dry-run would still remove config, added test coverage.
- Redesigned `--delete-merged` as a repeatable upstream selector with optional positional patterns limiting deletion scope.
- Protect same-name upstream branches independently of push-default configuration.
- Simplified flags handling where local caching became complicated when mutating values.
- Clarified assertions in tests.

Changes in v18:
- Instead of keeping the whole chain of upstream branches, keep only the ones an unmerged branch still needs. When a kept (merged) branch in turn tracks a branch that is being deleted, clear its now-stale upstream config.
- Rework spare_stacked_bases() to record the kept bases and, in a second pass, clear the upstream of any whose own base is going away. Build the to-delete list with strset_for_each_entry() instead of re-walking the candidate array.

Changes in v17:
- Keep a merged branch when another surviving branch still tracks it as its upstream, so `--delete-merged` no longer deletes a branch out from under one stacked on top of it.
- Move the `--dry-run` and `branch.<name>.deleteMerged` opt-out fully into their own commits.

Changes in v16:
- Convert `delete_merged_branches()` to take an `unsigned int flags` argument instead of separate `quiet`/`dry_run` booleans, matching `delete_branches()`
- Reuse the `strbuf` across the skip-config loop (`strbuf_reset` per iteration, single `strbuf_release` after) instead of allocating and freeing it each time
- Rewrite the `--delete-merged` tests as integration tests: branches that land commits upstream, with deletion and the checked-out, upstream-gone, and push-equals-upstream safety cases exercised together in one run and output asserted via `test_cmp`
- Collapse the many per-aspect test repos into a single reused `repo` set up by a `setup_repo_for_delete_merged` helper, and rename helpers off the old `pm_`/`prune` naming
- Nest single-repo setup sequences in `( cd ... )` subshells instead of prefixing every command with `-C`

Changes in v15:
- Renamed --prune-merged to --delete-merged throughout. Not necessarily
  final, but something to advance the discussion.
- --delete-merged now silently skips not-yet-merged branches instead of
  warning.
- Initialized the delete_branches() flag locals where declared. Only force
  stays deferred.
- delete_branches()/check_branch_commit() doc and code cleanups: redundant
  branch NULL checks dropped, ref_array candidates = { 0 }, a BUG() for the
  unreachable non-branch ref, and reworked --delete-merged doc wording.
- Broadened the --forked tests (local commits for realism, remote add -f,
  --forked <pattern> <branch> coverage), renamed the misleading trunk
  fixture, and replaced the misnamed detached branch with git checkout
  --detach.

Changes in v14:
- Fixed a git branch -d -r regression (broke t5404/t5505/t5514): the
  remotes path set a local force but not the DELETE_BRANCH_FORCE bit that
  check_branch_commit() reads, so it wrongly ran the merge check.
- Made flags the single source of truth in delete_branches() so the bit and
  the derived locals can't disagree.
- Works locally, but GitHub CI has problems that are there for other
  branches too, hopefully not related
  (https://github.com/git/git/pull/2285).

Changes in v13:
- Reworked --forked into a real ref-filter applied in apply_ref_filter()
  instead of a post-pass, so non-matching branches are never allocated.
- Match exact --forked patterns on full refnames (only globs use the
  abbreviated upstream), and dropped the old helper machinery, forward
  declaration, and string_list in favor of a strvec.
- Replaced the boolean parameters of
  delete_branches()/check_branch_commit() with a single unsigned int flags.
- --prune-merged now collects candidates via filter_refs() rather than its
  own branch walk.
- --prune-merged now takes its <branch> patterns as positional arguments
  (e.g. git branch --prune-merged origin/main 'feature*') instead of
  repeating the option.

Changes in v12:
- Reworked --forked from a standalone action into a --list-mode filter.
- Switched --forked and --prune-merged to repeatable OPT_STRING_LIST
  options.
- Dropped the bare-remote-name resolution for --forked, the argument is now
  a ref or a glob.

Changes in v11:
- The flags now take a branch, not a remote. --forked and --prune-merged
  accept a literal upstream short name like origin/main or a wildmatch
  pattern like origin/*. The old --all-remotes flag is gone, since origin/*
  covers that case.
- The prune guard now compares @{push} against @{upstream}. A branch is
  spared when these are equal. That is the trunk like case, such as local
  main tracking and pushing to origin/main, where "fully merged to
  upstream" cannot be told apart from "just pulled". Only branches that
  push somewhere other than their upstream, typically fork based topics,
  are candidates. The earlier <remote>/HEAD by name guard that the reviewer
  rejected is gone.
- New --dry-run for --prune-merged.

Changes in v10:
- --forked / --prune-merged now take a branch glob instead of a remote name
  — origin, origin/*, origin/release-- all work. This replaces the
  remote-only form and subsumes the old --all-remotes flag, which has been
  dropped.
- New --dry-run for --prune-merged.

Changes in v9:
- --force no longer has special meaning with --prune-merged; reachability
  is always enforced. Use git branch -D to delete an unmerged branch.
  Matches how git branch's other read/safe actions treat --force.
- Synopsis drops [-f]; "not fully merged" hint points at git branch -D.
- Dropped the --prune-merged --force tests.

Changes in v8:
- Delete only when the branch's work is actually reachable from its
  upstream
- Skip branches whose upstream is gone (even with --force)
- Simplified the internal safety flag to live in one place

Changes in v7:
- --prune-merged now checks if a branch is merged into its own upstream
  first. If the upstream is gone, it checks against the remote's default
  branch instead. If neither exists, the branch is refused (use --force to
  delete anyway).

Changes in v6:
- --prune-merged now measures merged-ness against the remote's default
  branch instead of the candidate's upstream — so the decision no longer
  depends on which branch happens to be checked out locally.
- delete_branches() / check_branch_commit() gained a per-candidate override
  that lets a caller substitute a different "what counts as merged"
  reference (or skip the check). branch -d callers pass NULL and keep their
  existing semantics.
- prune_merged_branches() resolves each candidate's push-remote HEAD and
  threads it through, so --prune-merged --all-remotes measures each
  candidate against its own remote rather than a single global reference.

Changes in v5:
- Drop commit 'fetch: add --prune-merged'

Changes in v4:
- Resolve each remote's HEAD and collect the targets into a
  protected_default_refs set in collect_forked_set.
- In prune_merged_branches, skip a candidate when its upstream is a
  protected default ref and the local branch name matches the default
  branch's leaf name (so a local main tracking origin/main is spared, but a
  renamed trunk tracking origin/main is not).
- Also skip when the candidate's push ref points at a protected default
  ref, so a topic branch configured to push to origin/main is never pruned.
- Tests: spare the local default branch; only protect by matching leaf name
  (not by upstream alone); spare a branch whose push ref is the remote
  default.

Changes in v3:
- s/remote-tracking refs/remote-tracking branches/g

Changes in v2:
- The whole feature moved out of git fetch and into git branch. git fetch
  --prune-merged now just calls git branch --prune-merged after fetching.
- The fetch.pruneLocalBranches and remote.<name>.pruneLocalBranches config
  options are gone, replaced by per-branch opt-out via
  branch.<name>.pruneMerged.
- New git branch --forked <remote> lists local branches whose upstream
  lives on the given remote (read-only building block).
- New git branch --prune-merged <remote> deletes those branches, but only
  if their tip is reachable from the upstream tracking ref; --force skips
  that safety check.
- New git branch --all-remotes lets --forked/--prune-merged operate across
  every configured remote at once.
- The currently checked-out branch in any worktree is always preserved.
- branch.<name>.pruneMerged=false lets you exempt a branch (e.g. a
  long-running topic branch) even with --force; doesn't affect explicit git
  branch -d.
- delete_branches() got a warn_only mode so bulk deletion prints a one-line
  warning per skipped branch instead of the noisy four-line hint that git
  branch -d shows.
- New section in git-branch docs; git-fetch docs trimmed to just mention
  --prune-merged.
- New tests in t3200-branch.sh for the new branch flags; t5510-fetch.sh
  shrunk since most logic moved.

### DIFF
--- a/Documentation/config/branch.adoc
+++ b/Documentation/config/branch.adoc
@@ -103,3 +103,10 @@ for details).
 	`git branch --edit-description`. Branch description is
 	automatically added to the `format-patch` cover letter or
 	`request-pull` summary.
+
+`branch.<name>.deleteMerged`::
+	If set to `false`, branch _<name>_ is exempt from
+	`git branch --delete-merged`.  Useful for a topic branch you
+	intend to develop further after an initial round has been
+	merged upstream.  Defaults to true.  Explicit deletion via
+	`git branch -d` is unaffected.
--- a/Documentation/git-branch.adoc
+++ b/Documentation/git-branch.adoc
@@ -13,6 +13,7 @@ git branch [--color[=<when>] | --no-color] [--show-current]
 	   [--column[=<options>] | --no-column] [--sort=<key>]
 	   [--merged [<commit>]] [--no-merged [<commit>]]
 	   [--contains [<commit>]] [--no-contains [<commit>]]
+	   [(--forked <branch>)...]
 	   [--points-at <object>] [--format=<format>]
 	   [(-r|--remotes) | (-a|--all)]
 	   [--list] [<pattern>...]
@@ -51,7 +52,8 @@ merged into the named commit (i.e. the branches whose tip commits are
 reachable from the named commit) will be listed.  With `--no-merged` only
 branches not merged into the named commit will be listed.  If the _<commit>_
 argument is missing it defaults to `HEAD` (i.e. the tip of the current
-branch).
+branch).  With `--forked`, only branches whose configured upstream matches
+the given branch or pattern will be listed.
 
 The command's second form creates a new branch head named _<branch-name>_
 which points to the current `HEAD`, or _<start-point>_ if given. As a
@@ -310,6 +312,14 @@ superproject's "origin/main", but tracks the submodule's "origin/main".
 `--no-merged [<commit>]`::
 	Only list branches whose tips are not reachable from
 	_<commit>_ (`HEAD` if not specified). Implies `--list`.
+
+`--forked <branch>`::
+	Only list branches whose configured upstream matches
+	_<branch>_. The argument can be a ref (e.g. `origin/main`,
+	`master`), a remote name like `origin` for the branch its
+	`origin/HEAD` points at, or a shell-style glob (e.g.
+	`'origin/*'`). The option can be repeated to widen the
+	filter. Implies `--list`.
 
 `--points-at <object>`::
 	Only list branches of _<object>_.
--- a/Documentation/git-branch.adoc
+++ b/Documentation/git-branch.adoc
@@ -224,7 +224,8 @@ A branch is not deleted when:
   distinguished from a branch that just looks fully merged right
   after a pull; this is determined by the remote's configured push and
   fetch refspecs,
-* it is the local upstream of a branch that is not being deleted.
+* it is the local upstream of a branch that is not being deleted, or
+* `branch.<name>.deleteMerged` is set to `false`.
 --
 +
 When such a local upstream branch has its own upstream deleted by the
--- a/Documentation/git-branch.adoc
+++ b/Documentation/git-branch.adoc
@@ -25,6 +25,7 @@ git branch (-m|-M) [<old-branch>] <new-branch>
 git branch (-c|-C) [<old-branch>] <new-branch>
 git branch (-d|-D) [-r] <branch-name>...
 git branch --edit-description [<branch-name>]
+git branch (--delete-merged <pattern>)... [<branch-pattern>...]
 
 DESCRIPTION
 -----------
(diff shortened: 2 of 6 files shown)


Edgarruiz8585:
-> by: Edgar Manuel Ruiz Arias @Edgarruiz8585 ***@***.***
***@***.*** ***@***.***

chiara5555 (commented):
On 03/06/2026 10:04, Harald Nordgren via GitGitGadget wrote:
> From: Harald Nordgren <haraldnordgren@gmail.com>
> > Add no_head_fallback and dry_run flags to delete_branches() so a
> bulk caller (the upcoming --prune-merged) can ask strictly about
> merged-into-upstream without a silent fallback to HEAD, and
> rehearse deletions with the same "Would delete branch ..." wording
> as the live run. Existing callers pass 0 for both and keep current
> behavior.
> > When no_head_fallback is set, head_rev stays NULL through to
> branch_merged(), whose "merged to X but not yet merged to HEAD"
> reminder otherwise compares against HEAD. For the bulk caller
> every candidate is known to have an upstream, so HEAD is
> irrelevant. Guard the block on head_rev so the NULL case skips
> it instead of treating "NULL != reference_rev" as "diverges from
> HEAD" and emitting a spurious warning.

Same comment as the last patch - use a flags argument rather than lots of individual booleans that make the call sites hard to read.

Thanks

Phillip

mohommedsmohommeds-source:
4796beta 024

HaraldNordgren:
@dscho Good point! Btw, who is paying for the minutes?

HaraldNordgren:
@dscho Can we make it so that a new push on the same PR will cancel previous builds?

HaraldNordgren:
@dscho I feel like your post is needlessly rude. I didn't realize that jobs weren't canceled already, so let's fix it now: https://github.com/git/git/pull/2369

PR #2292, opened 2026-05-07
path.c: translate Windows paths recorded by Windows git on POSIX hosts

johnnyshields:
### The Problem

If I create a git worktree using Windows Git, then git inside WSL Linux cannot resolve its path, because the `gitdir` specifier will be a Windows-style path `C:\repo\...` rather than a `POSIX-style /mnt/c/repo...`

As background, I use a hybrid Windows and WSL Ubuntu setup on the same machine; it works nearly perfectly except for the above issue, specifically with worktrees.

### Related Pull-Requests

The vice-versa issue also affects creating a worktree in WSL and accessing it via Windows Git; I've raised a patch to Windows Git here: https://github.com/git-for-windows/git/pull/6234 but please let me know if I should upstream this to the master git repo.

### Root Cause

Worktrees use a bi-directional link that include absolute paths in the `gitdir` value to point from the main repo to the worktree repo, and vice versa. It is specifically this `gitdir` value that needs Windows-to-POSIX path translation handling. (Nothing else besides this path requires additional logic for worktrees to work across WSL and Windows native.)

```
  <main-repo>/   ← e.g. C:\workspace\myrepo (or /mnt/c/workspace/myrepo from WSL)
  └── .git/
      └── worktrees/
          └── <name>/
              ├── gitdir      ← file. content: "<worktree>/.git\n"
              ├── commondir   ← file (sometimes). content: relative or absolute path back to main .git
              ├── HEAD
              ├── index
              └── config.worktree

  <worktree>/    ← e.g. C:\workspace\myrepo-wt
  └── .git       ← FILE (not a dir). content: "gitdir: <main-repo>/.git/worktrees/<name>\n"
```

When `git worktree add` is run from native Windows, git writes absolute paths into the worktree's `.git` file, into `<commondir>/worktrees/<id>/gitdir`, and (when present) into `<commondir>/commondir`, in `<x>:/...` or `<x>:\...` form. Reading those files back from a non-Windows-native build of git fails because neither form is meaningful on POSIX, so the worktree appears broken even though every byte of it is reachable - the most common scenario being a worktree on a Windows drive opened from inside WSL2 (where the Windows filesystem is mounted at `/mnt/<x>/`) or from Cygwin/MSYS (where it is `/cygdrive/<x>/`).

### The Solution (What this Patch Does)

This patch adds a small helper `translate_windows_path()` that recognises this shape at the start of a path and rewrites it to the POSIX mount form appropriate for the current build (`/cygdrive/<x>/` on Cygwin, `/mnt/<x>/` everywhere else), converting any backslashes in the remainder to forward slashes. Call it at the three places where non-Windows-native git reads a recorded worktree-related path back from disk:

  * `read_gitfile_gently()` - the `gitdir:` line in a worktree's `.git` file.
  * `get_common_dir_noenv()` - the `commondir` file inside a worktree's git directory, which points at the main repo.
  * `get_linked_worktree()` - the `gitdir` file inside `<commondir>/worktrees/<id>/`, which points at the worktree's `.git` link.

Translation only happens for `<x>:/` or `<x>:\` where `<x>` is a single ASCII letter; anything else is left alone. The helper is a no-op on `GIT_WINDOWS_NATIVE` builds, where the input is already in native form. On non-WSL Linux hosts the translation still produces a syntactically valid POSIX path; if the corresponding `/mnt/<x>/` mount does not exist, the next stat()/open() fails as it would have without translation - i.e. the change cannot make a working configuration stop working.

### Why this is Safe

This PR is safe because:
- The logic is **POSIX-only:** the helper compiles to a no-op if `GIT_WINDOWS_NATIVE`, so a `/mnt/c/...` path on a Linux host - where it might really exist - is still treated literally.
- The logic is **read-only:** it only affects how Windows-style paths are read from POSIX; it does NOT mutate how the paths are stored in .git's definitions.
- **Path traversal safety:** the path translation only happens for the left-anchored part of the string `<x>:\`  where `<x>` is a single ASCII letter (e.g. `C:\`). Anything else (e.g. `http://`) is left alone. The helper also doesn't evaluate/collapse relative paths, so it doesn't introduce any path-traversal vectors that aren't existing today.

### DIFF
--- a/path.c
+++ b/path.c
@@ -18,6 +18,49 @@
 #include "lockfile.h"
 #include "exec-cmd.h"
 
+int translate_windows_path(struct strbuf *path)
+{
+#ifndef GIT_WINDOWS_NATIVE
+#ifdef __CYGWIN__
+	static const char drive_prefix[] = "/cygdrive/";
+#else
+	static const char drive_prefix[] = "/mnt/";
+#endif
+	const size_t drive_prefix_len = sizeof(drive_prefix) - 1;
+	const size_t expansion = drive_prefix_len + 1 - 2;
+	char drive;
+	size_t i;
+
+	if (path->len < 3)
+		return 0;
+	if (!isalpha((unsigned char)path->buf[0]))
+		return 0;
+	if (path->buf[1] != ':')
+		return 0;
+	if (path->buf[2] != '/' && path->buf[2] != '\\')
+		return 0;
+
+	drive = tolower((unsigned char)path->buf[0]);
+
+	/* Rewrite "<letter>:" as "<drive_prefix><drive>", then convert any
+	 * backslashes in the remaining path to forward slashes. */
+	strbuf_grow(path, expansion);
+	memmove(path->buf + 2 + expansion, path->buf + 2, path->len - 2 + 1);
+	memcpy(path->buf, drive_prefix, drive_prefix_len);
+	path->buf[drive_prefix_len] = drive;
+	path->len += expansion;
+
+	for (i = drive_prefix_len + 1; i < path->len; i++) {
+		if (path->buf[i] == '\\')
+			path->buf[i] = '/';
+	}
+	return 1;
+#else
+	(void)path;
+	return 0;
+#endif
+}
+
 static int get_st_mode_bits(const char *path, int *mode)
 {
 	struct stat st;
--- a/path.c
+++ b/path.c
@@ -20,8 +20,57 @@
 
 int translate_windows_path(struct strbuf *path)
 {
-#ifndef GIT_WINDOWS_NATIVE
-#ifdef __CYGWIN__
+#ifdef GIT_WINDOWS_NATIVE
+	/*
+	 * Windows-native: rewrite POSIX-mount paths (WSL `/mnt/<x>/`,
+	 * Cygwin `/cygdrive/<x>/`, MSYS `/<x>/`) to drive form `<x>:/...`.
+	 * The single-letter + separator check below is what stops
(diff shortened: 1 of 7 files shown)


dscho:
@johnnyshields I do not want to be a gatekeeper, so that's why I issued that `/allow`.

But I'm still quite wary of the proposed design. It can do too much harm by voluntarily giving up on a clean design and introducing ambiguities where there were none before. It also makes way too many assumptions for my liking about what the users' setup might be.

For example, the code that you introduced would be also in effect on macOS where there is no WSL. It would be in effect on a Raspberry Pi, where there is no Windows.

The proposed solution is too narrowly focused on solving a particular problem, without considering the undesirable fall-out of that solution. Essentially, it reminds me of [this guidance](https://www.schneier.com/crypto-gram/archives/2002/0415.html#1) how to think about problems (it's framed for security, but applies to a lot more than that, really):

1. What problem are you trying to solve?
2. How well does the proposed design solve the problem?
3. What other problems does the proposed design _cause_?
4. What are the costs of the proposed solution (including the incurred, potentially unintended _other_ problems of step 3)?
5. Given the answers to steps two through four, is the proposed design worth the costs?

In this instance, my answer to the question in step five is squarely no. No, it is not worth the fallout. Honestly, I would much rather introduce support for config settings like `path.<path>.insteadOf` (obviously inspired by [`url.<url>.insteadOf`)](https://git-scm.com/docs/git-config#Documentation/git-config.txt-urlbaseinsteadOf)).

But again, I do not want to be a gatekeeper, so feel free to submit this patch series to the Git mailing list anyways.

PR #2311, opened 2026-05-23
completion: hide dotfiles for selected path completion

alibaba0010:
The completion helper for index paths uses `git ls-files` rather than shell filename completion. As a result, leading-dot paths such as a tracked `.gitignore` were offered even when the user had not started the path with `.`.

Hide leading-dot path components for `git rm`, `git mv`, and `git ls-files` when completing an empty path component. Explicit dot completion is still preserved, so `git rm .` can still complete `.gitignore`.

This removes the existing TODO expectations in `t/t9902-completion.sh` and adds coverage for explicit dot completion.

Validation:
- `git diff --check -- contrib/completion/git-completion.bash t/t9902-completion.sh`
- `bash -n contrib/completion/git-completion.bash`
- `./t9902-completion.sh`

cc: Zakariyah Ali <zakariyahali100@gmail.com>
cc: "D. Ben Knoble" <ben.knoble@gmail.com>


alibaba0010:
The failures from the github actions seems to be from the next branch

alibaba0010:
Thanks for the review. I'll examine all callers of _git_complete_index_file to better understand the current behavior and evaluate whether a more explicit or consistent API would make sense before preparing the next revision.

dscho:
> Thanks for the review. I'll examine all callers of _git_complete_index_file to better understand the current behavior and evaluate whether a more explicit or consistent API would make sense before preparing the next revision.

You will need to follow the guidance in the "Reply to this" link to ensure that your answer actually finds the intended recipient(s). PR comments here are _not_ mirrored to the Git mailing list.

dscho:
@alibaba0010 ☝️

alibaba0010:
Thank you for the guidance @dscho, i would follow suit.

Edgarruiz8585:
> The completion helper for index paths uses `git ls-files` rather than shell filename completion. As a result, leading-dot paths such as a tracked `.gitignore` were offered even when the user had not started the path with `.`.
> 
> Hide leading-dot path components for `git rm`, `git mv`, and `git ls-files` when completing an empty path component. Explicit dot completion is still preserved, so `git rm .` can still complete `.gitignore`.
> 
> This removes the existing TODO expectations in `t/t9902-completion.sh` and adds coverage for explicit dot completion.
> 
> Validation:
> - `git diff --check -- contrib/completion/git-completion.bash t/t9902-completion.sh`
> - `bash -n contrib/completion/git-completion.bash`
> - `./t9902-completion.sh`
> Edgar Manuel Ruiz Arias <ruizariasedgarmanuel@gmail.com>
> cc: Zakariyah Ali <zakariyahali100@gmail.com>

contrib/completion/git-completion.bash
Edgarruiz8585, line 1: + > [ruizariasedgarmanuel@gmail.com](https://github.com/git/git/pull/2311)

Edgarruiz8585 (commented):
+ > ruizariasedgarmanuel@gmail.com

PR #2312, opened 2026-05-26
environment.c: migrate 'trust_executable_bit' into 'repo_config_values'

malon7782:
The 'core.filemode' configuration, which is stored as a global variable 'trust_executable_bit', is a core filesystem capability flag.

Move it into 'repo_config_values' to tie it to the specific repository instance it was read from. Eager parsing is maintained because this flag is heavily consulted in hot paths.

To avoid falling back to 'the_repository', refactor the signatures of helper functions:

 - ce_mode_from_stat()
 - st_mode_from_ce()
 - fake_lstat()
 - check_removed()

These functions now accept a 'struct index_state *istate', ensuring the correct context is seamlessly passed down to the lowest levels.

- Patch 1: Cleans up redundant local 'extern' declarations for 'trust_executable_bit' and 'has_symlinks' in 'read-cache.c'.

- Patch 2: Moves 'ce_mode_from_stat()' from 'read-cache.h' to 'read-cache.c', removing its inline status.

- Patch 3: Refactors the signatures of low-level helpers.

- Patch 4: Performs the actual migration of 'trust_executable_bit' into 'repo_config_values'.

Previous related work: https://lore.kernel.org/git/837b5360b40f992351f489a0ae05fedf49884c6e.1685716420.git.gitgitgadget@gmail.com/

Mentored-by: Christian Couder <christian.couder@gamil.com>
Mentored-by: Ayush Chandekar <ayu.chandekar@gmail.com>
Signed-off-by: Tian Yuchen <cat@malon.dev>

### DIFF
--- a/read-cache-ll.h
+++ b/read-cache-ll.h
@@ -442,7 +442,7 @@ void fill_stat_cache_info(struct index_state *istate, struct cache_entry *ce, st
  * for lstat() for a tracked path that is known to be up-to-date via
  * some out-of-line means (like fsmonitor).
  */
-int fake_lstat(const struct cache_entry *ce, struct stat *st);
+int fake_lstat(struct index_state *istate, const struct cache_entry *ce, struct stat *st);
 
 #define REFRESH_REALLY                   (1 << 0) /* ignore_valid */
 #define REFRESH_UNMERGED                 (1 << 1) /* allow unmerged */
--- a/read-cache.c
+++ b/read-cache.c
@@ -204,7 +204,6 @@ void fill_stat_cache_info(struct index_state *istate, struct cache_entry *ce, st
 
 static unsigned int st_mode_from_ce(const struct cache_entry *ce)
 {
-	extern int trust_executable_bit, has_symlinks;
 
 	switch (ce->ce_mode & S_IFMT) {
 	case S_IFLNK:
--- a/read-cache.c
+++ b/read-cache.c
@@ -202,9 +202,22 @@ void fill_stat_cache_info(struct index_state *istate, struct cache_entry *ce, st
 	}
 }
 
+unsigned int ce_mode_from_stat(const struct cache_entry *ce,
+				unsigned int mode)
+{
+	if (!has_symlinks && S_ISREG(mode) &&
+	    ce && S_ISLNK(ce->ce_mode))
+		return ce->ce_mode;
+	if (!trust_executable_bit && S_ISREG(mode)) {
+		if (ce && S_ISREG(ce->ce_mode))
+			return ce->ce_mode;
+		return create_ce_mode(0666);
+	}
+	return create_ce_mode(mode);
+}
+
 static unsigned int st_mode_from_ce(const struct cache_entry *ce)
 {
-
 	switch (ce->ce_mode & S_IFMT) {
 	case S_IFLNK:
 		return has_symlinks ? S_IFLNK : (S_IFREG | 0644);
--- a/read-cache.c
+++ b/read-cache.c
@@ -206,10 +206,13 @@ unsigned int ce_mode_from_stat(struct index_state *istate,
 				const struct cache_entry *ce,
 				unsigned int mode)
 {
+	struct repository *repo = (istate && istate->repo) ? istate->repo : the_repository;
+	struct repo_config_values *cfg = repo_config_values(repo);
+
 	if (!has_symlinks && S_ISREG(mode) &&
 	    ce && S_ISLNK(ce->ce_mode))
 		return ce->ce_mode;
-	if (!trust_executable_bit && S_ISREG(mode)) {
+	if (!cfg->trust_executable_bit && S_ISREG(mode)) {
 		if (ce && S_ISREG(ce->ce_mode))
 			return ce->ce_mode;
 		return create_ce_mode(0666);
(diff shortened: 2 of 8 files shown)


Anjalo133 (commented):
T

read-cache.c
```diff
@@ -202,28 +202,50 @@ void fill_stat_cache_info(struct index_state *istate, struct cache_entry *ce, st
...
 	case S_IFREG:
-		return (ce->ce_mode & (trust_executable_bit ? 0755 : 0644)) | S_IFREG;
+		return (ce->ce_mode & (cfg->trust_executable_bit ? 0755 : 0644)) | S_IFREG;
 	case S_IFGITLINK:
 		return S_IFDIR | 0755;
 	case S_IFDIR:
 		return ce->ce_mode;
 	default:
-		BUG("unsupported ce_mode: %o", ce->ce_mode);
+		bug("unsupported ce_mode: %o", ce->ce_mode);
```
chriscool: Why is BUG() changed into bug()?

read-cache-ll.h
```diff
@@ -20,6 +20,7 @@ struct cache_header {
 #define INDEX_FORMAT_UB 4
 
 struct cache_entry {
+
```
chriscool: Why is this blank line added?

read-cache.c
```diff
@@ -202,28 +202,50 @@ void fill_stat_cache_info(struct index_state *istate, struct cache_entry *ce, st
...
 
-static unsigned int st_mode_from_ce(const struct cache_entry *ce)
+unsigned int ce_mode_from_stat(struct index_state *istate,
+				const struct cache_entry *ce,
+				unsigned int mode)
+{
+	struct repository *repo = (istate && istate->repo) ? istate->repo : the_repository;
+	struct repo_config_values *cfg = repo_config_values(repo);
+
+	extern int has_symlinks;
```
chriscool: "read-cache.c" already includes "environment.h" which has `extern int has_symlinks;` so I am not sure this is useful here. I think it could be dropped.

read-cache.c
```diff
@@ -202,28 +202,50 @@ void fill_stat_cache_info(struct index_state *istate, struct cache_entry *ce, st
...
+			return ce->ce_mode;
+		return create_ce_mode(0666);
+	}
+	return create_ce_mode(mode);
+}
+
+static unsigned int st_mode_from_ce(struct index_state *istate, const struct cache_entry *ce)
 {
-	extern int trust_executable_bit, has_symlinks;
+	extern int has_symlinks;
```
chriscool: "read-cache.c" already includes "environment.h" which has `extern int has_symlinks;` so I am not sure this is useful here. I think it could be dropped.

chriscool:
It seems to me that the single commit in this branch is doing too many things:

1. Moving `ce_mode_from_stat()` from "read-cache.h" to "read-cache.c" making it non inline.
2. Moving `trust_executable_bit` from a global into `struct repo_config_values`.
3. Passing a `struct index_state *istate` to many functions.

I think the commit could be split into three commits to help with reviews.

Also the usefulness of 3. is dubious for now because repo_config_values() still contains:

```
	if (repo != the_repository)
		BUG("trying to read config from wrong repository instance");
```

So 3. makes it look like some functions could work for a repo that isn't the_repository, but it's in fact not true. At least 3. should be explained more thoroughly in a dedicated commit message.

It looks like there could even be a small cleanup patch before the main three patches to remove the:

`extern int trust_executable_bit, has_symlinks;`

which was not necessary since "read-cache.c" already does `#include "environment.h"`.

read-cache.h
```diff
@@ -4,21 +4,12 @@
 #include "read-cache-ll.h"
 #include "object.h"
 #include "pathspec.h"
+#include "repository.h"
+#include "environment.h"
```
chriscool: Not sure that including both is necessary.

chriscool:
Also about the commit message, maybe you can start to systematically add `Mentored-by: ...` trailers for both of your co-mentors. I think it will simplify things to add them systematically rather than keeping track about who helped you on which commit.

read-cache.h
chriscool: In fact it seems to compile correctly without them.

read-cache.c
```diff
@@ -751,7 +779,7 @@ int add_to_index(struct index_state *istate, const char *path, struct stat *st,
 		int pos = index_name_pos_also_unmerged(istate, path, namelen);
 
 		ent = (0 <= pos) ? istate->cache[pos] : NULL;
-		ce->ce_mode = ce_mode_from_stat(ent, st_mode);
+		ce->ce_mode = ce_mode_from_stat(istate->repo->index, ent, st_mode);
```
chriscool: It looks like this could simply be:

ce->ce_mode = ce_mode_from_stat(istate, ent, st_mode);

chriscool:
The nice thing with a pull request (or merge request) is that its description can contain what will be in the cover letter of the patch series.

In the cover letter it's nice to talk about previous work related to the series.

About the `trust_executable_bit`, there was this patch before:

https://lore.kernel.org/git/20260301190017.53539-1-dronarajgyawali@gmail.com/

and it looks like you participated in the discussion.

This one is maybe relevant too:

https://lore.kernel.org/git/837b5360b40f992351f489a0ae05fedf49884c6e.1685716420.git.gitgitgadget@gmail.com/

So you might want to mention them in the cover letter/this PR's description.

read-cache.c
```diff
@@ -205,7 +204,6 @@ void fill_stat_cache_info(struct index_state *istate, struct cache_entry *ce, st
 
 static unsigned int st_mode_from_ce(const struct cache_entry *ce)
 {
-	extern int trust_executable_bit, has_symlinks;
```
chriscool: I think the blank line following the `extern int ...` line could be removed too.

PR #2315, opened 2026-05-28
doc: add missing --message long option to merge docs

brandondong:


dscho:
> See builtin\merge.c. New doc line is formatted in the same way as
> git commit.

@brandondong this commit message is likely to elicit rather adverse reviews on the Git mailing list. You will probably want to follow the guidance in https://github.blog/2022-06-30-write-better-commits-build-better-projects/ to improve it, in particular with a strong focus on this part:

-- | -- | --

PR #2316, opened 2026-05-28
environment: migrate trust_executable_bit (test)

malon7782:
> This commit series is currently for test use. Will refine associated descriptions once completed.
  > Based on @chriscool's  suggestion, this branch is a reordered version of #2312

The 'core.filemode' configuration, which is stored as a global variable 'trust_executable_bit', is a core filesystem capability flag.

Move it into 'repo_config_values' to tie it to the specific repository instance it was read from. Eager parsing is maintained because this flag is heavily consulted in hot paths.

To avoid falling back to 'the_repository', refactor the signatures of helper functions:

 - ce_mode_from_stat()
 - st_mode_from_ce()
 - fake_lstat()
 - check_removed()

These functions now accept a 'struct index_state *istate', ensuring the correct context is seamlessly passed down to the lowest levels.

Note: 'repo_config_values()' still does not support any 'struct repository' other than 'the_repository'. In other words, this series of patches is laying the groundwork for the eventual elimination of 'the_repository'.

Previous related work:   
 - [PATCH 2/6] config: add trust_executable_bit to global config
 - [PATCH] Refactor 'trust_executable_bit' to repository-scoped setting (This series of patches was unsuccessful because the target location selected was 'struct repo_settings', which our analysis indicated was not the optimal choice. For further details, please see: [1])

[1] https://lore.kernel.org/git/837b5360b40f992351f489a0ae05fedf49884c6e.1685716420.git.gitgitgadget@gmail.com/
[2] https://lore.kernel.org/git/20260301190017.53539-1-dronarajgyawali@gmail.com/
[3] https://lore.kernel.org/git/xmqq1pht6nyx.fsf@gitster.g/

Mentored-by: Christian Couder <christian.couder@gmail.com>
Mentored-by: Ayush Chandekar <ayu.chandekar@gmail.com>
Mentored-by: Olamide Caleb Bello <belkid98@gmail.com>
Signed-off-by: Tian Yuchen <cat@malon.dev>

### DIFF
--- a/read-cache.c
+++ b/read-cache.c
@@ -204,10 +204,12 @@ void fill_stat_cache_info(struct index_state *istate, struct cache_entry *ce, st
 
 unsigned int ce_mode_from_stat(const struct cache_entry *ce, unsigned int mode)
 {
+	struct repo_config_values *cfg = repo_config_values(the_repository);
+
 	if (!has_symlinks && S_ISREG(mode) &&
 	    ce && S_ISLNK(ce->ce_mode))
 		return ce->ce_mode;
-	if (!trust_executable_bit && S_ISREG(mode)) {
+	if (!cfg->trust_executable_bit && S_ISREG(mode)) {
 		if (ce && S_ISREG(ce->ce_mode))
 			return ce->ce_mode;
 		return create_ce_mode(0666);
@@ -217,11 +219,13 @@ unsigned int ce_mode_from_stat(const struct cache_entry *ce, unsigned int mode)
 
 static unsigned int st_mode_from_ce(const struct cache_entry *ce)
 {
+	struct repo_config_values *cfg = repo_config_values(the_repository);
+
 	switch (ce->ce_mode & S_IFMT) {
 	case S_IFLNK:
 		return has_symlinks ? S_IFLNK : (S_IFREG | 0644);
 	case S_IFREG:
-		return (ce->ce_mode & (trust_executable_bit ? 0755 : 0644)) | S_IFREG;
+		return (ce->ce_mode & (cfg->trust_executable_bit ? 0755 : 0644)) | S_IFREG;
 	case S_IFGITLINK:
 		return S_IFDIR | 0755;
 	case S_IFDIR:
@@ -321,6 +325,7 @@ static int ce_modified_check_fs(struct index_state *istate,
 static int ce_match_stat_basic(const struct cache_entry *ce, struct stat *st)
 {
 	unsigned int changed = 0;
+	struct repo_config_values *cfg = repo_config_values(the_repository);
 
 	if (ce->ce_flags & CE_REMOVE)
 		return MODE_CHANGED | DATA_CHANGED | TYPE_CHANGED;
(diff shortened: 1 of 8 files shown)


chriscool:
It looks like in the description and commit messages `gamil.com` is used instead of `gmail.com` in my email address (`christian.couder@gamil.com` instead of `christian.couder@gmail.com`).

chriscool:
The commit title `read-cache: move 'ce_move_from_stat()' to 'read-cache.c'` has a typo too. It should talk about `ce_mode_from_stat()` not `ce_move_from_stat()`.

read-cache.c
```diff
@@ -205,7 +204,6 @@ void fill_stat_cache_info(struct index_state *istate, struct cache_entry *ce, st
 
 static unsigned int st_mode_from_ce(const struct cache_entry *ce)
 {
-	extern int trust_executable_bit, has_symlinks;
```
chriscool, line 211: The blank line after the  `extern int trust_executable_bit, has_symlinks;` line should be removed too, so that the body of the function doesn't start with a blank line.

read-cache.h
```diff
@@ -5,20 +5,8 @@
 #include "object.h"
 #include "pathspec.h"
 
-static inline unsigned int ce_mode_from_stat(const struct cache_entry *ce,
-					     unsigned int mode)
-{
-	extern int trust_executable_bit, has_symlinks;
```
chriscool, line 11: The commit message should probably say that when moving the function to "read-cache.c", the `extern int trust_executable_bit, has_symlinks;` line is discarded because it's not necessary when the function lives in "read-cache.c".

read-cache.c
```diff
@@ -208,7 +208,7 @@ unsigned int ce_mode_from_stat(const struct cache_entry *ce,
 	if (!has_symlinks && S_ISREG(mode) &&
 	    ce && S_ISLNK(ce->ce_mode))
 		return ce->ce_mode;
-	if (!trust_executable_bit && S_ISREG(mode)) {
+	if (!repo_config_values(the_repository)->trust_executable_bit && S_ISREG(mode)) {
```
chriscool: In the next commit, this line is changed again to use `cfg->trust_executable_bit` instead of `repo_config_values(the_repository)->trust_executable_bit`.

I think it would be simpler to use `cfg->trust_executable_bit` already in this commit.

chriscool: Same for the other places in this commit where `repo_config_values(the_repository)` is used and then replaced with `cfg` in the next commit.

chriscool:
It seems to me that the goal of the "read-cache: pass 'istate' to stat/mode helper functions" is to help with getting rid of `the_repository` in the long term, but it is not mentioned clearly in its commit message.

I think saying something like "This is a step in the long term direction of getting rid of `the_repository` variable." could help justify the commit. 

In the commit message you might also want to take into account what I said in this comment:

https://github.com/git/git/pull/2312#issuecomment-4552192032

chriscool:
I am not sure why the description doesn't also mention:

https://lore.kernel.org/git/20260301190017.53539-1-dronarajgyawali@gmail.com/

I also think it would be nice if there was a small explanation of what the previous works on this tried and why they weren't successful.

malon7782 (commented):
> Note that the 'extern int trust_executable_bit, has_symlinks;' line is
> discarded because it's not necessary when the function lives in
> "read-cache.c".

Specified the wrong range when replacing quotes via Vim, which has resulted in inconsistent quote types in the filenames here. 

"read-cache.c" vs 'read-cache.h'

How did I miss it? :((

PR #2317, opened 2026-05-29
worktree: copy-on-write creation and shared-branch worktrees

nevion:
When many worktrees share one repository -- e .g. a fleet of agents each
needing an isolated checkout -- "git worktree add" is costly at scale.
Objects are shared via the common dir, but the working tree is not: each
add rewrites every tracked file, so N worktrees cost N full checkouts of
disk and I/O. And a branch can only be checked out in one worktree.

Patch 1 adds "git worktree add --reflink": on a copy-on-write filesystem
it populates the new worktree by reflinking the current worktree's files
and index, then "git reset --hard" rewrites only the paths that differ
from <commit-ish>. A reflink_file() helper in copy.c uses FICLONE (Linux)
and clonefile() (macOS); elsewhere (other filesystems, Windows) it is
probed up front and falls back to a normal checkout. Defaulting is via
the worktree.reflink config (true/false/auto); --no-reflink overrides.

Patch 2 lets a branch be checked out in several worktrees, for parallel
work on one checkout. A branch mid-rebase or mid-bisect elsewhere is
still refused.

Benchmark (Linux-kernel fork, 93k files, ~33 GB tree incl. build output,
btrfs): a normal add allocates ~0.9 GB of real disk per worktree (~5.3 GB
for four, linear); --reflink allocates ~0 at any count and also carries
the untracked build tree. ("Real disk" = btrfs exclusive bytes.)

worktree-reflink-bench

Note: patch 2 changes a default (same-branch checkout now allowed); two
t2400 assertions were updated accordingly.

cc: "brian m. carlson" <sandals@crustytoothpaste.net>

### DIFF
--- a/Documentation/config/worktree.adoc
+++ b/Documentation/config/worktree.adoc
@@ -17,3 +17,13 @@
 Note that setting `worktree.useRelativePaths` to "`true`" implies enabling the
 `extensions.relativeWorktrees` config (see linkgit:git-config[1]),
 thus making it incompatible with older versions of Git.
+
+`worktree.reflink`::
+	Controls whether `git worktree add` populates new worktrees with
+	copy-on-write (reflink) clones, as if `--reflink` had been given
+	(see linkgit:git-worktree[1]). May be set to "`true`", "`false`"
+	(the default), or "`auto`". With "`true`", a filesystem that does
+	not support reflinks produces a warning before falling back to an
+	ordinary checkout; with "`auto`", the fallback is silent. An
+	explicit `--reflink` or `--no-reflink` on the command line
+	overrides this setting.
--- a/Documentation/git-worktree.adoc
+++ b/Documentation/git-worktree.adoc
@@ -93,8 +93,9 @@ then, as a convenience, the new worktree is associated with a branch (call
 it _<branch>_) named after `$(basename <path>)`.  If _<branch>_ doesn't
 exist, a new branch based on `HEAD` is automatically created as if
 `-b <branch>` was given.  If _<branch>_ does exist, it will be checked out
-in the new worktree, if it's not checked out anywhere else, otherwise the
-command will refuse to create the worktree (unless `--force` is used).
+in the new worktree, even if it is already checked out in another worktree.
+(A branch that another worktree is in the middle of rebasing or bisecting is
+refused unless `--force` is used.)
 +
 If _<commit-ish>_ is omitted, neither `--detach`, or `--orphan` is
 used, and there are no valid local branches (or remote branches if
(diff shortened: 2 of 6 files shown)


nevion:
this feature is used/reimplemented within grok build https://x.com/theskory/status/2059729539287167068 (I have nothing to do with that) but the best place for this to be done is inside git so all tools can benefit.

nevion:
Benchmark — Linux-kernel fork (93k tracked files, ~33 GB working tree incl. build artifacts, btrfs):

worktree --reflink benchmark

Left: real disk written vs. number of worktrees (normal grows ~1.3 GB each; `--reflink` stays at 0). Right: one worktree — ~0 new disk, and it carries the full build tree (464k files) vs. source-only (93k). "Real disk" = btrfs exclusive bytes; shared CoW extents count as 0.

nevion:
The non-Linux CI failures here are unrelated to this change (which only touches `worktree`, `copy`, and `t2400`):

- **`win test (4)` — `t3070.966 iwildmatch (via ls-files)`**: this is in wildmatch/pathspec matching, untouched by this series. It is already failing on git/git's own `master` CI, independent of this PR — e.g. run [#11454](https://github.com/git/git/actions/runs/26709491999) (`1666c126`) and run [#11420](https://github.com/git/git/actions/runs/26496083241) (`c69baaf5`, "The 9th batch" — the very commit this branch is based on) both show `win test (4) = failure`, whereas it still passed at run #11401 (`56a4f3c3`). So it's a recent base regression on the Windows shard, not introduced here; `t3070-wildmatch.sh` and `wildmatch.c` are byte-identical to v2.54.0, so rebasing onto a release would not avoid it. It'll be resolved on `master` independently of this topic.

- **`osx-gcc` / `osx-meson` (macos-14)**: intermittent. These *passed* on the first CI run for this exact commit and failed only on a re-run (failing step: `ci/run-build-and-tests.sh`, no specific test reported). This is macOS-runner flakiness; a re-run clears it.

- **`linux-TEST-vars (ubuntu:20.04)`** (now green): the earlier failure was a runner infrastructure error — `System.IO.IOException: No space left on device` — not a test failure. The `linux32`/`osx` "cancelled" entries were fail-fast cancellations from that run. A re-run brought all `linux-*` jobs green.

Every job that actually exercises this change (all `linux-*`, almalinux, debian, fedora, and `win test` 0-3/5-9) passes.

dscho:
Yes, sadly Git's CI is not always in a pristine shape :-(

nevion:
@dscho do I do anything else?

dscho:
Why, [yes](https://github.com/git/git/pull/2317#issuecomment-4575005983), you `/submit` (or maybe first `/preview`).

PR #2319, opened 2026-06-03
ci: fix expression injection pattern in l10n workflow

XananasX7:
The `l10n.yml` workflow injects GitHub Actions expressions directly into shell `run` steps:

```yaml
base=${{ github.event.pull_request.base.sha }}
head=${{ github.event.pull_request.head.sha }}
base=${{ github.event.before }}
head=${{ github.event.after }}
```

While these values happen to be hex SHAs, the pattern is flagged by security scanners (zizmor, actionlint) and violates GitHub's security hardening best practices.

Fix: move expressions into env vars and reference env vars in the shell script.

Ref: https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections


XananasX7:
Thanks for the review. The git-po-helper warning about author vs committer email mismatch — I'll fix the commit authorship so both author and committer use the same email. The l10n workflow fix itself (moving `${{ github.event.pull_request.number }}` to an env var) is the right approach per GitHub's security hardening guide. I'll amend and push a clean commit.

dscho:
@XananasX7 you're supposed to squash the fixups before submitting the patch. Also, you will want to avoid being ignored due to a commit message that is vastly different from existing ones. I suggest getting inspired by https://github.blog/2022-06-30-write-better-commits-build-better-projects/ to improve it, in particular with a strong focus on this part:

-- | -- | --

XananasX7:
Thanks for the detailed feedback @dscho — addressed everything:

1. **Squashed the fixup commits** — the branch now has a single clean commit on top of master, no more `ci: fix committer identity` fixup floating on top.

2. **Commit message** — rewrote it in the style used for recent l10n workflow changes (see `l10n: bump mshick/add-pr-comment from v2 to v3`). Subject uses the `l10n:` prefix, body explains the *why* (the injection risk, why the SHA fields are currently safe but the pattern is still worth hardening), and ends with `Signed-off-by`.

3. **Author/committer identity** — both are now `El Mehdi Abenhazou <mehdiananas007@gmail.com>` — no more mismatch.

The change itself is the same: the four context values (`pull_request.base.sha`, `pull_request.head.sha`, `before`, `after`) are lifted from inline shell interpolation into an `env:` block on the same step. Happy to adjust the commit message wording further if anything still feels off.

XananasX7:
Fixed the `git-po-helper` CI failure — the commit message had a URL on line 32 that was 137 chars, exceeding the 72-char limit. I've wrapped the footnote reference across multiple lines:

```
[1] https://docs.github.com/en/actions/security-guides/
    security-hardening-for-github-actions
    #understanding-the-risk-of-script-injections
```

All body lines are now ≤72 chars. Force-pushed to fix.

PR #2320, opened 2026-06-04
Fix TAP output

pks-gitlab:
Test PR to check what breaks.


pks-gitlab:
Merged to next via b5a4cd26ee (Merge branch 'ps/t7527-fix-tap-output' into next, 2026-06-11). Closing.

PR #2323, opened 2026-06-06
rebase: add --fixup to fold a range into its oldest commit

HaraldNordgren:
Adds `git rebase --autosquash --fixup [<upstream>]` to fold a range of commits into its oldest one, reusing that commit's message.


HaraldNordgren:
Superseded by #2337 (same change, renamed branch).

PR #2330, opened 2026-06-11
rebase: mention --abort when an exec step fails

HaraldNordgren:
Mention `git rebase --abort` both in the warning shown when an `exec` step fails and in the `git status` advice while splitting or editing a commit, since rebase pointed users at `--continue` there without saying how to bail out, unlike every comparable command.


sequencer.c
amin404ala, line 1: 4

PR #2332, opened 2026-06-11
clone: accept DEPTH env var as fallback for --depth

h8d13:
This is useful for CI/CD where the clone might clone something else if that makes sense.


h8d13:
Beep beep boop boop I signed with gpg :)

PR #2333, opened 2026-06-11
clone: accept `GIT_CLONE_DEPTH` env var as fallback for `--depth`

h8d13:
cc: "Matt Hunter" <m@lfurio.us>

### DIFF
--- a/builtin/clone.c
+++ b/builtin/clone.c
@@ -898,6 +898,7 @@ int cmd_clone(int argc,
 	int option_reject_shallow = -1; /* unspecified */
 	int deepen = 0;
 	char *option_template = NULL, *option_depth = NULL, *option_since = NULL;
+	char *env_depth = NULL;
 	char *option_origin = NULL;
 	struct string_list option_not = STRING_LIST_INIT_NODUP;
 	const char *real_git_dir = NULL;
@@ -1021,6 +1022,12 @@ int cmd_clone(int argc,
 	if (argc == 0)
 		usage_msg_opt(_("You must specify a repository to clone."),
 			builtin_clone_usage, builtin_clone_options);
+
+	if (!option_depth) {
+		const char *value = getenv("GIT_CLONE_DEPTH");
+		if (value && *value)
+			option_depth = env_depth = xstrdup(value);
+	}
 
 	if (option_depth || option_since || option_not.nr)
 		deepen = 1;
@@ -1639,6 +1646,7 @@ int cmd_clone(int argc,
 	string_list_clear(&server_options, 0);
 
 	free(remote_name);
+	free(env_depth);
 	strbuf_release(&reflog_msg);
 	strbuf_release(&branch_top);
 	strbuf_release(&buf);
--- a/Documentation/git-clone.adoc
+++ b/Documentation/git-clone.adoc
@@ -278,6 +278,10 @@ corresponding `--mirror` and `--no-tags` options instead.
 	`--no-single-branch` is given to fetch the histories near the
 	tips of all branches. If you want to clone submodules shallowly,
 	also pass `--shallow-submodules`.
+	When `--depth` is not given on the command line, the value of the
+	`GIT_CLONE_DEPTH` environment variable is used instead. This makes
+	nested clones spawned by wrapper tools shallow without threading
+	`--depth` through each invocation.
 
 `--shallow-since=<date>`::
 	Create a shallow clone with a history after the specified time.
--- a/t/t5601-clone.sh
+++ b/t/t5601-clone.sh
@@ -642,6 +642,22 @@ test_expect_success 'shallow clone locally' '
 	( cd ddsstt && git fsck )
 '
 
+test_expect_success 'GIT_CLONE_DEPTH acts as a fallback for --depth' '
+	GIT_CLONE_DEPTH=1 git clone --no-local src depth-env &&
+	test_path_is_file depth-env/.git/shallow &&
+	test 1 = $(git -C depth-env rev-list --count HEAD)
+'
+
+test_expect_success 'explicit --depth overrides GIT_CLONE_DEPTH' '
+	GIT_CLONE_DEPTH=1 git clone --depth=2 --no-local src depth-override &&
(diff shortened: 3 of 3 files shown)


builtin/clone.c
```diff
@@ -1022,6 +1022,12 @@ int cmd_clone(int argc,
 		usage_msg_opt(_("You must specify a repository to clone."),
 			builtin_clone_usage, builtin_clone_options);
 
+	if (!option_depth) {
+		const char *env_depth = getenv("DEPTH");
+		if (env_depth && *env_depth)
+			option_depth = xstrdup(env_depth);
+	}
+
```
davvid, line 1031: Without regard to whether or not this is a good feature, at minimum I would suggest using a more specific name. Something like `GIT_CLONE_DEPTH` would be better than `DEPTH` as it avoids an overly-generic name.

Other than that, this patch needs to update the appropriate tests in`t/` and the corresponding `Documentation/` files.

davvid:
When renaming the var and adding tests, please squash those changes into the first commit and revise its message (which also mentions DEPTH) to account for the rename.

h8d13:
@davvid did just that :) thanks again for the tips.

h8d13:
> One might argue this isn't very impactful, since we would expect the
process to exit after git-clone completes, but there are already several
explicit calls to free and related functions at the end of cmd_clone.

In `PKGBUILD` the common case is;

```shell
git clone --depth 1 <url> <dest>
cd dest
bash run.sh
# here run.sh has its own clone dependencies (perhaps even multiple)
# --depth 1 is now lost
```

Hopefully that explains better the need

Mroik:
> > One might argue this isn't very impactful, since we would expect the
> > process to exit after git-clone completes, but there are already several
> > explicit calls to free and related functions at the end of cmd_clone.
> 
> In `PKGBUILD` the common case is;
> 
> ```shell
> git clone --depth 1 <url> <dest>
> cd dest
> bash run.sh
> # here run.sh has its own clone dependencies (perhaps even multiple)
> # --depth 1 is now lost
> ```
> 
> Hopefully that explains better the need

Just so you know, the ones on the mailing list (including the maintainer) won't be able to see what you write here. Development discussions happen on the mailing list.

h8d13:
Sry I have newer used a mailing list :grinning: 

I was hoping this satanic bot would transmit the info

dscho:
> Sry I have newer used a mailing list 😀
> 
> I was hoping this satanic bot would transmit the info

@h8d13 That's quite an interesting way to show gratitude for the amount of effort that was put into GitGitGadget, for the effort trying to make contributing patches to the Git mailing list less of a challenge. Quite revealing.

h8d13:
Sorry I meant that as a joke, because of the needs signature  and commit format rules, etc, I learned a few new `git` tricks along the way `-s` notably :)

h8d13:
Does replying in email to <gitgitgadget@gmail.com> suffice ? @Mroik 

Sorry I'm very confused with this mailing list stuff

Mroik:
> Does replying in email to [gitgitgadget@gmail.com](mailto:gitgitgadget@gmail.com) suffice ? @Mroik
> 
> Sorry I'm very confused with this mailing list stuff

You should've received all the emails that are sent to the mailing list regarding this topic, you should be able to find them in your inbox (hadean-eon-dev@proton.me) as GitGitGadget cc's everything to you.

To participate in the discussion you can simply hit "reply-all" to the email you want to reply to, in your email client of choice (or webclient if you access email from a web browser). It's important that you hit "reply-all" and not "reply" or "reply-list", not everyone that participate in the discussion is subscribed to the mailing list.

I sense that you haven't read the guidelines/contribution guide before trying to submit a patch, if so I invite you to give it a read: https://git-scm.com/docs/MyFirstContribution

h8d13:
Thanks I think I've managed to do it. :mailbox_with_mail: 

Will catch-up on some reading lol

But it was not working from proton for some reason.

bertyuan:
> But it was not working from proton for some reason.

@h8d13 Hope this tutorial could help you a little bit. ;-)

https://useplaintext.email/#protonmail

But it seems you have to wrap your email manually if using protonmail.

h8d13:
Thanks @bertyuan :=)
Seems to work now. Replid on thread https://lore.kernel.org/git/xmqqeciafowq.fsf@gitster.g/T/#u

bertyuan:
@h8d13 Congrats!

image
But your reply is missing the header `In-Reply-To`, so it cannot make it a strict thread on the mailing list.<br>

You may check your email client. Logically speaking, if you reply to all in an email, the email header `In-Reply-To` will be automatically added.

The maintainer often uses `b4` to pull iterations from the mailing list. If `In-Reply-To` header is missing, he might meet problems when pulling. So next time when sending your v2, it'd be a better choice to check all these works fine ;-)

h8d13:
Header as in: content first line or subject itself. 

When I use the `mail-to` I see `Re: [PATCH] ...`

Then I should be adding as first line:

`In-Reply-To: ID@domain`

?

Sorry this very new 😅😅😅

bertyuan:
Take Junio's as an example:
image
It has From/To/Cc/Subject/Date/Message-ID/In-Reply-To, these are all called headers (less rigorously speaking)

And yours:
image
Missing In-Reply-To, so it cannot be recognized as a thread for archiving.

But as I just said, if you click "Reply all" to reply, this problem should not occur here. I haven't seen Gmail having any problems here.

bertyuan:
If you are doing this way,
image
I'm afraid your MUA doesn't support setting that header ;-)

h8d13:
Yeah that waht I was doing aha. And it seems to not have the correct `In-Reply-To:`

So if I add it in the content first line that works ?

bertyuan:
> So if I add it in the content first line that works ?

I dunno. I guess that doesn't work. 

As far as I know, if your MUA supports editing headers, you can add the header in the raw file.

image

Here it says "Be sure your reply has a **Subject:** header at the top and a blank line before the message body." I don't know how about In-Reply-To, maybe you could have a try by sending an email to yourself ;-) I don't know much about emails though.

PR #2335, opened 2026-06-13
bisect: add --reset-when-found to leave when done

HaraldNordgren:
Add a `--reset-when-found` option to `git bisect` that resets the bisect session when culprit is found.

Changes in v7:
- Drop unrelated formatting changes to `bisect_next()` and `bisect_auto_next()`. Restore the original control flow by removing the unnecessary `else`.

Changes in v6:
- Reuse the existing bad bisect ref instead of propagating the culprit OID through `bisect_next_all()`.
- Remove the redundant `reset_when_found_arg_seen` flag and use `RESET_WHEN_FOUND_NONE` to detect whether the option was given.

Changes in v5:
- Move automatic reset handling to `cmd_bisect()` after subcommand resources are closed.
- Propagate the first-bad commit OID from `bisect_next_all()` and remove `defer_reset` plumbing.
- Separate checkout from state cleanup in `bisect_reset()`, and use `bool` for its quiet flag.

Changes in v4:
- Simplify translation calls.
- Avoid `git` subshell calls in tests, that can bury errors.

Changes in v3:
- Rename `--auto-reset` to `--reset-when-found`, including internal names.
- Defer `git bisect run` cleanup until captured output is printed and `BISECT_RUN` is closed. Drop the open-descriptor preparatory change, retaining the existing filename-based output handling.

Changes in v2:
- Add option `--auto-reset[=<where>]` with option to go to final commit as well as original.
- Refactored tests.

### DIFF
--- a/Documentation/git-bisect.adoc
+++ b/Documentation/git-bisect.adoc
@@ -10,7 +10,7 @@ SYNOPSIS
 --------
 [synopsis]
 git bisect start [--term-(bad|new)=<term-new> --term-(good|old)=<term-old>]
-		 [--no-checkout] [--first-parent] [<bad> [<good>...]] [--] [<pathspec>...]
+		 [--no-checkout] [--first-parent] [--reset-when-found[=<where>]] [<bad> [<good>...]] [--] [<pathspec>...]
 git bisect (bad|new|<term-new>) [<rev>]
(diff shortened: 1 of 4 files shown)


Anjalo133 (commented):
G

PR #2337, opened 2026-06-13
history: add squash subcommand to fold a range

HaraldNordgren:
Adds `git history squash <revision-range>` to fold a range of commits.

Changes in v15:
- Fixed istarts_with() → starts_with() in squash_check_autosquash_subject() so fixup! ABCdef (mixed/upper-case hex) is no longer accepted.
- Fixed a typo in the adjacent comment (squshed → squashed).

Changes in v14:
- Update commit message trailers.
- Simplify `rev-parse` and `tr` complexity in the tests.

Changes in v13:
- Split the squash implementation into five patches covering parsing, range validation, branch protection, commit creation, and message editing.
- Print the sorted names of local branches that prevent a squash, with clearer advice for `--update-refs=head`.
- Isolate `--no-edit` behavior from the final patch that enables default editor-based message combining.

Changes in v12:
- Incorporated fixups from Phillip:
  - Reworks range validation into a single walk that rejects roots and multiple tips while preserving every parent when squashing into a merge.
  - Resolves fixup!, squash! and amend! targets directly, preserving message intent and safely consolidating related markers.
  - Builds the editor template from the exact selected revisions, including exclusions, while retaining `--no-edit` behavior.
  - Protects descendant local branches while leaving tags and remote-tracking refs unchanged.

Changes in v11:
- Make message editing the default with the autosquash-style template, add `--no-edit` instead of squash-specific `--reedit-message`.
- Validate one actual boundary and tip, rejecting root-reaching and multi-tip ranges.
- Protect only interior local branches, leaving tags and remote-tracking refs unchanged.
- Move sequencer preparation before squash and fold the later message-editing patch into the feature commit.

Changes in v10:
- Record the full revision expression in squash reflog.
- Preserve the boundary-walk invariant when sanitizing rev-list options.
- Clarify `amend!` and `--reedit-message` documentation.

Changes in v9:
- Use the last `amend!` targeting the oldest folded commit as the default squashed message. Ignore `amend!` markers targeting later commits while selecting that replacement message.
- Improve tests.

Changes in v8:
- `--reedit-message` now builds the same editor template as `git rebase -i --autosquash`: `fixup!`, `squash!` and `amend!` commits are grouped under the commit they target instead of shown in commit order, and an `amend!` replaces its target's message.
- A `fixup!`, `squash!` or `amend!` is refused only when its target is outside the range, so several fixups for an in-range commit fold together. A range that is entirely markers for one below-range target is combined into a single commit, keeping the last `amend!` message.
- Merges inside the range are folded when the range has a single base, with no dedicated opt-in flag, `--ancestry-path` ensures only commits descended from the base are folded, and a range reaching more than one base is rejected.
- Rev-list options are accepted and sanitized the way `git replay` does, forcing the walk order back with a warning, which also fixes `git history squash -- --reverse` slipping past the previous option check.
- Kept this as an explicit `squash` subcommand rather than making `--reedit-message` the default or renaming the command.

Changes in v7:
- `--reedit-message` now builds the same editor template `git rebase -i` shows for a squash (a `combination of N commits` banner with each folded message under its own header) and follows autosquash for markers: a `fixup!` message falls out (commented under a `will be skipped` header), while a `squash!` or `amend!` keeps its body with only the marker subject commented so its remark can be reworded in. Only the message text is affected, every commit's changes are always folded in.
- Reuse `git rebase -i`'s squash-message code: a preparatory `sequencer:` commit extracts the banner, header and marker-comment helpers so both rebase and `git history squash` build the identical template from one source.
- Refuse a range whose oldest commit is a `fixup!`, `squash!` or `amend!`, since the marker's target cannot be inside the range.
- Reorder the squash usage so dashed options come before `<revision-range>`, and spell out `HEAD` instead of `@` in the documentation and examples.
- Expand the squash commit message and documentation with this overview, and scope the merge limitation so it no longer contradicts squash folding a single-base interior merge.

Changes in v6:
- `git history squash` now accepts multiple revision arguments, read like the arguments to `git-rev-list`, so a compound range such as `@~3.. ^topic` works.
- The base to reparent onto is now the oldest in-range commit's parent; a boundary other than that base means the range has more than one base and is rejected. This also fixes the earlier overly-restrictive handling of merges and side branches.
- A single-commit range (e.g. `@^!`) is rejected with "nothing to squash" (this also covers the `@^!`-style example that previously succeeded silently).
- Commit messages reworded: the squash commit now gives an overview of `fixup!`/`squash!`/`amend!` handling, rewording, merge-parent and ref behavior.

Changes in v5:
- The range walk now uses `--ancestry-path`, so only commits descended from the base are folded; a single revision such as `HEAD` or `HEAD~1` is now rejected as "not a `<base>..<tip>` range" rather than treated as a squash down to the root.
- This adopts the `--ancestry-path` suggestion; the multi-base rejection is unchanged, so a side branch that forked before the base and merged in is still refused.
- Added tests covering more merge topologies: two interior merges, a nested merge, an octopus merge, an octopus arm forked before the base, a merge among the descendants replayed above the range, and a ref pointing at an interior merge commit.

Changes in v4:
- `git history squash` now detects when another ref points at a commit inside the range being folded and refuses, with an `advice.historyUpdateRefs` hint to use `--update-refs=head`.
- A merge inside the range is folded fine as long as the range has a single base; a range with merge commit at the tip or base also folds correctly. Only a range with more than one base is rejected.

Changes in v3:
- Moved the feature out of `git rebase` and into a new `git history squash <revision-range>` subcommand, per the list discussion. `git rebase --squash` is dropped.
- Takes an arbitrary range (`git history squash @~3..`, `git history squash @~5..@~2`), folding it into the oldest commit and replaying any descendants on top.
- Implemented as a single tree operation rather than picking each commit, so there are no repeated conflict stops (addresses Phillip's efficiency point).
- A merge inside the range is folded fine, only a range with more than one base is rejected.
- `--reedit-message` seeds the editor with every folded-in message, not just the oldest.

cc: "D. Ben Knoble" <ben.knoble@gmail.com>
cc: "Matt Hunter" <m@lfurio.us>
cc: Tuomas Ahola <taahol@utu.fi>


taahol:
Related: https://github.com/gitgitgadget/git/issues/1135

Documentation/git-history.adoc
```diff
@@ -97,8 +98,22 @@ linkgit:gitglossary[7].
 It is invalid to select either all or no hunks, as that would lead to
 one of the commits becoming empty.
 
-OPTIONS
```
taahol, line 116: Was that deleted accidentally?

HaraldNordgren: Yes, accident! Good catch!

taahol:
Perhaps something like
```diff
diff --git a/t/meson.build b/t/meson.build
index c5832fee05..cf6db143d1 100644
--- a/t/meson.build
+++ b/t/meson.build
@@ -399,6 +399,7 @@ integration_tests = [
   't3451-history-reword.sh',
   't3452-history-split.sh',
   't3453-history-fixup.sh',
+  't3454-history-squash.sh',
   't3500-cherry.sh',
   't3501-revert-cherry-pick.sh',
   't3502-cherry-pick-merge.sh',
```
?

PR #2350, opened 2026-06-25
rust: validate object map insert algorithms

wufengwind:
ObjectMap::insert() accepts a storage OID and a compatible OID in either
order, but it currently checks whether oid1 uses the compatible algorithm,
and if not, assumes oid2 is the compatible one without validating oid2.

That means inputs with two OIDs using the same hash algorithm, or an OID
using an unknown hash algorithm, are accepted and one of them is silently
treated as the storage OID. This breaks the object map invariant that each
entry must contain exactly one storage hash and one compatible hash.

Teach ObjectMap::insert() to decode and validate both OID algorithms before
inserting anything. The function now accepts only the two valid permutations:
(storage, compat) and (compat, storage). Unknown algorithms and mismatched
algorithm pairs are rejected via ObjectMapInsertError.

The tests cover successful insertion in either order, same-algorithm input,
and unknown-algorithm input.

Tested with:

- cargo fmt --all -- --check
- git diff --check
- cargo test

### DIFF
--- a/src/loose.rs
+++ b/src/loose.rs
@@ -510,6 +510,17 @@ pub struct ObjectMap {
     batch: Option<ObjectMemoryMap>,
 }
 
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum ObjectMapInsertError {
+    InvalidHashAlgorithm(u32),
+    MismatchedAlgorithms {
+        storage: HashAlgorithm,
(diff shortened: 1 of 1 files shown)


wufengwind:
This is my first contribution via GitGitGadget. Could someone please /allow me so that I can submit the patch to the mailing list? Thanks.

PR #2353, opened 2026-07-07
unpack-trees: avoid quadratic index scan in next_cache_entry()

hferreiro:
Changes since v1: adjust the synthetic index size based on the EXPENSIVE prerequisite.

cc: Henrique Ferreiro <hferreiro@igalia.com>

### DIFF
--- a/t/perf/p0009-diff-pathspec.sh
+++ b/t/perf/p0009-diff-pathspec.sh
@@ -1,0 +1,32 @@
+#!/bin/sh
+
+test_description='Tests performance of diffing the working tree with a pathspec'
+
+. ./perf-lib.sh
+
+test_perf_fresh_repo
+
+count=10000
+if test_have_prereq EXPENSIVE
+then
+	count=100000
+fi
+
+# The entries exist only in the index, which is enough to
+# exercise the index scan.
+test_expect_success 'setup' '
+	blob=$(echo content | git hash-object -w --stdin) &&
+	{
+		printf "100644 $blob\taaa/file\n" &&
+		printf "100644 $blob\tf%s\n" $(test_seq $count)
+	} | git update-index --index-info &&
+	git commit -q -m initial &&
+	mkdir -p aaa &&
+	echo content >aaa/file
+'
+
+test_perf 'diff pathspec subtree' '
+	git diff HEAD -- aaa/file
+'
+
+test_done
--- a/unpack-trees.c
+++ b/unpack-trees.c
@@ -671,8 +671,10 @@ static struct cache_entry *next_cache_entry(struct unpack_trees_options *o)
 
 	while (pos < index->cache_nr) {
 		struct cache_entry *ce = index->cache[pos];
-		if (!(ce->ce_flags & CE_UNPACKED))
+		if (!(ce->ce_flags & CE_UNPACKED)) {
+			o->internal.cache_bottom = pos;
 			return ce;
+		}
 		pos++;
 	}
 	return NULL;


abhisharma30 (commented):
## Review summary

Thanks for the clear diagnosis and the huge win (Chromium-scale index: minutes → tens of ms). I walked `next_cache_entry()` / `mark_ce_used()` / `switch_cache_bottom` / `restore_cache_bottom` and the leftover drain loop, and also had a second pass focused on the perf test.

**Verdict: looks correct and ready to land** (matches Junio’s “sounds sensible” / “Will merge to 'next'?” thread). Residual notes below are non-blocking.

### Algorithm

The quadratic path is real:

1. Pathspec-limited tree walk + `switch_cache_bottom` / `restore_cache_bottom` rewinds `cache_bottom` onto already-`CE_UNPACKED` early entries (non-`--cached` only — matches the “`--cached` was already fast” claim).
2. `mark_ce_used()` only advances `cache_bottom` when the marked entry *is* `cache[cache_bottom]`.
3. The leftover loop then calls `next_cache_entry()` for each remaining index entry; without recording progress, each call rescans a growing unpacked prefix → O(n²).

Setting `o->internal.cache_bottom = pos` when returning the first non-unpacked entry restores the invariant “bottom is a safe restart point for the next scan” and is consistent with how `find_cache_pos()` already slides the bottom past unpacked entries. Tree-vs-index order (Junio’s question) is handled: out-of-order marks leave holes, but the first not-yet-unpacked entry at/after the previous bottom is still the correct resume point.

### Historical note worth remembering

`cache_bottom` has been correctness-sensitive before (`f2a454e0a5` / `99430aa12c` sparse-index history). This change is small and in the right direction, but it is not “just a micro-optim”. Existing ambient sparse/pathspec tests don’t construct this exact layout; the new perf script is the primary guard.

### Process

Already in `seen` as `hf/unpack-trees-quadratic-scan`. EXPENSIVE sizing feedback from v1 is addressed in v2.

t/perf/p0009-diff-pathspec.sh
```diff
@@ -0,0 +1,32 @@
...
+test_perf_fresh_repo
+
+count=10000
+if test_have_prereq EXPENSIVE
+then
+	count=100000
+fi
+
+# The entries exist only in the index, which is enough to
+# exercise the index scan.
```
abhisharma30, line 16: **[nit]** Comment says entries “exist only in the index”, but the setup `git commit`s them into HEAD as well. Only the *worktree* is sparse (only `aaa/file` is checked out). The bulk `f*` paths still live in the tree + index.

Suggest rephrasing, e.g. that the bulk of the index need not be present in the worktree; what this exercises is the leftover index scan after a pathspec-limited tree walk (the non-`--cached` `restore_cache_bottom` path).

t/perf/p0009-diff-pathspec.sh
```diff
@@ -0,0 +1,32 @@
...
+
+. ./perf-lib.sh
+
+test_perf_fresh_repo
+
+count=10000
+if test_have_prereq EXPENSIVE
+then
+	count=100000
+fi
```
abhisharma30, line 13: **[suggestion]** Default `count=10000` (100k only under `EXPENSIVE`) may be a soft signal on noisy machines — O(n²) means 10k is ~100× less work than the 100k numbers in the commit message (~7s → 0.02s).

Not blocking. Optional: bump default a bit, or one-line note in the test description that a clear before/after wants `--long-tests` / `GIT_TEST_LONG=t`. The EXPENSIVE split itself is the right pattern (same as `p4209-pickaxe.sh`).

unpack-trees.c
```diff
@@ -671,8 +671,10 @@ static struct cache_entry *next_cache_entry(struct unpack_trees_options *o)
 
 	while (pos < index->cache_nr) {
 		struct cache_entry *ce = index->cache[pos];
-		if (!(ce->ce_flags & CE_UNPACKED))
+		if (!(ce->ce_flags & CE_UNPACKED)) {
+			o->internal.cache_bottom = pos;
```
abhisharma30, line 675: **[nit / optional]** When the scan exhausts and returns `NULL`, `cache_bottom` is left pointing at the last examined region rather than `cache_nr`. A subsequent `next_cache_entry()` would rescan the fully-unpacked prefix again (still O(n) once, not quadratic in the usual leftover-loop pattern).

Not required for this fix — after the leftover loop finishes, `mark_all_ce_unused()` clears state. Only worth considering if you want the “bottom is past all unpacked” invariant to hold on the empty return as well:

```c
o->internal.cache_bottom = index->cache_nr;
return NULL;
```

unpack-trees.c
```diff
@@ -671,8 +671,10 @@ static struct cache_entry *next_cache_entry(struct unpack_trees_options *o)
...
 			return ce;
```
abhisharma30, line 676: **[suggestion — coverage, non-blocking]** Pure perf bug ⇒ no functional test that fails without the fix is expected. Given past `cache_bottom`/sparse correctness churn, a *tiny* functional case (dir-first path + a few later siblings, `git diff HEAD -- aaa/file` output check, optionally sparse-index) would be a cheap safety net that “wrong but fast” cannot hide behind p0009 alone.

I would not block landing on that; the algorithm change is small and the perf shape is well chosen.

PR #2362, opened 2026-07-22
http: add a config to limit the connection time

GalaxySnail:


GalaxySnail:
> Error: Could not determine full name of GalaxySnail

What's happening?

dscho:
> > Error: Could not determine full name of GalaxySnail
> 
> What's happening?

@GalaxySnail The Git project [requires "Random J Developer <random@developer.example.org>"-like sign-offs](https://git-scm.com/docs/SubmittingPatches#sign-off), and GitGitGadget enforces that even for your own Cc: to which your email gets sent (so that the conversation on the Git mailing list about your patch includes you).

To determine your name and email, GitGitGadget [looks at your public GitHub info](https://github.com/gitgitgadget/gitgitgadget/blob/faf4896d9395c7bb1bf26cec688eaab704ebe646/lib/ci-helper.ts#L1259-L1271). The user name is obtained [via GitHub's REST API](https://github.com/gitgitgadget/gitgitgadget/blob/faf4896d9395c7bb1bf26cec688eaab704ebe646/lib/github-glue.ts#L494), and if you look at the response of https://api.github.com/users/GalaxySnail you will notice that both `name` and `email` are `null`. GitGitGadget can therefore not send anything with you in Cc: and refuses to send anything at all.

GalaxySnail:
Thanks! I have set my name in the user profile now. However, although I have already set my public email to `me@glxys.nl`, the `email` field in this api is still `null`. I also checked that the "keep my email addresses private" setting is "off". That's weird.

edit: It seems that the `email` field doesn't matter.

dscho:
> Thanks! I have set my name in the user profile now. However, although I have already set my public email to `me@glxys.nl`, the `email` field in this api is still `null`. I also checked that the "keep my email addresses private" setting is "off". That's weird.
> 
> edit: It seems that the `email` field doesn't matter.

I was puzzled at first, too. Because it _does_ matter. And the really curious thing is: GitGitGadget [successfully determined your email address](https://github.com/gitgitgadget-workflows/gitgitgadget-workflows/actions/runs/29993716074/job/89162288679#step:6:49)!

So I think that the solution to the riddle is that anonymous REST API requests don't get that `email` field filled, whereas GitGitGadget issues an authenticated request, which does get that information.

GalaxySnail:
> So I think that the solution to the riddle is that anonymous REST API requests don't get that `email` field filled, whereas GitGitGadget issues an authenticated request, which does get that information.

That makes sense, thank you for the explanation!

PR #2364, opened 2026-07-25
checkout -m: refine autostash fallback

HaraldNordgren:
Avoiding `checkout -m` autostash retries when no tracked local changes exist and visually separating autostash conflict advice from the subsequent branch-switch message. Addresses #leftoverbits from here: https://lore.kernel.org/git/cfd09dbf-8d77-4464-8030-3a0ffb4aeae7@gmail.com/

Changes in v5:
- Improve commit messages.
- Create helper `do_stash_apply`.

Changes in v4:
- Conflicts now exit with status 1 like `merge-tree`, other failures exit 128 so exit 1 unambiguously means conflicts. Stash changes split into their own commit.
- The autostash apply helpers use the return value (`enum stash_apply_result`) instead of an out-parameter, and only claim conflicts when `git stash apply` actually reported them.

Changes in v3:
- Use enum for git stash return values, to separate conflict from generic error.

Changes in v2:
- Simplify logic and combine to one commit.
- Test full output with test_cmp.

### DIFF
--- a/Documentation/git-stash.adoc
+++ b/Documentation/git-stash.adoc
@@ -426,6 +426,15 @@ include::includes/cmd-config-section-all.adoc[]
 :git-stash: 1
 include::config/stash.adoc[]
 
+EXIT STATUS
+-----------
+
+The `git stash` subcommands exit with status 0 on success.  The
+subcommands that apply a stash entry, i.e. `apply`, `pop` and `branch`,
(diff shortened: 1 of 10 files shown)


HaraldNordgren:
I didn't realize until after creating that one that leftoverbits are
usually left for beginners, would it be better to leave this as a good
first task?


Harald

spacexm406-web (commented):
Improve

PR #2378, opened 2026-08-09
refs: support local GIT_NAMESPACE views

johnrichardrinehart:
Git namespaces currently isolate refs for `git-upload-pack` and
`git-receive-pack`, but ordinary local Git commands do not use
`GIT_NAMESPACE` as a repository view.

This prevents a shared object database from providing normal local branch
semantics. For example, `GIT_NAMESPACE=bar git rev-parse refs/heads/main`
cannot resolve `refs/namespaces/bar/refs/heads/main`.

This series makes ordinary ref access namespace-aware when `GIT_NAMESPACE`
is set. Local commands resolve and update logical refs inside the selected
namespace, while upload-pack, receive-pack, and HTTP backend retain their
existing server behavior.

The series also makes garbage collection protect refs in every namespace.
This preserves objects shared by namespaces without requiring per-namespace
object stores or a separate garbage-collection command.

The series contains two commits:

- `refs: support local namespace views`
- `gc: protect refs in every namespace`

The tests cover local branch and ref operations, fetch, push, worktrees,
existing server namespace behavior, hidden-ref queries, and cross-namespace
garbage collection.

Checks run:

- `t1500-rev-parse.sh`
- `t3200-branch.sh`
- `t5509-fetch-push-namespaces.sh`
- `t5510-fetch.sh`
- `t6021-rev-list-exclude-hidden.sh`

This work was implemented by GPT-5.6 Luna High.

### DIFF
--- a/Documentation/gitnamespaces.adoc
+++ b/Documentation/gitnamespaces.adoc
@@ -8,6 +8,7 @@ gitnamespaces - Git namespaces
 SYNOPSIS
 --------
 [verse]
+GIT_NAMESPACE=<namespace> git <command>
 GIT_NAMESPACE=<namespace> 'git upload-pack'
 GIT_NAMESPACE=<namespace> 'git receive-pack'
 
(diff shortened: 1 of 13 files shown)


benknoble:
@johnrichardrinehart asked on Git Mentoring about this change and specifically the use of an LLM; current policy is https://github.com/git/git/blob/010afd3166ddc64c9863b1506f12cbcdda0d4ea1/Documentation/SubmittingPatches#L571, so that’s a useful place to start evaluating for yourself :)

PR #2380, opened 2026-08-10
copy: prefer reflinks for file copies

vladpetric:


vladpetric:
You’re right that a successful hardlink is preferable for immutable object files.

My motivation came from an actual local clone in my home dir (first repo cloned normally, second repo cloned with --reference to the first one).

Both repositories were on the same reflink-capable ZFS filesystem, but the normal git clone produced neither hardlinked nor reflinked object files, so the objects were copied byte-for-byte. With the FICLONE path enabled, the same setup successfully reflinked the object files.

That is a useful fallback when Git’s hardlink path does not produce links, but it does not justify trying reflink before a successful hardlink. I revised the order to:

1. hardlink;
2. reflink;
3. byte-for-byte copy.

With --no-hardlinks, the order would remain reflink followed by byte-for-byte copy.

dscho:
> You’re right that a successful hardlink is preferable for immutable object files.
> 
> My motivation came from an actual local clone in my home dir (first repo cloned normally, second repo cloned with --reference to the first one).
> 
> Both repositories were on the same reflink-capable ZFS filesystem, but the normal git clone produced neither hardlinked nor reflinked object files, so the objects were copied byte-for-byte. With the FICLONE path enabled, the same setup successfully reflinked the object files.
> 
> That is a useful fallback when Git’s hardlink path does not produce links, but it does not justify trying reflink before a successful hardlink. I revised the order to:
> 
> 1. hardlink;
> 2. reflink;
> 3. byte-for-byte copy.
> 
> With --no-hardlinks, the order would remain reflink followed by byte-for-byte copy.

@vladpetric please note that GitGitGadget only mirrors emails to PR comments, not the other way round. Therefore, Junio won't see any of your reply.

If you want that reply to reach the intended recipient, you _will_ need to follow the guidance behind that "how to reply to this email" link.

jibonhasnat7-dotcom:
> 

UFT-32BE

vladpetric:
> > 
> 
> UFT-32BE

I beg your pardon?

PR #2382, opened 2026-08-13
rev-parse: compute object names in another hash algorithm

xnox:
"git rev-parse --output-object-format=sha256 HEAD^{tree}" could not
answer in a SHA-1 repository.  And vice-versa.  Without
extensions.compatObjectFormat the option was rejected outright, and with
it the answer was wrong: the extension records a mapping only for objects
written while it is enabled, so for the objects a repository already
contained the lookup missed, repo_oid_to_algop() left the object ID
untouched, and rev-parse printed the SHA-1 name with a zero exit code.

The machinery to convert object contents between algorithms is already
here.  convert_tree_object() and friends rewrite the object IDs a tree,
commit or tag refers to, calling repo_oid_to_algop() for each one, so the
recursion is written; what is missing is a base case.  Give
repo_oid_to_algop() one: on a lookup miss, read the object, convert its
contents, and hash the result.  A blob needs no conversion at all, as its
contents are the same either way, only a rehash.

Names computed this way are remembered during the process execution,
so that a tree that reaches the same subtree by several paths pays for
it once.  These are not stored permamently.  In the future a side-car
cache could be added for these results, or eventual dual-format v3
could be used to lazy compute/convert and store these.

Commits are refused rather than computed.  A commit names its parents, so
converting one converts the entire history behind it; that is a repository
conversion, not an answer about a single object, and recursing over it
here would be bounded only by the length of the history.  A tree holding a
submodule is refused for the same reason, and now says which entry is at
fault instead of reporting an unreadable object.

Since names can now be computed, accept any algorithm git knows rather
than only the configured compatibility one, and check what
repo_oid_to_algop() returns so that a name we cannot produce is an error
instead of the storage name.

On git.git's tree at 9936c1b52a (3089 files, 26MB) this names the tree
in 0.26s, against 0.25s for reading and hashing that content by hand,
so close to the floor for the work involved. On linux.git tree of
1.6GiB it takes 8s to compute.

Initially, I have implemented contrib git-sha256-tree.sh script using
a temporary git repo and perform fast-export/fast-import of a single
commit and its tree.  It is a lot slower due to needless work of
fast-import that is discarded.  However, a contrib script carries less
maintainance and is more portable to existing installations.  If there
is interest, I can publish that implementation as well.  The goal is
to help with interop, and have the ability to record tree names in
either format today; to check again later if and when a given project
switches to sha256 format.

Signed-off-by: Dimitri John Ledkov <dimitri.ledkov@surgut.co.uk>
Assisted-by: Claude Opus 5 xHigh

### DIFF
--- a/Documentation/git-rev-parse.adoc
+++ b/Documentation/git-rev-parse.adoc
@@ -181,6 +181,19 @@ Specifying "sha256" translates if necessary and returns a sha256 oid.
 +
 Specifying "storage" translates if necessary and returns an oid in
 encoded in the storage hash algorithm.
++
+A translation that `extensions.compatObjectFormat` did not record, which is
+every object a repository already contained when that extension was turned
+on, is computed on demand.  Since an object's name in another algorithm is
+defined over the names of the objects it refers to, this reads everything
+reachable from the object, and so costs time proportional to the total size
+of that content.
++
+Commits cannot be named this way.  A commit refers to its parents, so naming
+one would mean converting the whole history behind it, which is a repository
+conversion rather than a question about a single object.  For the same
+reason a tree containing a submodule cannot be named, as its gitlink refers
+to a commit in another repository.
 
 Options for Objects
 ~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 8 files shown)


xnox:
Submitted at:
- https://lore.kernel.org/git/20260814134804.680154-1-dimitri.ledkov@chainguard.dev/T/#u

PR #2384, opened 2026-08-14
[RFC] Add git pull --hard mode

ArturBieniek4:

### DIFF
--- a/Documentation/git-pull.adoc
+++ b/Documentation/git-pull.adoc
@@ -24,7 +24,7 @@ with no arguments this defaults to the <<UPSTREAM-BRANCHES,upstream>>
 for the current branch.
 Then it integrates that branch into the current branch.
 
-There are 4 main options for integrating the remote branch:
+There are 5 main options for integrating the remote branch:
 
 1. `git pull --ff-only` will only do "fast-forward" updates: it
    fails if your local branch has diverged from the remote branch.
@@ -32,6 +32,7 @@ There are 4 main options for integrating the remote branch:
 2. `git pull --rebase` runs `git rebase`
 3. `git pull --no-rebase` runs `git merge`.
 4. `git pull --squash` runs `git merge --squash`
+5. `git pull --hard` runs `git reset --hard` to the fetched branch.
 
 You can also set the configuration options `pull.rebase`, `pull.squash`,
 or `pull.ff` with your preferred behaviour.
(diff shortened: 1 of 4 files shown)


SirJosh1987:
Allow

taahol:
It seems nothing happened as this workflow failed: https://github.com/gitgitgadget-workflows/gitgitgadget-workflows/actions/runs/32055421327 @rimrul

rimrul:
Yes, it couldn't process the `/submit` comment.

ArturBieniek4:
Thanks for the feedback.

I agree that whether replacing local history belongs under git pull is a 
separate design question, and I understand the distinction you are 
making between integration and replacement.

I am less convinced, though, by the more general argument that 
destructive operations should intentionally remain inconvenient. Git 
already exposes explicitly destructive operations as single commands, 
with git push --force being perhaps the clearest example. In that case 
the possible loss may even affect shared remote history rather than only 
local state.

My thinking with git pull --hard was similar: make the destructive 
behavior explicit in the option name, while allowing pull to perform the 
fetch and validate that there is exactly one integration candidate 
before resetting to it. That seemed preferable to users encoding the 
same operation in aliases or scripts around fetch and reset --hard.

That being said, I understand the objection that a hard reset is 
fundamentally not history integration, and therefore may simply not 
belong under git pull.

Thanks.

taahol:
@ArturBieniek4

Please (re-)read <https://gitgitgadget.github.io/reply-to-this>. As the development of Git happens on the mailing list (rather than Github), it is not enough, unfortunately, to reply to that “bot-carried” message. Instead, you should find the original email from Junio and reply to that and make sure, in particular, that `git@vger.kernel.org` remains in the CC list.

Indeed, although you did reply via email, it was regrettably a wrong message you wrote your answer to. And even though people can see your reply here in Github, it is expected (and required) that such messages go to the list instead. That way, the list archive can form a comprehensive story on how the patches came about.

Thanks.

ArturBieniek4:
Thanks for help, I'm doing such a thing the first time so I'm sorry for my mistakes. I first replied correctly (at least I think) via email but the list rejected the message because it was in HTML. So I replied in plain text but somehow Thunderbird then decided to change the recipients (possibly I just did something wrong and did not realize). Also I got a reply from the GitHub bot "The recipient's mailbox is full and can't accept messages now.". I will RTFM carefully and try again tomorrow.
Thanks.

taahol:
> Also I got a reply from the GitHub bot "The recipient's mailbox is full and can't accept messages now.".

Hmm, whose mailbox? But if that was `gitgitgadget@gmail.com`, then no, AFAIK the gadget doesn't even read its mailbox so it's not an useful place to send anything.

Again, I'm far from an expert in these things, so…

PR #2386, opened 2026-08-19
git-p4: honor client spec when unshelving

cazador481:
## Description

When `git-p4.useclientspec` is enabled, `git p4 unshelve` creates a `P4Sync` instance without initializing its client-spec mapping. The unshelved files are therefore written relative to the depot path instead of their client-relative paths.

Initialize the client-spec mapping in `P4Unshelve.run()` and add a regression test covering a depot-to-client remap.

## Testing

- `make -j4`
- `make test T=t9832-unshelve.sh` (skipped: this environment does not have `p4`/`p4d`)
- `make test T=t9809-git-p4-client-view.sh` (skipped: this environment does not have `p4`/`p4d`)
- `python3 -m py_compile git-p4.py`
- `sh -n t/t9832-unshelve.sh`
- `git diff --check`

GitHub Issues are disabled for `git/git`, so the problem description is included here as the issue record.

### DIFF
--- a/git-p4.py
+++ b/git-p4.py
@@ -4470,6 +4470,10 @@ class P4Unshelve(Command):
         sync = P4Sync()
         changes = args
 
+        if gitConfigBool('git-p4.useclientspec'):
+            sync.useClientSpec = True
+            sync.clientSpecDirs = getClientSpec()
+
         # only one change at a time
         change = changes[0]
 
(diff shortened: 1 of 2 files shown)


dscho:
> Signed-off-by: cazador481 <[...]>

@cazador481 could you please use your real name in the `Signed-off-by` line and in your author information? The Git project [insists on that](https://git-scm.com/docs/SubmittingPatches#dco). You can fix it by configuring `user.name`, then `git commit --amend --reset-author --signoff` (deleting the superseded `Signed-off-by` line), and force-pushing.

PR #2388, opened 2026-08-24
var: support broken-down idents, signing key, multiple args, and -z

anpl1623:
Teach `git var` to expose individual identity components and commit signing configuration, and allow querying multiple variables with optional NUL-termination.

### Changes since v7:
- Modelled multi-variable querying after `git var -l`, printing `VARIABLE=value` pairs (or `VARIABLE\nvalue\0` when `-z` is given) instead of bare values, per feedback from Junio C. Hamano.
- Formatted multi-valued variables in multi-variable mode as repeated `VARIABLE=value` entries (or `VARIABLE\nvalue\0` with `-z`), eliminating stream ambiguity and the need for extra trailing delimiters.
- When querying multiple variables, unset variables are omitted from output and the command terminates with exit code 1.
- Updated Documentation/git-var.adoc and t/t0007-git-var.sh accordingly.

### Changes since v6:
- Grouped `GIT_AUTHOR_*` and `GIT_COMMITTER_*` entries together into concise definitions in `Documentation/git-var.adoc` to avoid repetitive descriptions, per feedback from Junio C. Hamano.
- Added explicit `BUG()` checks in `ident_part()` for NULL name/email pointers and date/timezone to protect against unforeseen changes in `split_ident_line()`.
- Changed `struct git_var` member `multiread` to return `void` (`void (*multiread)(struct string_list *)`) since `list->nr` communicates length.
- Simplified `list_vars()` to use an unconditional `else` block for `ptr->multiread` instead of redundant `else if`.
- In `cmd_var()`, do not set a non-zero exit status for missing variables in multi-variable queries; callers can see the empty record (e.g. `GIT_CONFIG_SYSTEM` under `GIT_CONFIG_NOSYSTEM`), matching Phillip Wood's recommendation.
- Used `putc(term, stdout)` in `cmd_var()` when emitting delimiters.
- In `t/t0007-git-var.sh`, converted `-z` tests to pipe output through `nul_to_q` so `test_cmp` produces clean diffs rather than binary comparison errors.

### Changes since v5:
- Reverted cosmetic refactoring in `get_git_var()` to keep the diff minimal and focused on adding `ptr->multiread`.
- Added `BUG("unknown ident_part %d", part)` to default case in `ident_part()` and removed redundant NULL checks on `split.name_begin` and `split.mail_begin`.
- Adopted first-class `int (*multiread)(struct string_list *)` callback in `struct git_var` for multi-valued variables like `GIT_CONFIG_GLOBAL`, cleanly populating a `struct string_list` instead of relying on embedded NUL buffers.
- Improved `GIT_AUTHOR_*` and `GIT_COMMITTER_*` documentation in `Documentation/git-var.adoc` to describe the values that would be used if you were to run `git commit` right now.
- Explicitly documented that single-variable queries and `git var -l` do not print an extra delimiter after multi-valued variables.

### Changes since v4:
- Simplified `git_signing_key()` to directly call `get_signing_key()` as used throughout Git (in `tag`, `send-pack`, and `sign_buffer()`).
- Renamed `null_term` to `nul_term` across `builtin/var.c`, and simplified `show_config()` callback handling.
- Replaced the redundant pre-validation loop in `cmd_var()` by validating arguments directly in the main execution loop.
- When querying multiple variables, print an empty record (blank line or `\0` with `-z`) for any variable that has no value, and continue printing remaining variables instead of terminating prematurely.
- Switched multi-valued variable storage (`git_config_val_global()`) to internal `\0` delimiters, iterating directly through string sequences without allocating a temporary `string_list`.
- For multi-variable queries, output an extra delimiter (`\n` or `\0`) after multi-valued variables to clearly mark the end of their list.
- Explicitly documented the `git var -l -z` format and multi-variable handling in `Documentation/git-var.adoc`.
- Added comprehensive tests for unset variables and multi-valued stream delimiters in `t/t0007-git-var.sh`.

### Changes since v3:
- Renamed `GIT_DEFAULT_KEY` to `GIT_SIGNING_KEY` per feedback from Phillip Wood and Junio C Hamano; dropped the alias mechanism and `commit.gpgsign` check.
- Used `parse_options()` with `PARSE_OPT_STOP_AT_NON_OPTION` in `builtin/var.c`, strictly enforcing that options precede variable arguments.
- Adopted `git config list -z` format (`key\nvalue\0`) for `git var -l -z` to prevent ambiguity with `=` in config keys.
- Delimited multi-valued variable outputs (e.g. `GIT_CONFIG_GLOBAL`) with NUL bytes under `-z`.
- Replaced `char part` in `ident_part()` with `enum ident_part`.
- Split synopsis in `Documentation/git-var.adoc` into separate lines for `-l` and `<variable>...`, and removed awkward legacy phrasing ("of a piece of code").
- Added tests in `t/t0007-git-var.sh` covering the new `-z` format, multi-valued `-z`, and argument ordering.

### Changes since v2:
- Drop `git ident` / `git whoami` subcommand entirely.
- Add `GIT_AUTHOR_NAME`, `GIT_AUTHOR_EMAIL`, and `GIT_AUTHOR_DATE`.
- Add `GIT_COMMITTER_NAME`, `GIT_COMMITTER_EMAIL`, and `GIT_COMMITTER_DATE`.
- Add `GIT_SIGNING_KEY` to resolve commit signing keys.
- Teach `git var` to accept multiple variable arguments (`git var <var1> <var2> ...`).
- Add `-z` option to terminate outputs with NUL bytes (including `git var -l -z`).
- Update `Documentation/git-var.adoc` and `t/t0007-git-var.sh`.

CC: "brian m. carlson" <sandals@crustytoothpaste.net>, Jeff King <peff@peff.net>, Junio C Hamano <gitster@pobox.com>
cc: Ben Knoble <ben.knoble@gmail.com>


Timo7670:
Sleepwalk

anpl1623:
> **"brian m. carlson"** wrote [on the Git mailing list](https://lore.kernel.org/git/ao4IDZMbgvTTYksS@fruit.crustytoothpaste.net) ([how to reply to this email](https://gitgitgadget.github.io/reply-to-this)):
> 
> ```
> On 2026-08-25 at 20:46:42, Andrew Pleeter via GitGitGadget wrote:
> > From: anpl1623 <andrewpleeter@gmail.com>
> 
> You will probably want this to match your real name since you're using
> it in the email.  We prefer that people use their real names, but don't
> require it, but I assume that it's not a problem since you just CC'd
> yourself using it (and it's also in your email address).
> 
> > Add a builtin 'whoami' command to inspect and display the resolved
> > author and committer identity along with the commit signing
> > configuration (GPG/SSH key ID and commit.gpgsign status) used when
> > creating Git commits.
> > 
> > Support optional flags (--author, --committer, --name, --email,
> > --signing-key, and --verbose) for targeted querying and scripting.
> 
> I suspect users will want a way to get _all_ of the output in a
> machine-readable way, so you'd probably want to provide some method of
> doing that.  Note that because your existing endpoints provide
> translated strings, they are not suitable for this.  That doesn't mean
> that they should not be translated (because they should) but we'd
> probably want a format like the following:
> 
> user.author.name=A U Thor
> user.author.email=author@example.com
> 
> Other formats are possible, though.
> 
> Possibly a `-z` option for NUL-terminated instead of LF-terminated
> output might be warranted as well unless we're certain that our output
> will never contain a newline (hint: config options can).
> 
> > Include documentation in Documentation/git-whoami.adoc and regression
> > tests in t/t0015-whoami.sh.
> > 
> > Signed-off-by: anpl1623 <andrewpleeter@gmail.com>
> 
> Again, you'll want to sign this off with your real name.
> 
> >     MOTIVATION
> >     
> >     Users often work across multiple environments, profiles, or repositories
> >     with different global/local configs and signing keys. Currently,
> >     verifying what identity and signing key will be attached to a new commit
> >     requires checking several individual git config and git var settings.
> >     git whoami provides a simple, direct porcelain command to verify this in
> >     one step.
> 
> I think this should go in the commit message.  I thought to myself,
> "Well, there are already ways to get this information, so why add a new
> one?"  Telling us why your patch is compelling and solves an important
> purpose is appropriate for the commit message.
> 
> I might also like to see an explanation as to why this wouldn't work
> better in `git var` or elsewhere instead, since much of the information
> is already there.  Since that's an alternative you've rejected, tell us
> why and sell us on your vision.
> 
> > +	repo_config(the_repository, git_default_config, NULL);
> 
> Let's not add more uses of `the_repository`.  Use the `repo` argument to
> the main function above, taking care to handle the NULL case.
> 
> > +	repo_config_get_bool(the_repository, "commit.gpgsign", &gpgsign);
> > +	repo_config_get_string(the_repository, "user.signingkey", &signing_key);
> > +	repo_config_get_string(the_repository, "gpg.format", &gpg_format);
> > +	repo_config_get_string(the_repository, "gpg.ssh.defaultkeycommand", &ssh_default_key_cmd);
> > +
> > +	is_ssh = gpg_format && !strcmp(gpg_format, "ssh");
> > +
> > +	if (signing_key && *signing_key) {
> > +		resolved_key = xstrdup(signing_key);
> > +	} else if (is_ssh) {
> > +		if (ssh_default_key_cmd && *ssh_default_key_cmd)
> > +			resolved_key = get_signing_key_id();
> > +	} else if (gpgsign) {
> > +		resolved_key = get_signing_key_id();
> > +	}
> 
> Should this also do something useful for X.509 keys?
> 
> Overall, I don't have a strong need for this and I'm fine using the
> existing functionality.  However, I see how it could be useful and if it
> were merged and available in the versions of Git I use, then I might
> make use of it.
> 
> Perhaps others think this is compelling, though, so I'm interested to
> hear other opinions about the utility of the command.
> -- 
> brian m. carlson (they/them)
> Toronto, Ontario, CA
> ```

Hi Brian,

Thank you for the detailed feedback! I have updated the patch to address all of your points:

1. Real Name & Sign-off:
   Updated the commit author and Signed-off-by trailer to use my real name:
   "Andrew Pleeter <andrewpleeter@gmail.com>".

2. Machine-Readable & NUL-Terminated Output:
   Added a `--porcelain` option that outputs key-value pairs in the `user.author.*`,
   `user.committer.*`, `user.signingkey`, `gpg.format`, and `commit.gpgsign` namespace,
   as well as `-z` (`--null`) for NUL-terminated output across `--porcelain` and single-field
   selectors.

3. Motivation in Commit Message:
   Moved the motivation into the commit message, detailing why 'git whoami' provides
   a cleaner, unified porcelain interface compared to querying 'git var' / 'git config'
   separately.

4. Removal of the_repository:
   Removed '#define USE_THE_REPOSITORY_VARIABLE' and refactored 'builtin/whoami.c' to use
   the 'repo' argument with 'repo_config(repo, ...)', cleanly handling non-repository
   invocations.

5. Signing Key Resolution & X.509:
   Documented the explicit resolution order across OpenPGP, X.509, and SSH.

All regression tests in t/t0015-whoami.sh and documentation linters pass.

managedkaos:
I'm interested in how this plays out, with no preference either way.  Right now I use an alias:

```
alias whome="git config -l | grep user"
```

which works fine for me.  Perhaps having a built in function might be cool because I don't have this alias on every machine i workon.  But usually, those are the systems where I don't care too much about who `git` thinks I am; I'm pulling code anonymously instead of pushing code as my configured `user.*` values.

anpl1623:
> **Phillip Wood** wrote [on the Git mailing list](https://lore.kernel.org/git/07d1ef05-97df-481a-b647-cbe33a57f902@gmail.com) ([how to reply to this email](https://gitgitgadget.github.io/reply-to-this)):

Hi Phillip,

Thank you for the detailed review and actionable suggestions!

> > -			free(val);
> > +		} else if (ptr->multiread) {
> 
> We should just assume that ptr->multread is set when ptr->read is not, or possibly add an else clause that calls BUG().

Agreed. Since the loop invariant requires `ptr->read || ptr->multiread`, we have simplified this to an unconditional `else` block in v7.

> > +			if (!val) {
> > +				if (argc == 1)
> > +					return 1;
> > +				ret = 1;
> 
> What's the benefit of this? The caller can see there was an empty value so why do we want a non-zero exit code as well. For example, if the caller is asking for GIT_CONFIG_SYSTEM and GIT_CONFIG_GLOBAL but the user ran the script with GIT_CONFIG_NOSYSTEM then that shouldn't be an error - the caller should just not use the system config.

That makes complete sense. We have dropped `ret = 1` for multi-variable queries. When `argc > 1`, querying an unset variable outputs an empty record (a newline or a NUL byte) and continues with exit status 0, allowing caller scripts to reliably ingest positional output without premature script failure. Single-variable queries (`argc == 1`) continue to exit with code 1.

> > +			if (git_var->multiread(&list) || !list.nr) {
> 
> Why are we checking the return value of the function and the list length - surely the list length tells us everything we need to know.

Agreed. We changed the signature to `void (*multiread)(struct string_list *)` and now rely entirely on `!list.nr`.

> I think the above can be simplified to
> 
> 	} else {
> 		struct string_list list = STRING_LIST_INIT_NODUP;
> 		
> 		git_var->multiread(&list);
> 		if (argc == 1 && !list.nr) {
> 			return 1;
> 		}
> 		for (j = 0; j < list.nr; j++)
> 			printf("%s%c", list.items[j].string, term);
> 		if (argc > 1)
> 			putc(term, stdout);
> 
> 		string_list_clear(&list, 0);
> 	}

We adopted this exact structure in v7, including the switch to `putc(term, stdout)`. (We retained `STRING_LIST_INIT_DUP` because `git_config_val_global()` dynamically allocates paths via `git_global_config_paths()`, so string duplication ensures clean ownership before freeing `user` and `xdg`).

> I did notice they use test_cmp() on files containing '\0' which isn't a good idea because diff will see them as binary files. We have helpers like nul_to_q to translate nul to a printable character.

Thank you for catching this! We have updated all `-z` test cases in `t/t0007-git-var.sh` to pipe output through `nul_to_q` before comparing with `test_cmp`.

Best regards,
Andrew

anpl1623:
> **Junio C Hamano** wrote [on the Git mailing list](https://lore.kernel.org/git/xmqq4ifytlzk.fsf@gitster.g) ([how to reply to this email](https://gitgitgadget.github.io/reply-to-this)):

Hi Junio,

Thank you for the review and guidance!

> This is better than the previous iteration, but wastes reader's time
> with full of repetitions.  Have you looked at the one suggested in
> https://lore.kernel.org/git/xmqqbjaecjxb.fsf@gitster.g/ for example
> to present more information in much more concise way?

My apologies for not adopting your grouped suggestion sooner. In v7, we have replaced the individual entries with your grouped definitions in Documentation/git-var.adoc:

    `GIT_AUTHOR_IDENT`::
    `GIT_AUTHOR_NAME`::
    `GIT_AUTHOR_EMAIL`::
    `GIT_AUTHOR_DATE`::
        The authorship information that would be recorded in the
        resulting commit object if you ran `git commit` right now.
        `GIT_AUTHOR_IDENT` consists of the author's name, e-mail
        address, and timestamp+timezone. These three pieces of
        information are available separately as `GIT_AUTHOR_NAME`,
        `GIT_AUTHOR_EMAIL`, and `GIT_AUTHOR_DATE`.

    (and similarly for `GIT_COMMITTER_*`).

> This is better in that it no longer returns NULL upon an impossible
> condition like the previous iteration did.  Even better, we could
> protect ourselves against breakage caused by careless updates to
> split_ident_line() and git_*_info() functions we rely on by keep the
> check but mark BUG()...
> 
> With the way committer_date() and author_date() are called (below), can
> fmt_ident() ever return an ident without datestamp and timezone,
> requiring us to fall back on NULL returns like this?  You are not
> passing IDENT_NO_DATE flag anywhere, are you?

Following up on your due diligence suggestion: we audited `fmt_ident()` and `split_ident_line()`. When called without `IDENT_NO_DATE` (as is done for all ident variables here), `fmt_ident()` always produces date and timezone fields.

In v7, we added explicit defensive assertions:
- In `IDENT_NAME`: `BUG("split_ident_line() gave NULL names???")` if either pointer is NULL.
- In `IDENT_MAIL`: `BUG("split_ident_line() gave NULL mail???")` if either pointer is NULL.
- In `IDENT_DATE`: asserted `!split.date_begin || !split.tz_end` with `BUG("split_ident_line() gave NULL date/tz???")`, and simplified extraction to `xmemdupz(split.date_begin, split.tz_end - split.date_begin)`.

> However, I am not sure why we want an extra delimiter only after a
> multi-valued variable.  Does it mean that the reading script needs
> to be aware of which variables are multi-valued and which are not?
> ...
> If I were designing this, I'd rather (1) get rid of the "empty line"
> convention for multi-valued variables, and (2) model multi-variable
> mode more after "-l" mode.  IOW, instead of thinking of the case
> where the user gave us two variables like two "git var VARIBLE$N"
> calls given back to back, thinking it more like "git var -l | grep
> -e VARIABLE1= -e VARIABLE2=".

Regarding the multi-variable design and delimiters:

1. **Why values-only instead of `key=val`:**
   The primary motivation for querying multiple variables (e.g. `git var GIT_AUTHOR_NAME GIT_AUTHOR_EMAIL`) was to allow scripts to easily unpack values directly without stripping variable names, e.g.:

       read -r name email < <(git var GIT_AUTHOR_NAME GIT_AUTHOR_EMAIL)

   or with `-z`:

       { IFS= read -r -d '' name; IFS= read -r -d '' email; } < <(git var -z GIT_AUTHOR_NAME GIT_AUTHOR_EMAIL)

   If `git var <var1> <var2>` emitted `VARIABLE=value`, callers would need an extra stripping step, reducing the convenience over `git var -l`.

2. **Stream disambiguation for multi-valued variables:**
   Because single-valued variables emit exactly one delimiter, a reader querying only single-valued variables knows that line 1 is variable 1, line 2 is variable 2, etc.

   When multi-valued variables are mixed in, an extra delimiter is appended after the list so that the stream remains deterministic:
   - A variable with 0 values emits an empty record (a lone delimiter).
   - A multi-valued variable with N values emits N delimited records followed by an extra delimiter marking the end of that list.

   A consumer querying `git var -z GIT_AUTHOR_NAME GIT_CONFIG_GLOBAL GIT_AUTHOR_EMAIL` can thus read:
   - 1 record for author name.
   - 0 or more records until an empty record marks the end of `GIT_CONFIG_GLOBAL`.
   - 1 record for author email.

   `git var -l` (and `git var -l -z`) remains the canonical interface if callers want `key=value` pairs. However, if the project consensus is that multi-argument queries should instead act as filters on `git var -l` (e.g., `git var -l VAR1 VAR2` printing `VAR1=val1\nVAR2=val2\n`), we are happy to adapt to that model!

Thanks,
Andrew

dscho:
@anpl1623 please note that _none_ of your replies in this PR will ever reach the intended recipient, as [the welcome message](https://github.com/git/git/pull/2388#issuecomment-5402238258) said:

> Please make sure to monitor the discussion in that thread and to address comments and suggestions (while the comments and suggestions will be mirrored into the PR by GitGitGadget, you will still want to [reply via mail](https://gitgitgadget.github.io/reply-to-this)).

Your PR comments are _not_ mirrored to the Git mailing list. Neither Phillip nor Junio got your messages.

PR #2403, opened 2026-09-12
ci: only warn about perforce/git-lfs/JGit on platforms that need them

HaraldNordgren:
Only warn about a missing perforce/git-lfs/JGit install on the platforms that actually need and attempt them (ubuntu-*, plus macOS for perforce), since every other platform never installs them and was warning regardless.


Cyberverse-cent0:
not bad

